PR #26, opened 2013-09-09
Extend error and notice notification with threadsafe variants

pepijnve:
Extends the C wrapper interface with a means to associate user data with the notification handlers. This enables threadsafe handling of error and notice messages on the consumer side.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,10 +268,10 @@ else()
    find_program(SH sh)
    if(SH)
       execute_process(COMMAND ${SH} -c 
-      "cd ${CMAKE_SOURCE_DIR} && ${CMAKE_SOURCE_DIR}/tools/svn_repo_revision.sh")
+      "cd ${CMAKE_CURRENT_SOURCE_DIR} && ${CMAKE_CURRENT_SOURCE_DIR}/tools/svn_repo_revision.sh")
 
-      file(RENAME "${CMAKE_SOURCE_DIR}/geos_svn_revision.h"
-      "${CMAKE_BINARY_DIR}/geos_svn_revision.h")
+      file(RENAME "${CMAKE_CURRENT_SOURCE_DIR}/geos_svn_revision.h"
+      "${CMAKE_CURRENT_BINARY_DIR}/geos_svn_revision.h")
    else()
       message("*** sh-compatible command not found, cannot create geos_svn_revision.h")
       message("*** Check SVN revision and create revision header manually:")
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -92,7 +92,10 @@ extern "C" {
  *
  ************************************************************************/
 
+typedef struct GEOSContextHandle_HS *GEOSContextHandle_t;
+
 typedef void (*GEOSMessageHandler)(const char *fmt, ...);
+typedef void (*GEOSMessageHandler_r)(GEOSContextHandle_t handle, void *data, const char *fmt, ...);
 
 /* When we're included by geos_c.cpp, those are #defined to the original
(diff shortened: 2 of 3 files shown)


pepijnve:
@srtk revisiting this based on a question on the spatialite mailing list. What was the naming issue we got stuck on exactly? I tried to figure it out based on the past discussion, but I can't seem to pinpoint what the exact issue was.

strk:
It was the confusion between GEOSContext_setNoticeHandler_r and GEOSContext_setNoticeMessageHandler_r signatures (and the one for Error).
I guess adding deprecation notices on those functions too could do it (re-reading the changes)

pepijnve:
Right, thanks for refreshing my memory.
Agreed the distinction between those two and which one is preferred is not obvious.
I see three options:

1. Come up with a more distinctive name for the new variants. I still don't have any that don't sounds ridiculous I'm afraid.
2. Change the name to `GEOSContext_setNoticeHandler_r_ex` or `GEOSContext_setNoticeHandler_r` _v2 or something like that. Not an uncommon solution in C libraries although it's a bit ugly IMO.
3. Keep the `GEOSContext_setNoticeMessageHandler_r` name

If you let me know which one has your preference I'll make the necessary changes. 

Either way a deprecation comment on the older version pointing to the new one is indeed a good idea.

strk:
I surrender to GEOSContext_setNoticeMessageHandler_r (current naming).
The deprecation warning must be in place _and_ it must be ensured that
old clients do not break.

strk:
@robe2 I think this would be better included in next release (3.5.0)
@pepijnve it needs a ticket on the official tracker to ensure this is the case, could you file one ?
http://trac.osgeo.org/geos/

strk:
I found an already existing trac ticket: https://trac.osgeo.org/geos/ticket/663

pepijnve:
@strk I'm a little strapped for time at the moment. What's the target release date for 3.5.0? In other words, by when do you need this PR polished up and ready to merge?

strk:
It's @robe taking care of release this round.
I think we're talking about 2/3 weeks from now.
Ideally the patch would be clean and ready within 1 week
for further testing.

BTW, @esseffe spotted a few initialization problems, hopefully he'll
be submitting the patch here soon

spatialite:
I've rather extensively tested this patch on SpatiaLite, and I can confirm that it works nicely and doesn't break any back compatibility. more important than all, it definitely fixes all thread-safety issues.

anyway during my tests (heavily based on Valgrind) I discovered a rather trivial error (uninitialized variables) causing crash, the following patch effectively fixes the problem.
```
GEOSContextHandle_HS(GEOSMessageHandler_r nf,
                     void *nd, void(*free_nd)(void*),
                     GEOSMessageHandler_r ef, void *ed,
                     void(*free_ed)(void*))
{
    memset(msgBuffer, 0, sizeof(msgBuffer));
    geomFactory = GeometryFactory::getDefaultInstance();
    WKBOutputDims = 2;
    WKBByteOrder = getMachineByteOrder();
+   noticeData = NULL;
+   freeNoticeData = NULL;
+   errorData = NULL;
+   freeErrorData = NULL;
    setNoticeHandler(nf, nd, free_nd);
    setErrorHandler(ef, ed, free_ed);
    initialized = 1;
}
```

strk:
@pepijnve please also rebase to trunk after accepting @spatialite contrib.
Please let me know if you'll be able to do within the upcoming week.

strk:
@spatialite there's no "freeNoticeData" nor "freeErrorData" in the code I'm reading ?

strk:
So this was merged in r4060 and @spatialite concerns were addressed in r4061, those two should be shortly available in the git mirror.

PR #30, opened 2013-11-22
This upgrade the detection of an intersection between a line and a point

Remi-C:
... by adding a tolerance over digits from 13 to 17

The problem is the test to know is point is left or right of a line is precise, but the test to know if a point is on a line doesn't work well.e found to intersect the line.
(an example of such a line : "LINESTRING(1447.451535 534.348973,1447.89016 543.902917)" )
In previous version, some line had no point at all (ie : no points were on the line,  except the points on extremities.).
Most of the time ( p>(1-1/10^6) ), computing  a point on a line using ST_ClosestPoint or ST_LineInterpolatePoint would give a point which was not on the line according to the test.

So instead, we create 2 points offsetted by an epsilon (corresponding to the 13 digits of the coordinates), and then check if the left offsetted point is left of the line, and the right offseted point is right of the line, or if any is on the line.

NOTE : now we have a probability of error < 1/10^6 for coordinates up to digits (i didn't tested more). This means there is no warranty that a line will have points at all , even if odds are very low that a line would have no points. Snatching vertex to grid would certainly solve the problem if such a line was encountered.

The postgis functions that have been improved are : 
 ST_Union ;ST_UnaryUnion ;ST_Intersection;  ST_Relate, ST_Substring, ST_BuildArea ...

image

The plpgsql code to test the changes regarding the new behavior of st_intersects is  :  

```SQL
---building a test table with random lines
    SELECT setseed(0.192169354297221); --settign seed so every execution gives same result
            DROP TABLE  IF EXISTS  temp_test_precision_geos_serie_max;
            CREATE TABLE temp_test_precision_geos_serie_max AS
                WITH point_number AS (
                    SELECT power(10,prec.p-1)::bigint As pprec, 1000000 AS pnum, trunc(random() ::numeric,prec.p) AS rand, prec.p AS prec
                    FROM (SELECT 8 aS p) AS prec
                ),
                 serie AS (
                        SELECT s1 AS s
                        FROM point_number,generate_series(1,pnum) AS s1
                ),
                points AS(
                    SELECT ST_MakePoint(
                            trunc((s_random()/pnum )::numeric,prec)_pnum
                            ,trunc((s_random()/pnum )::numeric,prec)_pnum )AS p1
                        ,ST_MakePoint(
                            trunc((s_random()/pnum )::numeric,prec)_pnum
                            ,trunc((s_random()/pnum )::numeric,prec)_pnum ) AS p2
                    FROM point_number,serie
                ),
                line AS (
                    SELECT _, St_MakeLine(p1,p2) AS line
                    FROM points
                )
                SELECT row_number() over() AS uid
                    ,_ 
                    , ST_LineInterpolatePoint(
                        line,  rand) AS ipoint 
                FROM point_number,line
                ORDER BY p1 ASC , p2 ASC;

--testing for each line
SELECT setseed(0.192169354297221);
            DROP TABLE IF EXISTS temp_test_precision_postgis;
            CREATE TABLE temp_test_precision_postgis AS
            with test AS (
                SELECT uid, p1, p2, line, ipoint
                FROM temp_test_precision_geos_serie_max
                --LIMIT 1
            ),
            closest AS (
                SELECT \* , ST_ClosestPoint(line,ST_Translate(ipoint, ST_X(ipoint)_round(random()::numeric,10),ST_Y(ipoint)_round(random()::numeric,10))) AS cpoint
                FROM test
            ),
            verif AS (
                SELECT St_Intersects(line,ipoint) AS intersect_line_interpolated_point
                    , St_Intersects(line,cpoint) AS intersect_line_closest_point
                    , ST_Relate(line, cpoint) AS relate_line_closest_point
                    , line, ipoint  
                FROM closest
            )
            SELECT row_number() OVEr () AS id_qgis,*
            FROM verif
            WHERE intersect_line_interpolated_point=false  OR  intersect_line_closest_point=FALSE   ;
```

### DIFF
--- a/src/algorithm/LineIntersector.cpp
+++ b/src/algorithm/LineIntersector.cpp
@@ -335,8 +335,8 @@ LineIntersector::computeIntersection(const Coordinate& p,const Coordinate& p1,co
 
 	// do between check first, since it is faster than the orientation test
 	if(Envelope::intersects(p1,p2,p)) {
-		if ((CGAlgorithms::orientationIndex(p1,p2,p)==0)&&
-			(CGAlgorithms::orientationIndex(p2,p1,p)==0)) {
+		if  (LineIntersector::hasIntersection(p, p1, p2) ==true) //checking if there is an intersection at all
+			{
 			isProperVar=true;
 			if ((p==p1)||(p==p2)) // 2d only test
 			{
@@ -368,10 +368,33 @@ bool
 LineIntersector::hasIntersection(const Coordinate& p, const Coordinate& p1, const Coordinate& p2)
 {
 	if(Envelope::intersects(p1,p2,p)) {
-		if ((CGAlgorithms::orientationIndex(p1,p2,p)==0)&&
-			(CGAlgorithms::orientationIndex(p2,p1,p)==0)) {
-			return true;
-		}
+	//computing the 2points Left and right that are the offset of the point by min_precision.
+	//if the Left point is left from the line and the right point right of the line, or any on the line, return true, else return false
+	
+		double min_precision = pow(10,-12) ; 
+		//this parameter control the epsilon. for instance, 2 lines having the same coordinates to 12 digits will be same lines
+	
+
+			//computing theoffsetted points
+			Coordinate pL;Coordinate pR;
+			short int positionL; short int positionR;
+	
+			 pL.x= p.x+min_precision*(p1.y-p2.y)/2;
+			 pL.y= p.y+min_precision*(p2.x-p1.x)/2;
+			
+			 pR.x= p.x-min_precision*(p1.y-p2.y)/2;
+			 pR.y= p.y-min_precision*(p2.x-p1.x)/2;
+			
+			positionL = CGAlgorithms::orientationIndex(p1,p2,pL);
+			positionR = CGAlgorithms::orientationIndex(p1,p2,pR);
+			
+			//note : I don't know if we have to reverse check, I'm guessing it was for robustness issue: now it's not really necessary??
+			
+			if(  (positionL==1 && positionR==-1) || positionL==0 || positionR==0) {
+				return true;
+			} else { 
+				return false;
+			}
 	}
 	return false;
 }


mloskot:
## IMPORTANT

Since [`svn-trunk` has been renamed to `master`](https://trac.osgeo.org/geos/ticket/822), we have rebased this PR against the current `master`. Hopefully, it won't break anything.

dr-jts:
@Remi-C this is an interesting approach.  Do you feel that it improves the robustness of operations, or just returns a cleaner output?  Did you do much testing on datasets of polygons to see what kind of impact this might have?  I.e. does it reduce or increase the incidence of TopologyExceptions?

We are going to be looking at eliminating TopologyExceptions in overlay operations. Likely this will involve snap-rounding to a tolerance.   Likely the tolerance will be user-defined, but perhaps there is also justification for always using a minimum tolerance.

dr-jts:
Also, I'm wondering about the computation you use for checking on line with a tolerance. Would it be any faster/simpler just to check the distance from the original point to the line?

Remi-C:
Hi @dr-jts !
Has far as I remember (6 years latter ;-) ),
distance from point to line is actually hard to compute in a robust fashion.
You are going to have to do several multiplications / divisions, each dropping your precision.
The ideal solution involves either exact arithmetic (not an option !), or some kind of adaptive method (as long as the result is ambiguous, use more bits to compute).

Both solutions are fairly complex.
Many papers are dedicated to the subject actually.

Instead I proposed a stupid trick that improves the situation in most of the cases, but does not offer any guarantees. 
The approach exploits the fact that we have a very robust an fast way to tell if a point is left or right of a line. It amounts to computing the sign of a determinant [here](https://github.com/libgeos/geos/blob/f43aa53d27b3889823365317fcf4ff45e60dc90f/src/algorithm/CGAlgorithmsDD.cpp#L113).
It's also fairly robust to create  offset points.

In all honesty this solution feels very hacky, 
but solved the problem most of the time. 
In perticular, the solution involves using a threshold, which would need to be managed properly I guess.

The bottom of the problem was that using ST_ClosestPoint or ST_LineInterpolatePoint would give a point which was not on the line according to various tests (ST_intersects, ...).

dr-jts:
Ok, that's indeed a clever solution to checking if a point is within a distance of a line segment.  (Although I wonder why computing an offset point is more robust than computing distance to line, since they both involve distance-like computation.)

As you say, introducing an arbitrary tolerance likely indicates that this value should be surfaced and be user-selectable.  This might be implemented as a tolerance value as an optional parameter to spatial predicates.  That way users can be very specific about what approximation they want to have.  

Providing an explicit tolerance parameter changes the topological model.  This makes it obvious that "is intersects consistent with intersection" is not an appropriate question.  What can be asked is "is intersects consistent with intersection up to a given tolerance".  It seems likely the answer to this is true for any reasonable tolerance value.

pramsey:
I feel like doubledouble has moved beyond this.

PR #40, opened 2014-12-01
Port DoubleDouble and CGAlgorithmsDD support

mloskot:
This is a prototype of **DoubleDouble** support for GEOS, as an analog to JTS DoubleDouble.

The PR includes:
* [ttmath](http://www.ttmath.org ) 0.9.3 library
* [doubledouble](http://www.boutell.com/fracster-src/doubledouble/doubledouble.html) 2.2 library (GPL!)
* Minimal port of JTS `CGAlgorithmsDD` class
* DoubleDouble (DD) support can be switched to either ttmath or doubledouble at build-time, see detailed logs of the commits below.
* A bunch of curious test cases:
   * Test case of `orientationIndex` method for orientation vs inverse (see comment in 
[RobustDeterminant.java/RobustDeterminant.orienationIndex](http://sourceforge.net/p/jts-topo-suite/code/626/tree//trunk/jts/java/src/com/vividsolutions/jts/algorithm/RobustDeterminant.java); JTS r931 added this test in [jts/algorithm/CGAlgorithmsTest.java](http://sourceforge.net/p/jts-topo-suite/code/931/))
   * Test cases of **test on segment** vs **test on vertex of segment** based on geometries in both formats, WKT and WKB. See commit logs for explanation of the **curious** results obtained using both JTS and GEOS (with DD support), for WKT vs WKB. Here is corresponding patch with those tests for JTS [jts-algorithm-test-point-on-segment.patch
](https://gist.github.com/mloskot/54c3b88a800006bd9cde).

This PR is directly related to the problems discussed in this GEOS ticket http://trac.osgeo.org/geos/ticket/591

Next, I'm going to contact @dr-jts  if he would be willing to comment on the JTS results obtained for the **test on segment** case.

### DIFF
--- a/include/geos/algorithm/CGAlgorithmsDD.h
+++ b/include/geos/algorithm/CGAlgorithmsDD.h
@@ -1,0 +1,94 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2014 Mateusz Loskot <mateusz@loskot.net>
+ * Copyright (C) 2001-2002 Vivid Solutions Inc.
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation. 
+ * See the COPYING file for more information.
+ *
+ **********************************************************************
+ *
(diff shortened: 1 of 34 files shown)


strk:
See also #42 as it contains a testcase that might be fixed by this work

PR #42, opened 2014-12-16
Floating point precision robustness during intersection

tokheim:
See http://trac.osgeo.org/geos/ticket/716

### DIFF
--- a/include/geos/algorithm/RayCrossingCounter.h
+++ b/include/geos/algorithm/RayCrossingCounter.h
@@ -93,6 +93,22 @@ public:
 	static int locatePointInRing(const geom::Coordinate& p,
 	         const std::vector<const geom::Coordinate*>& ring);
 
+	/** \brief
+	 * Returns the index of the direction of the point <code>q</code>
+	 * relative to a vector specified by <code>p1-p2</code>.
+	 *
+	 * @param p1 the origin point of the vector
+	 * @param p2 the final point of the vector
+	 * @param q the point to compute the direction to
+	 *
+	 * @return 1 if q is counter-clockwise (left) from p1-p2
+	 * @return -1 if q is clockwise (right) from p1-p2
+	 * @return 0 if q is collinear with p1-p2
+	 */
+	static int orientationIndex(const geom::Coordinate& p1,
+	         const geom::Coordinate& p2,
+	         const geom::Coordinate& q);
+
 	RayCrossingCounter(const geom::Coordinate& point)
 	:	point( point),
 		crossingCount( 0),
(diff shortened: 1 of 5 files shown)


strk:
This test still fails applying the patch in this PR: https://github.com/libgeos/libgeos/commit/836e7010be584527ea9b0245a07e0e803e9b8809 (for the record)

strk:
@tokheim could you please add another case to the test with the second geometry reversed ?
I'd be 010300000001000000040000000000000000002E400000000000001440020018EA49852440000008012F174140FC0000505D92843FFFFF7762C29C1F400000000000002E400000000000001440

Even better if the test contains 4 cases with all 4 combinations.
This is because the current code finds the geometries intersecting if  you reverse the second one, so it's good to ensure getting a consistent result with the patch.

Reverse of first geometry:
0103000000010000000400000001DA86DAF74BA83F2FA9595EF7FE1F40000000000000244000000000000024408CF48F8922722440C9D18A61E80A414001DA86DAF74BA83F2FA9595EF7FE1F40

strk:
Please consider also adding a unit test, maybe in tests/unit/algorithm/CGAlgorithms/computeOrientationTest.cpp, so the "reverse" invariant can be more explicit too

strk:
\cc @dr-jts @mloskot

mloskot:
As confirmed in the ticket comment, tested with x86-64 build of the CGAlgorithmsDD port (#40)

```
xmltester.exe bug716.xml
Files: 1
Tests: 1
Failed: 0
Succeeded: 1
```

tokheim:
@strk I've added tests for the other orientations.

The effect you see in https://github.com/libgeos/libgeos/commit/836e7010be584527ea9b0245a07e0e803e9b8809 is similar to what happened to my geometries. I've extracted the relevant coordinates for a similar (currently failing) test case below, is that what you wanted in computeOrientationTest?
```
Coordinate p0(0.010044793132693013, 7.903085268568247);
Coordinate p1(10.260329547338191, 34.181121949106455);
Coordinate p(0.04745459120333707, 7.998990511152711);
```
Both sets of coordinates have in common that there is a rather large difference in the magnitude of ```p0``` and ```p1```  in the same dimension. This means that when ```orientationIndex``` subtracts ```p``` with ```p1```, ```p``` is left with more of its least significant digits in that dimension than when we swap  ```p1``` and  ```p2```. 

You can make ```orienetationIndex``` insensitive to swapping by using vectors from ```p```. In addition to passing https://github.com/libgeos/libgeos/commit/836e7010be584527ea9b0245a07e0e803e9b8809, it would have some other desired properties like that ```contains``` gives the same result for CW and CCW versions of the polygons above. I created a branch at https://github.com/tokheim/libgeos/commit/0275587edc8f72a8257c01ea76e63d400ef0edbf, but I can include it in this/add another pull request if interested. Now for the bad part, it breaks a isCCW test of an almost collapsed ring from http://trac.osgeo.org/geos/ticket/398. It would seem like that test tests the floating point precision handling to the extreme, and I'm not sure if CW orientaion is required for correctness. The xml-test for the same ticket still passes.

strk:
Merged in trunk and 3.4 branches. Thanks.
Please use another PR for the orientation-agnostic improvements if you think they are still needed after the DoubleDouble porting.

PR #43, opened 2015-01-23
Allow failure of cmake/clang/64bit build (bug outside geos, I believe)

strk:
just a test for allowing travis failures, which we should probably not allow as it doesn't fail in 3.4...

https://travis-ci.org/libgeos/libgeos/builds/47505801
https://travis-ci.org/libgeos/libgeos/builds/48031859

\cc @mloskot ideas about the difference between 3.3 and 3.4 branch when it comes to CMake/clang/64bit ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,14 @@ compiler:
     - clang
 
 env:
-    matrix:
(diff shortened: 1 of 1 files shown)


strk:
travis is not failing anymore, so closing

PR #44, opened 2015-01-29
Change the README to markdown

BenMorel:
This makes it more readable on GitHub, with minor changes to the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,17 +5,20 @@ Unix
 ----
 
 Using Autotools:
-  ./autogen.sh  # in ${srcdir}, if obtained from SVN
-  ${srcdir}/configure # in build dir
+
+    ./autogen.sh  # in ${srcdir}, if obtained from SVN
+    ${srcdir}/configure # in build dir
 
 Using CMake:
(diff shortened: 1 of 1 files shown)


strk:
I don't like the .md extension :)  -- but no problem with dropping a symlink with that name to the README (which is ok if renders correctly as markdown)

BenMorel:
Then you can choose the alternative `.markdown` extension! :)

Joke aside, yes there are just minor modifications to make it render fine as markdown (such as 4 
spaces tabs instead of 2 spaces for code blocks).

I don't know if GitHub will render a symlink, though.

strk:
I think it will, but I just remembered official code is in SVN
which I dubt supports symlinks.

Gah, I hate it to be forced to change file extension !

BenMorel:
I don't want to force you, this is just a suggestion ... :)

strk:
I took a different approach, adding another .md file just for the sake of github:
https://github.com/libgeos/libgeos/commit/aa57155d18f8cbb33b5090c5ea4bae349abf4db2

BenMorel:
That's not so bad :)

PR #46, opened 2015-04-02
Syncronize CoordinateArraySequenceFactory::create signatures

manisandro:
The overriding `CoordinateArraySequenceFactory::create` methods have different default parameter values than the `create` methods in the `CoordinateSequenceFactory` base class. This is potentially ambiguous, since

    ((CoordinateArraySequenceFactory*)factory)->create(0)

will fail to compile, while

    ((CoordinateSequenceFactory*)factory)->create(0)

will compile successfully.

The patch syncronizes the `create` signatures, and adds necessary

    reinterpret_cast<std::vector<Coordinate> *>(0)

throughout the code to resolve  the ambiguities.


strk:
I would like to avoid the reinterpret_cast. Would be maybe easier to expose a method taking no argument and meaning "empty sequence with automatically guessed dimensions"...

manisandro:
Ping?

strk:
Works for me. Could you please add a test to tests/unit/geom/CoordinateArraySequenceFactoryTest.cpp for the new method, 
remove empty spaces from blank lines (git  diff --color shows them in red) and squash the commits ?

manisandro:
Ok, done.

strk:
Reading the header, isn't there a conflict between these two methods of CoordinateArraySequenceFactory ?:

```
        CoordinateSequence *create(std::vector<Coordinate> *coords) const;
        CoordinateSequence *create(std::vector<Coordinate> *coords, std::size_t dims=0) const;
```

I guess that's the reason why there was no default value for "dims" before ?

strk:
I guess the solution is dropping the single-argument version, to further sync with base class...

manisandro:
Yes indeed, good catch. The thing that is slightly ugly is that the implementation of the single-parameter version is

    return new CoordinateArraySequence(coords,3);

i.e. as if the default parameter value for the `dims` variable was 3, whereas it is 0 in the two-parameter version. But in my opinion it makes sense to clean this up properly and hence yes, remove the single-parameter version.

strk:
Let's see if any existing test breaks :)

manisandro:
Tests still pass

strk:
Pushed as r4052 (possibly 3a36e1ad241a0326001755ef59d7d433ab5a594d once it gets here)

manisandro:
Thanks

PR #47, opened 2015-04-09
Support for MultiLineString->isClosed() in C API

BenMorel:
Currently the C API reports an error when `isClosed()` is called on a `MultiLineString`.


strk:
Could you please add a regression test for this ?

BenMorel:
Done, let me know if that's what you expected!

strk:
Thanks, merged as r4049 (note you missed update of the Makefile.am, done with r4050)

PR #51, opened 2015-09-21
Change README to a symlink, and improve formatting

mwtoews:
This PR does two things:
 1. Move the README document to README.md, and make a symbolic link
 2. Format the document in Markdown, and add a few minor modifications to the content

The end result is that the project page as viewed through GitHub (which I realise is currently not the primary repo) will look as it should, without needing to navigate to the unformatted `README`.


strk:
I'd actually like to avoid giving the impression that github is _the_ place to find GEOS.
In particular, the "build status" icon is partial as GEOS is also tested by other bots (namely jenkins, run by @robe2) and will possibly be tested but more in the future (gitlabs).

Speaking of which, the gitlab mirror fails to use the README.md file prepared for github, but rather uses the README file: https://gitlab.com/geos/libgeos

strk:
finally: I don't think SVN (the official repo) supports symlinks

mwtoews:
Good question about symbolic links for SVN. Following [this Q/A](http://stackoverflow.com/q/4056092/327026), it seems that it should work, even for MS Windows 7 and later.

strk:
Some news: 

1. gitlab now renders the README.md correctly
2. we now have an experimental service on OSGeo that gives precedence to README (unformatted) over README.md: https://git.osgeo.org/gogs/geos/geos
3. The build badges now include jenkins

So maybe we could just use a single README.md and drop the README completely

strk:
If you can rebase your work against trunk, I'll be happy to have a single README.md, as long as it's 80-cols wrapped and nicely readable by a human with a pager :)

strk:
Actually I've done this myself.
Pushed as r4199 = 1352683f32f9e1e513033d03f66693d73b6b4d60 (refs/remotes/trunk)
Thanks Mike!

PR #52, opened 2015-10-01
Add autoDestroy method to GeometryFactory

strk:
Only for heap-allocated factories, can be used to give ownership
of the factory to its Geometry childs. Deletion of last child will
destroy the factory.

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -838,6 +838,12 @@ GEOSGeom_clone(const Geometry *g)
     return GEOSGeom_clone_r( handle, g );
 }
 
+GEOSGeometry *
+GEOSGeom_setPrecision(const GEOSGeometry *g, double gridSize, int forceSnap)
+{
+	return GEOSGeom_setPrecision_r(handle, g, gridSize, forceSnap);
+}
+
 int
 GEOSGeom_getDimensions(const Geometry *g)
 {
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -941,6 +941,26 @@ extern const GEOSGeometry GEOS_DLL *GEOSGetGeometryN_r(
 extern int GEOS_DLL GEOSNormalize_r(GEOSContextHandle_t handle,
                                     GEOSGeometry* g);
 
+/**
+ * Set the geometry's precision, optionally rounding all its
+ * coordinates to the precision grid (if it changes).
+ *
+ * @param gridSize size of the precision grid, or 0 for FLOATING
+ *                 precision.
+ * @param forceSnap specifies whether snapping to the precision grid
+ *                  should be forced: -1 prevents snapping, which can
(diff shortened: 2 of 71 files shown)


strk:
problem: in C++ you can store a GeometryFactory in other places, like in a WKBReader, so "autoDestroy" is too a generic term, and also "Geometry" is not the only possible "child"...

strk:
another problem is that (the reason why travis failed) you must now be very very sure to not delete a GeometryFactory before all its child are deleted (it happened in the capi::GEOSPreparedGeometry test, due to a factory stack-allocated in a function body). This one I'm about to push a fix for.

manisandro:
What about preventing the user from explicitly invoking the GeometryFactory destructor, but instead offer a `destroy` method which ensures it is only deleted when it is safe to? Obvious downside is the API break.

strk:
yeah, I was thinking about that. if we have to break the API we can also at the same time prevent stack-allocation of a GeometryFactory. Being able to break the C++ API was the whole point of using the C-API btw, so not a huge deal to do that...

strk:
Here's an (old) list of GEOS clients, with specificatino of which one uses the C and which the C++ API: 
https://trac.osgeo.org/geos/wiki/Applications

manisandro:
FWIW, I'd say developers would be happier to do a trivial API port then debugging memory corruption issues caused by incorrect destruction order.

strk:
I've pushed the GEOSGeom_setPrecision interface.
While doing it, I realized we might be simplifying things too much because GeometryPrecisionReducer has some more options you can set, like whether or not you want to preserve topology and whether or not you want to remove collapsed elements.

strk:
I'm thinking it might be time to require a c++11 capable compiler and switch to shared_ptr

strk:
@mloskot I know you'd like this move, won't you ?

mloskot:
@strk It's a good move!

If you think of C++11 switch, `std::unique_ptr` is close replacement for `std::auto_ptr`.
In fact, the latter has been deprecated now.

strk:
Thanks @mloskot -- I'm doing the tedious task of adapting all unit tests after the API change. Terrible ! I feel the pain of C++ API users (well, they've been warned, right?)

strk:
Here we go. You cannot stack-allocate a GeometryFactory now, but you have to use its "create" methods to obtain a smart pointer. You cannot delete it also, the smart pointer will take care of that, ensuring it won't be deleted as long as Geometry childs exist.

It would be nice to further tweak this by adding more potential childs (more classes can hold references to a GeometryFactory) but it took already a long time to make this step.

This PR includes, in addition to the memory management of GeometryFactory, also a commit that exposes a GEOSGeom_setPrecision call. It should probably be moved in a separate thread.

strk:
So the memory management part was committed as r4091, while the PrecisionModel part continues in #53

PR #53, opened 2015-10-02
Add GEOSGeom_setPrecision function to C-API

strk:
Proposal for the ability to set the precision model for geometries
\cc @manisandro

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,7 +2,9 @@ Changes in 3.6.0
 2015-MM-DD
 
 - New things:
+  - CAPI: GEOSGeom_setPrecision (#713) - PHP: Geometry->setPrecision
 - Improvements:
+  - ...
 - C++ API changes:
   - Automatic memory management for GeometryFactory objects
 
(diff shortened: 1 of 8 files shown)


strk:
What I'm not convinced about in this proposal is that it is not flexible enough to expose a few configurations of the precision reducer, namely:

 - It enforces topology preservation in case of snapping
 - It enforces removal of collapsed elements

In the C++ API those two aspects are configurable.

A function with all configurations available could be:

 GEOSGeom_reduce(double gridsize, bool changepm, bool retaintopology, bool removecollapsed)

Then we'd miss the possibility to just change the PM w/out snapping, but that's many not really something we want to allow ?

manisandro:
Personally I don't see a scenario where you want to change the precision model of a geometry without snapping, that would be rather confusing IMO.

strk:
Alright, force-pushed a new version using flags. Supports all that GeometryPrecisionReducer supports. See how you like it. I'm writing a unit test now, with the crossing lines intersection...

strk:
There might possibly be a case for snapping w/out changing the PrecisionModel though (but with this API it can be added w/out interface change)

strk:
Committed to trunk: https://trac.osgeo.org/geos/changeset/4092

PR #56, opened 2015-11-29
Fix incorrect logic in Vertex::classify

nyalldawson:
As identified by a Coverity scan over GEOS, Vertex::classify always returns either DESTINATION or BETWEEN due to incorrect logic. 

This PR fixes the logic (restoring it to JTS behaviour), and also syncs Vertex.cpp/.h to r705 (brings in documentation improvements)

### DIFF
--- a/include/geos/triangulate/quadedge/Vertex.h
+++ b/include/geos/triangulate/quadedge/Vertex.h
@@ -12,7 +12,7 @@
  *
  **********************************************************************
  *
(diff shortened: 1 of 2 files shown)


strk:
Does JTS come with any unit test for this class ?
It's dangerous to commit fixes with no guarding tests..

Anyway, this is in as of r4121

nyalldawson:
@strk No, no tests there unfortunately.

Thanks for pushing this!

PR #63, opened 2016-04-12
Match .h declaration parameter name with .cpp definitiion.

snibble:
The Doxygen docs are a little confusing here because the parameter name doesn't match. This minor touch-up clears things up a little.


strk:
Thanks, pushed as 
r4182 = 654b92ee4a85f050772487990c2de555c4fd9593 (refs/remotes/trunk)
and
r4183 = ed75db63b6d794829bddf830fb5d9e66a009cefe (refs/remotes/trunk)

PR #68, opened 2016-09-02
check if geos_svn_revision.h exists

rkanavath:
GOES_BUILD_PACKAGED is used to decide wheather to generate a geos_svn_revision.h exists or not. This is to build fine when building from git clone or source download. 

This PR avoid additional cmake variable GEOS_BUILD_PACKAGED and test if the file in question exists in source or binary directory.


strk:
r4255 = 435a8d15e6ac18b2121df65d412738ebabaa6e51 (refs/remotes/trunk)

PR #69, opened 2016-09-02
Set/Get Geometry User Data in CAPI

gtourkas:
need to associate an numerical id with a set of geometry and build an SRTree to identify the geometries returned by GEOSSTRtree_nearest and GEOSSTRtree_query; Geometry's get/setUserData are appropriate for such a task but there are no wrappers in the CAPI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,12 @@ if(NOT MSVC)
     "Set to ON|OFF (default) to build GEOS with assert() macro enabled" OFF)
 endif()
 
+option(GEOS_BUILD_STATIC
+  "Set to OFF|ON (default) to build GEOS static libraries" ON)
+
+option(GEOS_BUILD_SHARED
+  "Set to OFF|ON (default) to build GEOS shared libraries" ON)
+
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
   option(GEOS_ENABLE_FLOATSTORE
     "Set to OFF|ON (default) to control IEEE754 conformance and remove extra precision" ON)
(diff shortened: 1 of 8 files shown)


strk:
If I'm not wrong the STRtree interface allows you to store
an opaque pointer as the value (while storing a Geometry as the key)
so you should be able to put your numerical ids into a larger
structure embedding a Geometry

gtourkas:
@strk yes, one can use the GEOSSTRtree_nearest_generic but will also need to specify the distance function, which might be straightforward but also not so slick 

given that the geometry is providing a set/getUserData i thought a set of wrapper functions won't harm

gtourkas:
@strk yes, one can use the GEOSSTRtree_nearest_generic but will also need to specify the distance function, which might be straightforward but also not so slick

given that the geometry is providing a set/getUserData i thought a set of wrapper functions won't harm

strk:
Agree those functions won't harm.
Could you add a unit test for them under tests/unit/capi ?

gtourkas:
tests committed

strk:
Please:

1. Rename the methods to GEOSGeom_setUserData / GEOSGeom_getUserData
2. Enable the new testcase in the test/unit/Makefile.am (make sure "make check" runs the test)
3. Squash-rebase to trunk

gtourkas:
@strk done

strk:
tests/unit/Makefile.am still doesn't have an entry for
capi/GEOSUserDataTest.cpp

gtourkas:
@strk it is - check the 5th section in the single push
screen shot 2016-09-06 at 10 44 29

strk:
The function names are still not changed to be GEOSGeom_xxxx

strk:
r4254 = 2b2645fd31cf881dbed0d2a1bc3f0013a57aa504 (refs/remotes/trunk)

PR #70, opened 2016-09-27
Wrong GEOS_DLL position

bartoli:
Should not be type* GEOS_DLL function(), but type GEOS_DLL *function(). Doesn't build on windows before the modif


strk:
thanks

PR #76, opened 2017-01-10
fixes for MSVC compilation

aescande:
missing GEOS_DLL and USE_MATH_DEFINES (M_PI is not standard C++)

### DIFF
--- a/tests/unit/capi/GEOSDistanceTest.cpp
+++ b/tests/unit/capi/GEOSDistanceTest.cpp
@@ -10,6 +10,9 @@
 #include <cstdlib>
 #include <memory>
 #include <math.h>
+#ifndef M_PI
+#define M_PI       3.14159265358979323846
+#endif
 
 namespace tut
 {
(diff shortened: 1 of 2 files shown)


tests/unit/capi/GEOSDistanceTest.cpp
```diff
@@ -9,6 +9,9 @@
 #include <cstdio>
 #include <cstdlib>
 #include <memory>
+#if defined(_MSC_VER)
+#  define _USE_MATH_DEFINES
+#endif
 #include <math.h>
```
strk: In other files we have this block:
```
#ifndef M_PI
#define M_PI        3.14159265358979323846
#endif
```

Please use the same (BTW, one such block is in platform.h.vc)

aescande:
Update to take into account the M_PI remark

strk:
Thanks, committed
r4325 = bd1b03ef5fe049ba43acbd53b136754bfd8a6043 (refs/remotes/trunk)

PR #80, opened 2017-04-06
[CMake] Simplify generation of geos_revision.h

mloskot:
Remove `tools/geos_revision_cmake.h.in` as no longer required - `geos_revision.h` is generated on-fly.

Add `geos_revision` target set as dependency of `geos_c` target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,62 +252,27 @@ include_directories(${PROJECT_BINARY_DIR})
 #################################################################################
 # Setup checks and generate config headers
 #################################################################################
+find_package(Git)
 
-#################################################################################
-#  MACRO: GET_SVN_REVISION
-#
-#  DESCRIPTION:
-#      MACRO FOR GETTING THE SVN revision for this build
-#################################################################################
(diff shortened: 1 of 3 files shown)


mloskot:
@strk Did you intentionally close it or it happened automagically?

strk:
I didn't do anything here, so I guess it was automatic
(would be, if you push the exact same commit to a branch)

mloskot:
@strk 
> I didn't do anything here, so I guess it was automatic 

Surprising behaviour.

> (would be, if you push the exact same commit to a branch)

It shouldn't happen w. `git push -f` which is a common way to update a PR

mloskot:
I suspect, `git push -f` behaviour is different if issued against upstream vs non-upstream repository

strk:
Oh wait. My script does a `push --mirror` to github, basically destroying any github branch which is not in the upstream. You should put your patches into your own fork to avoid feature branches destruction.

mloskot:
OK, that means, `OSGeo/geos` forks must be used to work via PRs

strk:
confirmed

PR #81, opened 2017-04-07
[CMake] Simplify generation of geos_revision.h

mloskot:
Remove `tools/geos_revision_cmake.h.in` as no longer required - `geos_revision.h` is generated on-fly.

Add `geos_revision` target set as dependency of `geos_c` target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,62 +252,27 @@ include_directories(${PROJECT_BINARY_DIR})
 #################################################################################
 # Setup checks and generate config headers
 #################################################################################
+find_package(Git)
 
-#################################################################################
-#  MACRO: GET_SVN_REVISION
-#
-#  DESCRIPTION:
-#      MACRO FOR GETTING THE SVN revision for this build
-#################################################################################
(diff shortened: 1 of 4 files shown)


strk:
Remember to push in upstream, not here.
If you find a way, we should ideally *forbid* merging any PR with the button here (I suspect there's a way but don't want to pollute my brain into learning use of proprietary tools...)

mloskot:
@strk 
> Remember to push in upstream, not here.

Yes, that's clear to me.

> we should ideally forbid merging any PR with the button here

I'll check that https://trac.osgeo.org/geos/ticket/823

hobu:
> we should ideally forbid merging any PR with the button here

It's like you purposefully make it as hard as possible for people to contribute.

strk:
On Sun, Apr 09, 2017 at 09:42:15PM -0700, Howard Butler wrote:
> > we should ideally forbid merging any PR with the button here
> 
> It's like you purposefully make it as hard as possible for people to contribute.

Contributing has never implied pushing a button.
It's just eventually harder for *mergers*

--strk;

PR #82, opened 2017-04-07
Remove trailing blanks, add check step to keep code free of them

strk:
@mloskot see how you like this


strk:
See also how Dronie likes it: https://git.osgeo.org/gogs/geos/geos/pulls/2

mloskot:
@strk
> @mloskot see how you like this

If you ask me, it doesn't solve any of my problems.
It doesn't unify tabs/spaces, indentation, brackets placement, and `<here long list>`.
Besides, check step is covered by only one of two official build configurations.

You know my opinion, `.editorconfig` is no improvement to the current situation.
It is as just a small helper that has very little to do with formatting style of a complex source code.

However, I don't want to interfere, do what you like.

PR #83, opened 2017-04-11
Add explicit to StringTokenizer ctor and move more to the initializer list in ctor

schwehr:
Just minor cleanup.  A little background on why I made this patch:

I am working on a fuzzer for WktReader and ran into ASAN detecting leak troubles very quickly with things like `MULTIPOLYGON  (`.  In trying to work on WktReader, I wanted to make sure I understand StringTokenizer.  In writing a test, I was passing `""` to the constructor, which created a temp object which was destroyed after the ctor resulting in a use-after-free in my test.

And might as well move everything to the constructor list that I can while I'm looking at the StringTokenizer ctor.

geos/tests/io/StringTokenizer_test.cc:

https://github.com/schwehr/gdal-autotest2/commit/dbe3d755f6137098b1f52edda40c980979685f27

### DIFF
--- a/include/geos/io/StringTokenizer.h
+++ b/include/geos/io/StringTokenizer.h
@@ -41,7 +41,7 @@ public:
 		TT_WORD
 	};
 	//StringTokenizer();
-	StringTokenizer(const std::string& txt);
+	explicit StringTokenizer(const std::string& txt);
 	~StringTokenizer() {}
 	int nextToken();
 	int peekNextToken();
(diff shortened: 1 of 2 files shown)


strk:
Thanks, merged as commit f825b41004b37fcb0dd4ea76f22b5b6aa447b2ac

strk:
@schwehr for some reason Appveyor is now reporting a failure after my commit, can you see why ?
Reported in https://trac.osgeo.org/geos/ticket/831

PR #85, opened 2017-06-22
ENH: add GenerateSourceGroups.cmake to EXTRA_DIST

grizonnetm:
Allow to compile geos source tarball using cmake.

See discussion here: https://trac.osgeo.org/geos/ticket/753#comment:22


strk:
I merged this already.
Only now I realize github did not detect it
probably due to https://trac.osgeo.org/geos/ticket/836

PR #86, opened 2017-07-14
Use -mfpmath=387 when available

strk:
Hopefully makes testsuite results consistent.
See https://travis-ci.org/OSGeo/geos/builds/253544101


strk:
Took a different pproach which seems to work (ovrlay result checker

PR #87, opened 2017-07-18
Fix build error with MSVC

szekerest:


mloskot:
I didn't notice this PR before I fixed this bug in upstream repo
https://git.osgeo.org/gogs/geos/geos/commit/6b236e95faf8a3f52d2ee4158d1e7f57829fe42f

PR #91, opened 2018-01-09
Cmake improve

cvvergara:
## Tasks done:

- [x] Using the cmake project Version
- [x] Separating the C flags from the C++ flags
  - ansi in C++ is C++98 and the min requirement is C++11
  - fPIC is for the GEOS C flag
- [x] Added some more flags for compiling

## TODO In additional PR's

- Check flags used for the different kinds of build
- Fix the warnings

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,7 @@ endif()
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
 
   # General options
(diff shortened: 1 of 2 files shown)


cvvergara:
@strk 
Fix the warnings is getting toooooo big
So what if you merge this one first and I will make PR with 9 to 10 commits fixing warnings

cvvergara:
I think this one is already [merged](https://git.osgeo.org/gitea/geos/geos/commit/d43e678578cfc60d02fbb051319d65e638474515), closing to avoid confusion

PR #92, opened 2018-01-10
Fix warnings

cvvergara:
Fixing warnings generated with the new flags

### DIFF
--- a/include/geos/profiler.h
+++ b/include/geos/profiler.h
@@ -80,8 +80,9 @@ public:
 	void stop()
 	{
 		gettimeofday(&stoptime, nullptr);
-		double elapsed = 1000000*(stoptime.tv_sec-starttime.tv_sec)+
-			(stoptime.tv_usec-starttime.tv_usec);
+		double elapsed = static_cast<double>(
+				1000000 * (stoptime.tv_sec - starttime.tv_sec)
+				+ (stoptime.tv_usec - starttime.tv_usec));
 
 		timings.push_back(elapsed);
 		totaltime += elapsed;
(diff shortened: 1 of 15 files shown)


include/geos/profiler.h
```diff
@@ -80,8 +80,9 @@ class GEOS_DLL Profile {
 	void stop()
 	{
 		gettimeofday(&stoptime, nullptr);
-		double elapsed = 1000000*(stoptime.tv_sec-starttime.tv_sec)+
-			(stoptime.tv_usec-starttime.tv_usec);
+		double elapsed = static_cast<double>(
```
strk: Does it fix the warning if you end the literal number with a dot? Or what does the warning say exactly?

cvvergara: The literal is not the problem, the problem is the whole computation.

strk:
> /home/vicky/potree/geos/vicky/include/geos/profiler.h:84:51: warning: conversion to ‘double’ from ‘__time_t {aka long int}’ may alter its value [-Wconversion]
>      1000000. * (stoptime.tv_sec - starttime.tv_sec)

Ok, I see `long int - long int` is converted to double
The message is warning us about possibly being unable to
represent a big number with a double, right ?

Do you get the same warning with this ?

     long int i = 9223372036854775807L;
     double d = i;

A static_cast<> simply hides the problem, right ?
Anyway, we should be able to only cast the (stop-start)
numbers ?

cvvergara:
Yes a static_cast hides the problem, but you make it visible, each time there is one, there is a conversion problem. (that maybe JTS people did not catch)

cvvergara:
Closing due to more knowledge about how things work

PR #97, opened 2018-02-13
Add support for nmake version 14.12.25835.0

cedricdelpas:
Add support for nmake version 14.12.25835.0


mattmakesmaps:
I also ran into this issue. Specific to this version of `nmake`, should the corresponding `GEOS_MSC` value be incremented from `1910` to `1912`?

See table in the following: https://blogs.msdn.microsoft.com/vcblog/2017/11/15/side-by-side-minor-version-msvc-toolsets-in-visual-studio-2017/

cedricdelpas:
You are right. It's a misstake in my commit. Do you want i make the modification in the branch CGX-GROUP:visual_15.5.6 ?

pramsey:
We've deprecated *.vc files so MSVC support is now exclusively via Cmake.

PR #99, opened 2018-05-29
Add GEOSGeom_getXMin and friends

dbaston:
Trac: https://trac.osgeo.org/geos/ticket/871


dbaston:
Travis is failing on some syntax that I believe is valid, but only in CMake builds:

`using geom_ptr= std::unique_ptr<GEOSGeometry, decltype(&GEOSGeom_destroy)>;
`

Is CMake not passing along -std=c++11 ?

strk:
Please reference the ticket in the commit (`closes #871`) and don't forget an entry in the NEWS file. You'll have to check yourself about CMake

PR #100, opened 2018-05-30
Add GEOSReverse to CAPI

dbaston:
I was surprised that the `Polygon` and `MultiPolygon` classes don't provide a `reverse` method. This PR provides a "reverse" operation at the CAPI only. Maybe in the future a virtual `Geometry::reverse` method could be created and some of this code could be removed.


strk:
JTS has it, maybe you could port it:


```
  /**
   * Computes a new geometry which has all component coordinate
   * sequences
   * in reverse order (opposite orientation) to this one.
   *
   * @return a reversed geometry
   */
  public abstract Geometry reverse();
```

strk:
> That is probably the way to go, but is there anything wrong with merging in this implementation in the meantime?

That you'll have to change the implementation after porting,
so if you plan to do the port why not doing it now ?
Doens't sound to be too much work (it's the nice thing of C++
interface being advertised as unstable...)

dbaston:
> Doens't sound to be too much work (it's the nice thing of C++
> interface being advertised as unstable...)

I'm a bit wary, because the virtual method would return a `Geometry*` instead of the specific type, so all existing usages will have to be found and dynamic casts added. I'll look and see.

dbaston:
Looks like it worked out OK.

strk:
Great. If you can update any port info (see top of files) that'd be great. When you do please update both the .cpp and the .h file

dbaston:
I don't think there's anything to update...the port comments reference JTS 1.12, which already had these methods.

strk:
On Thu, May 31, 2018 at 02:32:23PM +0000, Dan Baston wrote:
> I don't think there's anything to update...the port comments reference JTS 1.12, which already had these methods.

In other files I tried to reference exact commit.
Revision IDs from SVN are gone so it would be nice if
we could include something from GIT now (git-describe?)

PR #102, opened 2018-06-07
[CMake] Build the C API into static libs

dakotahawkins:
To support building (e.g., GDAL) against static libs only, add the C API
to the static target.

Otherwise, if building with CMake, the C API is only available in a
dynamic library.

In the future, this could be a lot cleaner -- e.g. with more idiomatic
CMake configuration of shared vs. static builds -- but for the time
being I hope this will be acceptable.

### DIFF
--- a/capi/CMakeLists.txt
+++ b/capi/CMakeLists.txt
@@ -21,8 +21,8 @@ set(geos_c_SOURCES
 
 file(GLOB geos_capi_HEADERS ${CMAKE_BINARY_DIR}/capi/*.h) # fix source_group issue
 
-if(NOT GEOS_ENABLE_MACOSX_FRAMEWORK)
-  # if building OS X framework, CAPI built into C++ library
+if(NOT GEOS_ENABLE_MACOSX_FRAMEWORK AND GEOS_BUILD_SHARED)
+  # if building OS X framework or only building static libs, CAPI built into C++ library
   add_library(geos_c SHARED ${geos_c_SOURCES})
 
   target_link_libraries(geos_c geos)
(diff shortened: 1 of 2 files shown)


dakotahawkins:
Fixes [trac #878](https://trac.osgeo.org/geos/ticket/878).

dbaston:
Any thoughts on this, @strk or @mloskot ?

mloskot:
@dbaston The whole CMake configuration should be rewritten completely. It's an old-school CMake written as quick prototype for convenience of developers who'd like to develop GEOS in IDEs. It used to be marked as *unofficial build setup*, not sure what's its status now.

Regarding @dakotahawkins-s hack, of it works it probably is fine.

dakotahawkins:
@mlsokot According to [this](https://trac.osgeo.org/geos/wiki/BuildingOnWindowsWithCMake):

> Starting at GEOS 3.5, we officially also support CMake. We will happily accept tickets for GEOS 3.5 against CMake. We would appreciate a patch if you have one, but not required..

It does seem like the CMake needs a rewrite, but this capability is a much more immediate need for me personally.

mloskot:
@dakotahawkins It's been years since I looked at that page. Looks like @robe2 updated it.

strk:
Dakota you're welcome if you want to take ownership of that project
(maintainance of CMake infrastructure)

dakotahawkins:
@strk I'm still coming up to speed on it, personally. I didn't mean what I said to sound insulting, apologies!

Personally, I'm glad this project has CMake support at all!

Based on looking through trac issues (two relevant issues [here](https://trac.osgeo.org/geos/ticket/756) and [here](https://trac.osgeo.org/geos/ticket/816), not sure why exactly the second one is closed) it seems like there is at least some consensus on room for improvement.

strk:
I didn't want to be insulting either, I mean it: if you want to
maintain it you are very welcome :)

dakotahawkins:
@strk I will look at it this weekend. The thing that worries me the most is weird MacOS stuff I won't be able to try very easily.

robe2:
Sorry I missed this and I already tagged 3.6.3 and 3.7.0beta2,  I've slated it for 3.6.4

https://trac.osgeo.org/geos/ticket/878

dakotahawkins:
For what it's worth we're using this in production and it's working well.

PR #103, opened 2018-06-07
[CMake] Build the C API into static libs (svn-3.6 backport)

dakotahawkins:
To support building (e.g., GDAL) against static libs only, add the C API
to the static target.

Otherwise, if building with CMake, the C API is only available in a
dynamic library.

In the future, this could be a lot cleaner -- e.g. with more idiomatic
CMake configuration of shared vs. static builds -- but for the time
being I hope this will be acceptable.


dakotahawkins:
Fixes [trac #878](https://trac.osgeo.org/geos/ticket/878).

PR #108, opened 2018-07-22
Polygonize graph

cvvergara:
This PR is for reviewing the process of changing from
java in disguise code to C++ code

### DIFF
--- a/include/geos/operation/polygonize/PolygonizeGraph.h
+++ b/include/geos/operation/polygonize/PolygonizeGraph.h
@@ -116,7 +116,7 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	 *
 	 * @return edgeRingList: the EdgeRings found by the polygonization process.
 	 */
-	std::vector<EdgeRing*> getEdgeRings();
+	std::vector<EdgeRing*> getEdgeRings() const;
 
 	/**
 	 * \brief
@@ -151,7 +151,7 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 
 	planargraph::Node* getNode(const geom::Coordinate& pt);
 
-	void computeNextCWEdges();
+	void computeNextCWEdges() const;
 
 	/**
 	 * \brief
@@ -163,7 +163,7 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	 *
 	 */
 	void convertMaximalToMinimalEdgeRings(
-			const std::vector<PolygonizeDirectedEdge*> &ringEdges);
+			const std::vector<PolygonizeDirectedEdge*> ringEdges) const;
 
 	/**
 	 * \brief
@@ -191,11 +191,11 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	 */
 	std::vector<PolygonizeDirectedEdge*>
 	findLabeledEdgeRings(
-			std::vector<planargraph::DirectedEdge*> &dirEdgesIn);
+			const std::vector<planargraph::DirectedEdge*> dirEdgesIn) const;
 
-	void label(std::vector<planargraph::DirectedEdge*> &dirEdges, long label);
+	void label(const std::vector<planargraph::DirectedEdge*> &dirEdges, long label) const;
 
-	void computeNextCWEdges(planargraph::Node *node);
+	void computeNextCWEdges(planargraph::Node *node) const;
 
 	/**
 	 * \brief
@@ -204,7 +204,7 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	 * This algorithm has the effect of converting maximal edgerings
 	 * into minimal edgerings
 	 */
-	void computeNextCCWEdges(planargraph::Node *node, long label);
+	void computeNextCCWEdges(planargraph::Node *node, long label) const;
 
 	/**
 	 * \brief
@@ -218,8 +218,8 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	std::vector<planargraph::DirectedEdge*>
 	findDirEdgesInRing(PolygonizeDirectedEdge *startDE) const;
 
-	/* not const because has side effect of saving the Edge Ring found */
-	EdgeRing* findEdgeRing(PolygonizeDirectedEdge *startDE);
+	/* has side effect of saving the Edge Ring found */
+	EdgeRing* findEdgeRing(PolygonizeDirectedEdge *startDE) const;
 
 	/*
 	 * Data members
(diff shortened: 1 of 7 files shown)


include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -67,20 +69,18 @@ namespace polygonize { // geos::operation::polygonize
...
-
-public:
-
+ public:
 	/**
 	 * \brief
 	 * Deletes all edges at a node
 	 */
-	static void deleteAllEdges(planargraph::Node *node);
+	void deleteAllEdges(planargraph::Node *node);
```
strk: Was this intentional ? Was it not a static method in JTS ?

include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -130,11 +130,10 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
...
 	void deleteDangles(std::vector<const geom::LineString*> &dangleLines);
 
-private:
-
-	static int getDegreeNonDeleted(planargraph::Node *node);
+ private:
+	int getDegreeNonDeleted(planargraph::Node *node) const;
 
-	static int getDegree(planargraph::Node *node, long label);
+	int getDegree(planargraph::Node *node, long label) const;
```
strk: Were these intentional ? Changed from static to non-static methods

include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -152,7 +151,7 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	 *
 	 */
 	void convertMaximalToMinimalEdgeRings(
-			std::vector<PolygonizeDirectedEdge*> &ringEdges);
+			std::vector<PolygonizeDirectedEdge*> ringEdges);
```
strk: Why copying the vector rather than passing a reference ?

include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -164,9 +163,10 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
 	 * @param intNodes : intersection nodes found will be pushed here
 	 *                   the vector won't be cleared before pushing.
 	 */
-	static void findIntersectionNodes( PolygonizeDirectedEdge *startDE,
-			long label, std::vector<planargraph::Node*>& intNodes
-);
+	std::vector<planargraph::Node*>
+	findIntersectionNodes(
+		 	PolygonizeDirectedEdge *startDE,
+			long label);
```
strk: Why not static ?

include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -175,15 +175,15 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
...
 	 * @param dirEdgesIn  a list of the DirectedEdges in the graph
-	 * @param dirEdgesOut each ring found will be pushed here
+	 * @result vector that contains each ring found
 	 */
-	static void findLabeledEdgeRings(
-			std::vector<planargraph::DirectedEdge*> &dirEdgesIn,
-			std::vector<PolygonizeDirectedEdge*> &dirEdgesOut);
+	std::vector<PolygonizeDirectedEdge*>
+	findLabeledEdgeRings(
+			std::vector<planargraph::DirectedEdge*> &dirEdgesIn);
```
strk: The output parameter was used to avoid a vector copy, passing a reference to an existing vector allows appending to a pre-allocated vector (for example)

include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -175,15 +175,15 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
...
 
-	static void label(std::vector<planargraph::DirectedEdge*> &dirEdges, long label);
+	void label(std::vector<planargraph::DirectedEdge*> &dirEdges, long label);
 
-	static void computeNextCWEdges(planargraph::Node *node);
+	void computeNextCWEdges(planargraph::Node *node);
```
strk: why dropping the static keyword ?

include/geos/operation/polygonize/PolygonizeGraph.h
```diff
@@ -201,28 +201,30 @@ class GEOS_DLL PolygonizeGraph: public planargraph::PlanarGraph {
...
 	 *
 	 * @param startDE the DirectedEdge to start traversing at
-	 * @param edgesInRing : the DirectedEdges that form a ring will
-	 *                      be pushed here.
+	 * @result a vector of the DirectedEdge that form a ring
 	 */
-	static void findDirEdgesInRing(PolygonizeDirectedEdge *startDE,
-		std::vector<planargraph::DirectedEdge*>& edgesInRing);
+	std::vector<planargraph::DirectedEdge*>
+	findDirEdgesInRing(PolygonizeDirectedEdge *startDE) const;
```
strk: This forces a new allocation of a vector, and a copy of it, whereas the output parameter avoids that

include/geos/operation/polygonize/PolygonizeGraph.h
cvvergara: Intentional

include/geos/operation/polygonize/PolygonizeGraph.h
cvvergara: I havent finished, I need to add const to some parameters, so that it uses the move constructor of the vector

include/geos/operation/polygonize/PolygonizeGraph.h
cvvergara: Explained before

include/geos/operation/polygonize/PolygonizeGraph.h
cvvergara: it is using the move constructor,
you actually want that on the parameter if you are going to "continue filling" the vector

include/geos/operation/polygonize/PolygonizeGraph.h
cvvergara: I can make the move explicit if you want

cvvergara:
C++11 or C++98?

cvvergara:
FYI:
 
```
std::vector<A> func() {
    std::vector<A> v;
    /* fill v */
    return v; // 'v' is converted to r-value and return value is move constructed.
}
```

Copy elision is defined in the [C++ standard](http://open-std.org/JTC1/SC22/WG21/docs/papers/2015/n4567.pdf) in:
12.8  Copying and moving class objects [class.copy]

> (31.1) — in a return statement in a function with a class return type, when the expression is the name of a non-volatile automatic object (other than a function or catch-clause parameter) with the same cvunqualified type as the function return type, the copy/move operation can be omitted by constructing the automatic object directly into the function’s return value

strk:
> The function is only used internally in the class, so no need of static

I think non-static functions take an implicit argument ("this") which
adds a (minimal) overhead to any call to it.

Are them static on he JTS side ?
Does it do any harm for them to remain static ?

strk:
> > (31.1) — in a return statement in a function with a class return type, when the expression is the name of a non-volatile automatic object (other than a function or catch-clause parameter) with the same cvunqualified type as the function return type, the copy/move operation can be omitted by constructing the automatic object directly into the function’s return value

Does "can be omitted" mean it is implementation-specific what will
really be done ?

cvvergara:
Sorry, I might have missed the point of the language that its being used: 
Java, C++98, C++11

include/geos/operation/polygonize/PolygonizeGraph.h
dbaston: Overhead arguments aside, I think keeping `static` shows intent more clearly.

cvvergara:
Closing as this I consider an experiment

PR #96, opened 2018-02-01
Update RectangleIntersectionTest.cpp

Komzpa:
Trac: https://trac.osgeo.org/geos/ticket/865

PR #109, opened 2018-07-27
Add test with infinite loop using Postgis input

Algunenano:
Same as https://github.com/libgeos/geos/pull/96 but using bigger precision.

Trac issue: https://trac.osgeo.org/geos/ticket/865


dbaston:
Superseded by #110

PR #110, opened 2018-07-28
Fix infinite loop in GEOSClipByRect (#865)

dbaston:
Infinite loop was caused by floating point error in the intersection between a near-vertical or near-horizontal line segment and an edge of the rectangle. This PR corrects for the special case and adds a guard to throw an exception instead of entering an infinite loop should a similar situation arise.


strk:
A promise of GEOSClipByRect is that it will prefer giving a not-so-perfect result than throw an exception, are we breaking that promise with this change ?

Algunenano:
I've tested the Postgis code that used to loop infinitely against this branch and it's fixed; no exceptions thrown to Postgis code.

There are some differences in the tests (GEOSClipByRect seems less precise), but I understand that's the intended behaviour.

dbaston:
> A promise of GEOSClipByRect is that it will prefer giving a not-so-perfect result than throw an exception

Yes, but the alternative of an infinite loop is clearly worse. I can modify the PR so that instead of throwing an exception, the INSIDE/OUTSIDE bits are ignored, which would also avoid the infinite loop. I'm not sure what would happen downstream of this point, though.

PR #115, opened 2018-08-10
Fix the user data of geometry in voronoi diagram.

whuaegeanse:
(1) Use a heap variable pointer as user data of  polygon of voronoi diagram;
(2) Set user data for polygon of voronoi diagram.

### DIFF
--- a/src/triangulate/VoronoiDiagramBuilder.cpp
+++ b/src/triangulate/VoronoiDiagramBuilder.cpp
@@ -1,4 +1,4 @@
-/**********************************************************************
+﻿/**********************************************************************
  *
  * GEOS - Geometry Engine Open Source
  * http://geos.osgeo.org
@@ -133,6 +133,7 @@ VoronoiDiagramBuilder::clipGeometryCollection(const geom::GeometryCollection& ge
 		{
 			result.reset( g->clone() );
       // TODO: check if userData is correctly cloned here?
+			result->setUserData(((Geometry*)g)->getUserData());
 		}
 		else if(clipEnv.intersects(g->getEnvelopeInternal()))
 		{
--- a/src/triangulate/quadedge/QuadEdgeSubdivision.cpp
+++ b/src/triangulate/quadedge/QuadEdgeSubdivision.cpp
@@ -1,4 +1,4 @@
-/**********************************************************************
+﻿/**********************************************************************
  *
  * GEOS - Geometry Engine Open Source
  * http://geos.osgeo.org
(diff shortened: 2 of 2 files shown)


whuaegeanse:
OK, I will add one.

kenash54845:
and memory management will be  ugly:
Coordinate *pcoord = (Coordinate *)polys->getGeometryN(i)->getUserData();
delete pcoord;

kenash54845:
libgeos wont fix this,i guess

whuaegeanse:
@kenash54845 I will add a test for this and have to modify some other tests.

kenash54845:
@whuaegeanse  have u tested whether userdata is null or not? call of   `setUserData`  should not be there . i think it should be: 
 \src\triangulate\VoronoiDiagramBuilder.cpp   

```
		if(clipEnv.contains(g->getEnvelopeInternal()))
		{
			result.reset( g->clone() );
      // TODO: check if userData is correctly cloned here?
		}
		else if(clipEnv.intersects(g->getEnvelopeInternal()))
		{
			result.reset( clipPoly->intersection(g) );
			
		}

		if(result.get() && !result->isEmpty() )
		{			result->setUserData(((Geometry*)g)->getUserData()); // TODO: needed ?

			clipped->push_back(result.release());
		}
```

pramsey:
OK, I see two PRs for this, which one is going forward? 
In Java world the ownership of the user data is not really an issue as GC will clean it up once orphaned.
In the GEOS case we seem to be counting on the setter of the userData to retain ownership and clean it up when finished. That model breaks down when we start setting userData ourselves behind the backs of the users.
This "fix" seems like it'll open up a bunch of problems to end users, who will be leaking memory all over the place. 
We could change the _userData to be a unique_ptr, but then we have a new problem for all the existing users of _userData who get their user data cleaned up when the geometry is deleted.
The whole idea of a void pointer feels blunt force, a map would be so much nicer... but even then the question of ownership arises, unless the _userData remains a user owned and a new map of geos-added flavourings that are managed by geos is added to hold stuff like this.

dbaston:
Returning the coordinate through `userData` seems like a hack in the first place. Maybe the `VoronoiDiagramBuilder` could be modified to produce something like `std::vector<std::pair<Geometry*, Coordinate>>` instead?

pramsey:
Could do. Still leaves us w/ the C API to deal with, in case anyone is hoping to get access to those coordinates through that route...

pramsey:
I think this is a worthy goal, but this approach isn't it. Something that uses @dbaston's <std::pair> return idea and also figures out how to expose that via the CAPI would be a complete solution.

PR #117, opened 2018-09-01
Fix Delaunay robustness by using longer floating point variable

Komzpa:
GEOS ticket: https://trac.osgeo.org/geos/ticket/922#ticket

PostGIS ticket: https://trac.osgeo.org/postgis/ticket/4152


dbaston:
Long double is 64 bits in MSVC; I think you need to ifdef out the test.

Komzpa:
@dbaston is there a chance to get it into 3.7.0 and backpatch to 3.6 before release of 3.7? :)

dbaston:
I think so. You said on the JTS ticket that you had reduced the failure case to a 5-point input; can you use that here?

dbaston:
@Komzpa ^^^

dbaston:
@Komzpa Great! Can you please squash the commits and force-push?

Komzpa:
done. isn't there squash-and-merge button on github PR merge though, why is it manual?

dbaston:
This is just a mirror of the GEOS repository.

PR #118, opened 2018-09-14
Voronoi Diagram:The userData of each polygon is NOT set to be the Coordinate  of the cell site

kenash54845:
hi,

 QuadEdgeSubdivision.h  line 400, i read
```
	/**
	 * Gets the cells in the Voronoi diagram for this triangulation.
	 * The cells are returned as a {@link GeometryCollection} of {@link Polygon}s
	 * The userData of each polygon is set to be the {@link Coordinate}
	 * of the cell site.  This allows easily associating external
	 * data associated with the sites to the cells.
	 *
	 * @param geomFact a geometry factory
	 * @return a GeometryCollection of Polygons
	 */
	std::unique_ptr<geom::GeometryCollection> getVoronoiDiagram(const geom::GeometryFactory& geomFact); 
```

The userData of each polygon is set,But different from the expected value.
 do as promised,please
modified 2 files and add a test

### DIFF
--- a/src/triangulate/VoronoiDiagramBuilder.cpp
+++ b/src/triangulate/VoronoiDiagramBuilder.cpp
@@ -137,11 +137,11 @@ VoronoiDiagramBuilder::clipGeometryCollection(const geom::GeometryCollection& ge
 		else if(clipEnv.intersects(g->getEnvelopeInternal()))
 		{
 			result.reset( clipPoly->intersection(g) );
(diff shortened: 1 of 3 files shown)


pramsey:
See #115 for more commentary, but I'm increasingly of the opinion that we cannot code to this contract, we have to remove the contract (providing the coordinates of the cell in user data). https://lists.osgeo.org/pipermail/geos-devel/2019-January/008801.html

kenash54845:
ok,close it then

PR #122, opened 2018-09-18
Improve PreparedGeometry performance by minimizing heap allocations

dbaston:
This PR improves the performance of PreparedGeometries by minimizing the number of heap allocations used in `IndexedPointInAreaLocator`. In the `master` branch, `IndexedPointInAreaLocator` individually heap-allocates every `LineSegment` in a geometry; in this PR, the `LineSegments` are stored as values in a `std::vector`. (`IndexedPointInAreaLocator` is non-copyable.)

Motivation for this is seeing `GEOSPreparedGeom_destroy` at 12% of my total program runtime; this PR brings it down to 5%. There is also an efficiency boost on the construction side.

Feedback would be appreciated.

### DIFF
--- a/include/geos/algorithm/locate/IndexedPointInAreaLocator.h
+++ b/include/geos/algorithm/locate/IndexedPointInAreaLocator.h
@@ -4,6 +4,7 @@
  * http://geos.osgeo.org
  *
  * Copyright (C) 2006 Refractions Research Inc.
+ * Copyright (C) 2018 Daniel Baston <dbaston@gmail.com>
  *
  * This is free software; you can redistribute and/or modify it under
  * the terms of the GNU Lesser General Public Licence as published
@@ -18,7 +19,9 @@
 
 #include <geos/algorithm/locate/PointOnGeometryLocator.h> // inherited
 #include <geos/index/ItemVisitor.h> // inherited
+#include <geos/index/intervalrtree/SortedPackedIntervalRTree.h> // inherited
 
+#include <memory>
 #include <vector> // composition
 
 namespace geos {
@@ -30,11 +33,6 @@ namespace geos {
 		class Coordinate;
 		class CoordinateSequence;
 		class LineSegment;
-	}
-	namespace index {
-		namespace intervalrtree {
-			class SortedPackedIntervalRTree;
-		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dbaston:
I was able to make similar improvements in `SortPackedIntervalRTree` using the same technique. The general pattern is this:

1) We have `N` objects that may need to refer to each other using pointers (e.g, tree nodes)
2) Rather than dynamically allocating `N` objects using `new`, create a `std::vector` and reserve size `N`. Insert the objects into the vector using `emplace`.
3) Allow the objects to point to each other using their addresses within the vector. Since we called `reserve` on the vector after creating it, the addresses will be stable as long as we do not add additional objects to the vector.

Are there downsides to this pattern? (For one, it would be nice to have a way to "lock" the vector to prevent an operation that would invalidate the addresses.)

cvvergara:
I can see you are dealing a lot with the pointer stuff +1
I also see you are using `const` where things are expected to not change +1
With `emplace_back`, indeed the ownership is ours.

Maybe you can have a look at this code snippet idea. ( I am thinking more on the vector<geometry *> kind of vector where I see it might be usefull. But I am taking from this PR the segments vector as example)
```
template<class T> class Vector_ref {
public:
  Vector_ref(){}
  // destructor deletes every object that was passed as pointer
  ~Vector_ref { 
    for (auto o : owned) { delete o; }
  }
 // vector like operations:
 // if an object is passed as reference, it is assumed to be owned by the caller,
 // so the caller is responsible for its life time
  void push back(T& ref) {v.push_back(&ref);}
 // if an object is passed as pointer, it is assumed to be allocated by new and
 // its Vector_ref responsability to delete it
  void push back(T* ptr) {v.push_back(ptr); owned.push_back(ptr)}
 // accessor
  T& operator[](size_t i) {return  *v[i];}
  const T& operator[](size_t i) const { return *v[i];}
 // etc .. begin() end() for range loops
private:
  // store elements as a pointers (not as a copy of the object) then you can put for example
  // a Polygon into a Vector_ref<geometry> without suffering slicing
  vector<T*> v;
  vector<T*> owned;
}

..
void foo() {
Vector_ref< geom::LineSegment > segments;
// for this one the Vector_ref is the owner
segments.push_back(new geom::LineSegment(<parameters>));

// for this one foo is the owner
auto ls = new geom::LineSegment(<parameters>);
segments.push_back(*ls);
// for this one also foo is the owner
auto ls1 = geom::LineSegment(<parameters>);
segments.push_back(ls1);

// deleting what foo owns
delete ls;
// on exit of foo() segments clears everything it owns
// on exist ls1 gets cleared
}
...
	
End of idea.

dbaston:
Thanks, @cvvergara . Yes, the ownership logic gets tricky on vectors of Geometries, that ultimately get converted into GeometryCollections, etc. In this case I think the ownership is pretty simple, and I'm proposing to store values instead of pointers to reduce fragmentation and the number of calls to malloc/free.

cvvergara:
Well, the use of unique_ptr  that you are making its great
or shared_ptr when needed is also good

I just added that snipet for you to think about it, as I mentioned, makes more sense on geometry than on linestring, but I added it as example, because linestring is much much simplre than geometry

PR #124, opened 2018-09-20
Fix memory leak in getNoded

Algunenano:
Reference: https://trac.osgeo.org/geos/ticket/923


cvvergara:
@Algunenano 
Thanks for the PR
@strk @dbaston Looks good to me

PR #126, opened 2018-09-21
Fix #926: Fixed crash in GEOSInterpolate() when used with empty LineString.

sir-sigurd:
https://trac.osgeo.org/geos/ticket/926


dbaston:
@sir-sigurd semantics of empty geometries are always tricky. Why did you choose to have `GEOSInterpolate` return an empty point instead of `NULL` in this case? Is this what it does for other invalid cases such as non-linear inputs?

dbaston:
Sounds logical to me. Any objections, @strk or others?

PR #127, opened 2018-09-21
Fix #927 -- Fixed crash in GEOSCoordSeq_isCCW() when used with empty coordseq.

sir-sigurd:
https://trac.osgeo.org/geos/ticket/927


dbaston:
Thanks for the fix!

PR #135, opened 2018-11-29
Avoid passing member variable to member function

dbaston:
@pramsey I went in to fix a new `-Wshadow` warning and came across this. Let me know if I'm missing anything with your intent here.


dbaston:
I see that `geomFact` is passed as an argument in the JTS version, but the method is static there. I guess an alternative change would be to make the GEOS method static.

pramsey:
No intent, just flailing wildly. +1

PR #137, opened 2018-12-04
Fix mem leaks in SIRtreePointInRing

dbaston:
Described in https://trac.osgeo.org/geos/ticket/944#ticket


dbaston:
Merged as  24652fb

PR #138, opened 2018-12-04
Performance improvement

doskabouter:
First check envelope before calling CGAlgorithms::isPointInRing

Performance of st_intersection went from 15.8 to 11.5 seconds for these polygons:
[Polygons.zip](https://github.com/libgeos/geos/files/2646124/Polygons.zip)

It can be improved even more, but not familiar enough with the code to do that right now.
When calculating the st_intersection, there were still 8526 calls to isPointInRing for a polygon with 5 points, and 4262 calls to isPointInRing for a polygon with 130327 points so perhaps a IndexedPointInAreaLocator or a SIRtreePointInRing can be added to the OverlayOp::computeLabelling() so that it can be used in EdgeEndStar::getLocation


doskabouter:
Accidentally added more commits...

PR #143, opened 2018-12-20
Move version numbers to a single file

pramsey:


CMakeLists.txt
```diff
@@ -23,9 +23,7 @@ endif()
 #################################################################################
 # Set GEOS project
 #################################################################################
-project(GEOS VERSION 3.8.0 LANGUAGES C CXX)
-# OPTIONS: "", "dev", "rc1" etc.
-SET(GEOS_PATCH_WORD "dev")
+project(GEOS LANGUAGES C CXX)
```
dbaston: Is it possible to read the file before defining the project so that the project version can still be set?

pramsey: Not sure, I guess I was too intimidated to try...

PR #144, opened 2018-12-22
Improve orientationIndex performance with fast filter

dbaston:
This PR adds the code path to use the (already-implemented ) orientationIndex filter. Benchmarked by running `TopologyPreservingSimplifier` on the GADM boundary of Canada; runtime reduced 50% from 443 seconds to 212 seconds. (Takes 193s with previous, non-robust orientationIndex)

@pramsey

### DIFF
--- a/src/algorithm/CGAlgorithmsDD.cpp
+++ b/src/algorithm/CGAlgorithmsDD.cpp
@@ -27,7 +27,11 @@ using namespace geos::algorithm;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


dbaston:
It's been in JTS since 2012: https://github.com/dbaston/jts-svn-history/commit/b229e8496f8e056f8d5af5109c4297b8d37d68f0

The filter function was already defined in GEOS but not called.

pramsey:
👍 then

PR #145, opened 2018-12-29
Add envelope tests for relate with multigeometries

doskabouter:
Not sure if there could be something gained by switching collection1 and collection2 in https://github.com/doskabouter/geos/blob/aec9825e1673a41ff0f6fc3a32a752c7da267669/src/operation/relate/RelateComputer.cpp#L93-L101 based on some calculation on number of parts and/or their envelopes


doskabouter:
Perhaps the added part can be in a separate function (to avoid duplicate code).
Any suggestions for a good place to add it?

doskabouter:
sounds like a plan. I'll leave out the static if you don't mind :)

doskabouter:
@dbaston : is there anything that I still need to do for this to get merged?

pramsey:
I'm not sure what benefit this RP is providing? I've read through it a couple times, it seems to have just moved some logic out of RelateComputer into GeometryCollection, which is a smelly place to put a static f(geometry,geometry), right? And even so, I'm not seeing the win? Is it having the envelope test in prepared also?

doskabouter:
Thanks for looking into this!
Thre main benefits are in the relatecomputer.cpp and basicpreparedgeometry.cpp, the rest is some refactoring to (try to) move the envelopeIntersects to a logical place.

The benefit is that instead of checking the intersection-ness of the envelope of multigeometries it checks if individual geometry-envelopes are intersecting. This saves a lot of time in f.e. checking intersects of country "The Netherlands" with country "Russia" (small part at -180deg and big part up to 180deg, in this case envelope of Russia is as wide as the world)

dbaston:
Yes, there is a practical problem that is being addressed by this PR. I just haven't had time to take a look at it.

dr-jts:
This does introduce a potential O(NxM) step in the algorithm.  One case where this would be a hotspot is when comparing two MultiPoints.  And in the case of prepared geometries, possibly when testing a MultiPoint against a prepared MultiPolygon.

These may be less common cases than the ones this will optimize, but it would be nice to not impact performance for them, since they might be common in some workloads.  In particular, it seems important to optimize handling MultiPoints, since these might be used as a pointcloud type.  (Note that the current implementation handles points efficiently currently using a Map)

Unfortunately this is going to introduce a lot of special-case pathways.  But that's life in the optimization fast lane.

So at a minimum the extra logic should be:

Relate- skip check if both inputs are Puntal
Prepared Geom - skip check if target is Puntal

Also, In the extreme case of comparing two very large Multigeoms it would be faster to build an index on one and use that.  Perhaps this could be done only in PreparedGeometry.  This will require introducing some tuning parameters to define the threshold for indexing.  JTS has mostly avoided doing this, but more advanced optimization will probably require this.

dr-jts:
An alternative might be to compute the envelope of intersection and then discard any components from each geometry which were disjoint with the common envelope.

This would subsume the disjoint component envelope check (in many cases although not all).  It would also improve performance in some cases in which there was interaction (by avoiding building topology for components which don't interact).

dr-jts:
re computing intersection of envelopes for initial test - I mean this would be computed using the single top-level envelope from each input geometry.  To go deeper would risk the O(NxM) behaviour again (and would potentially have a quite complex output).

Not sure I understand your point about this being equivalent to "testing against the largest envelope in the outer loop".  In the case of Russia VS Holland isn't the key point to compare to the smaller envelope?

Also, computing the "envelope of interaction" allows pruning components which do not intersect (which has O(N + M) performance, so is acceptable).  But this would occur later in the algorithm (i.e. when each component is added to the GeometryGraph).

re reuse of envelope index in PreparedGeometry - maybe.  One place would be optimizing point-in-polygon tests.  Although those are already indexed using PointInAreaLocator, so the payoff might not be worth the extra complexity.  Segment intersection is already indexed, so it wouldn't help there.

doskabouter:
I've added the changes you suggested, hopefully it's correct now.

re reuse of envelope index in PreparedGeometry: I did remember seeing some loop over all parts which could benefit from an index, but can't remember where exactly...

re computing intersection of envelopes for initial test: would it be adding an env parameter to https://github.com/libgeos/geos/blob/b88d4614aaf3c5e73fd275a88253f62f34c25bb9/src/operation/relate/RelateComputer.cpp#L92 and https://github.com/libgeos/geos/blob/b88d4614aaf3c5e73fd275a88253f62f34c25bb9/src/operation/relate/RelateComputer.cpp#L117? Or is more needed to make it work?

doskabouter:
Hi,
Any news on this one?

dr-jts:
I haven't had a chance to think about this more, but will try and prioritize it.

doskabouter:
thanks!

mloskot:
Regardless what's the correctness or usefulness of this contribution, it is (seriously) lacking of tests.

doskabouter:
Merged master into this so no mergeconflicts anymore.
About tests: I have some that I can add (a new xml in https://github.com/libgeos/geos/tree/master/tests/xmltester/tests) but I'm not sure if I'm able to test prepared geometries. 

Found some tests about prepared geometries but wondering if  [this](https://github.com/libgeos/geos/blob/3133734b73aec52c70643c41fc43f19cfb5f680f/tests/xmltester/tests/general/TestPreparedPolygonPredicate.xml#L4) is the trigger to work with prepared geometries?

dbaston:
I think it's probably best to test the `envelopeIntersects` method directly; the relate tests performed through XML files seem too high-level.

See also the discussion at https://github.com/libgeos/geos/pull/192#discussion_r291005721 for another case that points to the utility of a `bool Geometry::envelopeIntersects(const Geometry & other)` method.

doskabouter:
Is this ok for unit-tests?

doskabouter:
any news on this?

doskabouter:
Is this going to get any traction?

pramsey:
No, I think we're sticking with aping JTS for now. If you move the patch to JTS it'll downstream here eventually.

dr-jts:
Also, we're hoping to completely rework the predicate evaluation code, and that should subsume or obsolete this approach.

PR #147, opened 2019-02-22
src/algorithm/InteriorPointArea.cpp: fix -Wshadow warning (fixes #954)

### LINKED ISSUE #954: I have a multi-line, a region polygon, these two elements are intersecting, I want to calculate the region polygon is separated by multiple lines after multiple regions, is there a relevant API
just like the title

rouault:


rouault:
I've rebased on top of master, so now all CI tests pass

PR #149, opened 2019-02-22
.travis.yml: use ccache to speed-up builds

rouault:


rouault:
Fixes https://trac.osgeo.org/geos/ticket/956

PR #161, opened 2019-05-16
Modernize CMake configuration

pramsey:
Refactoring based on practices of the Modern CMake approach:

* Set project version based on `version.txt` spec.
* Build shared libraries by default.
* Build tools and `geos-config` script.
* Add make `check` target to run `ctest` (no dependencies).
* Move failing/overlarge tests into `failure` directory.
* Add target to deploy Config-module with installed targets:
  `GEOS::geos`, `GEOS::geos_c`.
* Add basic uninstall target
* Add basic client example configured with CMake to use GEOS targets.


pramsey:
Merged at gitea

PR #163, opened 2019-05-19
Reduce CMake version requirement

dbaston:
This PR reduces the CMake version requirement with Visual Studio builds to 3.13, which is the version currently installed on AppVeyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@
 # by the Free Software Foundation.
 # See the COPYING file for more information.
 #
-dist: trusty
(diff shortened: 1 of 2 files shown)


dbaston:
@pramsey any objection to this? FYI, with this patch applied, the Windows builds start failing on a ttmath issue.

pramsey:
I guess not. I thought I had all the ttmath stuff working on appveyor back when I added it in :/
Was hoping to de-appveyor us in favour of the Azure Pipelines, but that's in the future.

PR #165, opened 2019-05-21
Disable ttmath ASM for MSVC

dbaston:


dbaston:
@pramsey I think this fixes the ttmath issues but both builds are still failing on a missing `geos_revision.h`. I may be able to debug but will stand aside if you're working on it.

pramsey:
Ah. Yeah put that in, but we need to go back and find the magic flag to enable the MSCV ASM again if we can... shame to lose that.

pramsey:
I'm not working on the revision.h problem, am looking at 'make distcheck' now

PR #167, opened 2019-05-22
Pull removeRepeatedPoints out of CoordinateSequence

dbaston:
This refactoring supports the goal of a slimmed-down CoordinateSequence
interface against which it will be easier to develop alternative
implementations.

### DIFF
--- a/include/geos/geom/CoordinateArraySequence.h
+++ b/include/geos/geom/CoordinateArraySequence.h
@@ -128,8 +128,6 @@ public:
 
     void apply_ro(CoordinateFilter* filter) const override;
 
-    CoordinateSequence& removeRepeatedPoints() override;
-
 private:
     std::vector<Coordinate>* vect;
     mutable std::size_t dimension;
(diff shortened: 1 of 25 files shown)


dbaston:
Squashed and committed as 75c286d5b65ca

dr-jts:
Late to the party here.... but I would suggest moving the repeated points code out of `operation/valid` and into `geom/util`.  

And, JTS has followed the Java convention of putting utility methods into classes called `CoordinateArrays` and `CoordinateSequences` (which are actually in `geom`, but should probably be in `geom/util`).  So might be an idea to follow this pattern (which also reduces the class count, by avoiding creating single-method classes).

dbaston:
@dr-jts Yes, I see that point of view. I put it under `valid` (seems an odd place) because that's where I found the `RepeatedPointTester`, and I envisioned that a `RepeatedPointRemover` could end up having methods for geometries as well as `CoordiateSequence`. I have no objection to moving it though.

PR #168, opened 2019-05-22
Generate git revision hash into geos_c.h

pramsey:
instead of a separate .h file


mloskot:
LGTM. Thx!

capi/CMakeLists.txt
```diff
@@ -9,18 +9,20 @@
 # See the COPYING file for more information.
 ################################################################################
 
-if(GEOS_BUILD_FROM_GIT)
-  add_custom_target(geos_revision_header
```
mloskot: Replacing the target w/ plain command changes the procedure a bit
AFAIR, previous version re-generated the info on every build, not every re-configuration.

dbaston: I think it should be regenerated on each build, or just left out. A potentially incorrect hash seems to have more confusion potential than no hash.

pramsey: Well, generating every build has it's own downside, which Sandro alluded to on channel, which is the regen can end up kicking off large scale rebuilds over the code when not really necessary. I'm inclined to remove the whole thing.

dbaston: > I'm inclined to remove the whole thing.

Agreed

mloskot: >  regen can end up kicking off large scale rebuilds over the code 

It shouldn't, if `geos_revision.h` promises to stick to `geos_ts.cpp` only.

> I'm inclined to remove the whole thing.

Sweet

pramsey: Harsher approach at #174

PR #171, opened 2019-05-23
Use unique_ptr for prepared geometries

dbaston:


capi/geos_ts_c.cpp
```diff
@@ -5280,7 +5280,7 @@ extern "C" {
         const geos::geom::prep::PreparedGeometry* prep = 0;
 
         try {
-            prep = geos::geom::prep::PreparedGeometryFactory::prepare(g);
```
dr-jts: Was this a memory leak before?  Since it apparently wasn't freed?

dbaston: Well, it ultimately gets returned to the C API caller (line 5292) who would have responsibility for freeing it then as now.

dr-jts: Ok, so another case of unclear ownership, which unique_ptr helps to fix.  Good stuff!

PR #173, opened 2019-05-23
Remove CoordinateSequence::toVector()

dbaston:
Deprecated a decade ago, but still used a couple of times within GEOS.
Supports the goal of a smaller CoordinateSequence interface.

### DIFF
--- a/src/operation/overlay/snap/GeometrySnapper.cpp
+++ b/src/operation/overlay/snap/GeometrySnapper.cpp
@@ -59,8 +59,9 @@ private:
         using std::unique_ptr;
 
         assert(srcPts);
-        assert(srcPts->toVector());
-        LineStringSnapper snapper(*(srcPts->toVector()), snapTol);
+        std::vector<Coordinate> coords;
+        srcPts->toVector(coords);
+        LineStringSnapper snapper(coords, snapTol);
         unique_ptr<Coordinate::Vect> newPts = snapper.snapTo(snapPts);
 
         const CoordinateSequenceFactory* cfact = factory->getCoordinateSequenceFactory();
--- a/src/simplify/DouglasPeuckerSimplifier.cpp
+++ b/src/simplify/DouglasPeuckerSimplifier.cpp
@@ -105,12 +105,11 @@ DPTransformer::transformCoordinates(
 {
     ::geos::ignore_unused_variable_warning(parent);
 
-    const Coordinate::Vect* inputPts = coords->toVector();
-    assert(inputPts);
+    Coordinate::Vect inputPts;
+    coords->toVector(inputPts);
 
     std::unique_ptr<Coordinate::Vect> newPts =
-        DouglasPeuckerLineSimplifier::simplify(*inputPts,
-                distanceTolerance);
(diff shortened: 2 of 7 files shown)


src/operation/overlay/snap/GeometrySnapper.cpp
```diff
@@ -59,8 +59,9 @@ class SnapTransformer: public geos::geom::util::GeometryTransformer {
         using std::unique_ptr;
 
         assert(srcPts);
-        assert(srcPts->toVector());
-        LineStringSnapper snapper(*(srcPts->toVector()), snapTol);
+        std::vector<Coordinate> coords;
+        srcPts->toVector(coords);
```
mloskot: Ideally, if this could read

```
std::vector<Coordinate> coords;
srcPts->toVector(std::move(coords));
```

src/simplify/DouglasPeuckerSimplifier.cpp
```diff
@@ -105,12 +105,11 @@ DPTransformer::transformCoordinates(
 {
     ::geos::ignore_unused_variable_warning(parent);
 
-    const Coordinate::Vect* inputPts = coords->toVector();
-    assert(inputPts);
+    Coordinate::Vect inputPts;
+    coords->toVector(inputPts);
```
mloskot: also here, if possible

```cpp
 coords->toVector(std::move(inputPts));
```

src/operation/overlay/snap/GeometrySnapper.cpp
```diff
@@ -59,8 +59,9 @@ class SnapTransformer: public geos::geom::util::GeometryTransformer {
...
+        LineStringSnapper snapper(coords, snapTol);
```
mloskot: I think, grand objective would be to get this bit read as:

```cpp
std::vector<Coordinate> coords = srcPts->toVector();
LineStringSnapper snapper(std::move(coords), snapTol);
```

src/simplify/DouglasPeuckerSimplifier.cpp
dbaston: Sorry, I don't understand :(. Are you suggesting to change the signature of `CoordinateSequence::toVector(std::vector<Coordinate> & v)` ?

mloskot: Oh, sorry, my bad.
I meant to write

```cpp
auto inputPts = coords->toVector();
```

src/operation/overlay/snap/GeometrySnapper.cpp
dbaston: `LineStringSnapper` is taking `vector<T> const&`. But I see the merits of going to 

```
auto coords = srcPts->toVector();
LineStringSnapper snapper(coords, snapTol);
```

dr-jts: Ironically the LineStringSnapper is not mutating the input coords anyway, and so this is just making an extra copy.  This is an example of where JTS is not fully utilzing the CoordinateSequence interface (i.e. creating an extra copy if the CS is not actually an array).

PR #176, opened 2019-05-23
Fix MSVC warning C4189: local variable not referenced

mloskot:
This is to avoid flood of 100+ occurrences of the warning.

### DIFF
--- a/tests/unit/tut/tut.hpp
+++ b/tests/unit/tut/tut.hpp
@@ -235,7 +235,8 @@ class test_group : public group_base, public test_group_posix
                         " an SEH exception");
                 }
 #else
-                bool d = delete_obj();
+                bool d;
+                d = delete_obj();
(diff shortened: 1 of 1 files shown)


tests/unit/tut/tut.hpp
```diff
@@ -235,7 +235,8 @@ class test_group : public group_base, public test_group_posix
                         " an SEH exception");
                 }
 #else
-                bool d = delete_obj();
+                bool d;
+                d = delete_obj();
```
dbaston: Nitpick, but can you add an inline comment that explains this? Otherwise I'll be tempted to change it back when clang-tidy or some other thing complains about this version.

dbaston: Disregard, I did it myself.

mloskot:
Thanks

PR #177, opened 2019-05-23
Add GEOS_DLL missing from class RepeatedPointRemover

mloskot:
Fix @dbaston 's questionable behaviour 🤣


dbaston:
Thought of this at 5 AM but the thought did not translate to action....thanks for fixing

PR #178, opened 2019-05-24
Add make_unique to geos_util.h

dbaston:
This PR "backports" the `std::make_unique` function from C++14 to the `geos_util.h` header and applies it to a portion of the locations where it makes sense. `make_unique`, which I copied from QGIS (in turn copied from Stack Overflow) lets us replace this syntax:

`auto a = std::unique_ptr<std::vector<const Geometry*>>(new std::vector<const Geometry*>(5));`

with this syntax:

`auto a = make_unique<std::vector<const Geometry*>>(5);`

So there is a bit of a win in avoiding writing long type names two times. It also avoids the potential for a memory leak if one of the arguments to the `std::unique_ptr` construction throws; see https://stackoverflow.com/a/19472607 or Effective Modern C++

If people are OK with adding `make_unique` to our own header, I can make the rest of the conversions. I think there are only a couple dozen at this point. But if we're going to add `make_unique`, it makes sense to do it before embarking on a whole bunch of `unique_ptr` refactoring.

If the project ultimately moves to C++14 or later, we can just remove our version.

Past discussion: https://lists.osgeo.org/pipermail/geos-devel/2018-December/008768.html

### DIFF
--- a/include/geos/util.h
+++ b/include/geos/util.h
@@ -31,6 +31,8 @@
 #include <geos/util/GeometricShapeFactory.h>
 //#include <geos/util/math.h>
 
+#include <memory>
+
 //
 // Private macros definition
 //
@@ -38,8 +40,48 @@
 namespace geos {
 template<class T>
 void
-ignore_unused_variable_warning(T const&) {}
+ignore_unused_variable_warning(T const &) {}
+
+namespace detail {
+#if __cplusplus >= 201402L
+using std::make_unique;
+#else
+// Backport of std::make_unique to C++11
+// Source: https://stackoverflow.com/a/19472607
+template<class T>
+struct _Unique_if {
+    typedef std::unique_ptr<T> _Single_object;
+};
+
+template<class T>
+struct _Unique_if<T[]> {
+    typedef std::unique_ptr<T[]> _Unknown_bound;
+};
+
+template<class T, size_t N>
+struct _Unique_if<T[N]> {
+    typedef void _Known_bound;
+};
+
+template<class T, class... Args>
+typename _Unique_if<T>::_Single_object
+make_unique(Args &&... args) {
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
 
+template<class T>
+typename _Unique_if<T>::_Unknown_bound
+make_unique(size_t n) {
+    typedef typename std::remove_extent<T>::type U;
+    return std::unique_ptr<T>(new U[n]());
+}
+
+template<class T, class... Args>
+typename _Unique_if<T>::_Known_bound
+make_unique(Args &&...) = delete;
+
+#endif
+}
+}
 
 #endif // GEOS_UTIL_H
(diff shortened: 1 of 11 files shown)


nyalldawson:
For reference, over at the QGIS project we've used this approach for the past 12 months or more without issue.

pramsey:
Can we wrap our implementation in a C++14 check so that it automagically gets out of the way when in the presence of C++14?

dbaston:
@pramsey I'm not aware of a way to do so, unless we put our implementation into the `std` namespace instead of the `geos` namespace.

mloskot:
1. Do not define anything in `std` namespace, unless it falls into any of allowed exceptions here https://en.cppreference.com/w/cpp/language/extending_std

2. Do not define any internal utils in `namespace geos` (and `make_unique` is not part of public inteface)

3. I'd try something like this

```
#include <memory>

namespace geos { namespace detail {

#if __cplusplus >= 201402L

using std::make_unique;

#else

// Backport of std::make_unique to C++11
// Source: 
template<class T> struct _Unique_if
{
    typedef std::unique_ptr<T> _Single_object;
};

template<class T> struct _Unique_if<T[]>
{
    typedef std::unique_ptr<T[]> _Unknown_bound;
};

template<class T, size_t N> struct _Unique_if<T[N]>
{
    typedef void _Known_bound;
};

template<class T, class... Args>
typename _Unique_if<T>::_Single_object
make_unique( Args &&... args )
{
    return std::unique_ptr<T>( new T( std::forward<Args>( args )... ) );
}

template<class T>
typename _Unique_if<T>::_Unknown_bound
make_unique( size_t n )
{
    typedef typename std::remove_extent<T>::type U;
    return std::unique_ptr<T>( new U[n]() );
}

template<class T, class... Args>
typename _Unique_if<T>::_Known_bound
make_unique( Args &&... ) = delete;

#endif
}}

using namespace geos;

int main()
{
    auto p1 = detail::make_unique<int>();
}
```

dbaston:
Still trying to figure out the cleanest way to implement this, between the options of

(a) specifying the namespace at each call site, e.g.:

`auto cs = detail::make_unique<CoordinateArraySequence>()`

(b) importing `make_unique` at the top of each compilation unit, with

`using geos::detail::make_unique`

The latter seems cleaner and is easier to rip out later whenever the project moves to a C++14 minimum. However, it produces an error when compiling with MSVC in the many files that include a blanket `using namespace std` at the top.

I think the best solution is probably to stick with `using geos::detail::make_unique` and removing the many blanket `using namespace std` statements.

mloskot:
IMHO, always use least ambiguous variant i.e. `detail::make_unique<T>(...) `.
Trying to save on typing is not a good reason.

schwehr:
Yes, please remove `using namespace std`.  That is endless trouble.  BTW, doesn't hurt to check the implementation of things in absl.  It has many backports that have been used in production code for a long time. 

*Please* do not use `using geos::detail::make_unique;`

dbaston:
Thanks all for the feedback.

PR #179, opened 2019-05-24
Point CMAKE_*_OUTPUT_DIRECTORY to common location

mloskot:
Place executables and shared libraries in the same location for
convenience of direct execution from common spot and for
convenience in environments without `RPATH` support.

In future, if number of generated executables grows significantly,
with possible name clashes, it may be more convenient to replace
the common locations with setting `PATH` per test, on Windows only:

```
set_tests_properties(<test name> PROPERTIES
  ENVIRONMENT
    "PATH=$<SHELL_PATH:$<TARGET_FILE_DIR:geos>>$<SEMICOLON>$<TARGET_FILE_DIR:geos_c>>$<SEMICOLON>$ENV{PATH}"
)
```

----

@robe2 this should fix the test issues on Windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,18 @@ cmake_dependent_option(GEOS_BUILD_DEVELOPER
 mark_as_advanced(GEOS_BUILD_DEVELOPER)
 
 #-----------------------------------------------------------------------------
+# Setup build directories
+#-----------------------------------------------------------------------------
+# Place executables and shared libraries in the same location for
+# convenience of direct execution from common spot and for
+# convenience in environments without RPATH support.
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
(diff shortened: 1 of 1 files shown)


dr-jts:
So you decided to move all the execs into `bin` after all?  Was this to fix some problem with the deploy chain?

mloskot:
@dr-jts Yes, I decided to do it to make RPATH-less Windows builds happy

Alternative, or canonical way, is to use the `set_tests_properties` trick as displayed in the PR description, but since GEOS does not produce huge amount of EXE-s, the global workaround is fine.

-----

I highly (or even higher) recommend to buy https://crascit.com/professional-cmake/ by @audiofanatic - there is almost no detail he does not explain :-)

image

dr-jts:
Good catch, and good to know.  And I won't say I told you so :) . (Although I didn't realize there was such a hard technical reason for their previous location).

mloskot:
wise man are wise due to ability to doubt in their wiseness 😋

PR #181, opened 2019-05-24
Disable dist and distcheck targets for multi-configuration generators

mloskot:
Those targets are useless for the IDEs.
Those targets have also potential to be troublesome for the IDEs.

-----

```
5>  Microsoft (R) Build Engine version 16.1.76+g14b0a930a7 for .NET Framework
5>  Copyright (C) Microsoft Corporation. All rights reserved.
5>
5>  MSBUILD : error MSB1009: Project file does not exist.
5>  Switch: package_source
5>Done building target "CustomBuild" in project "dist.vcxproj" -- FAILED.
```

Yup, I remain stubborn all that Autotoolization is simply "habits worst enemy of human" cruft.
The `include(CPack)` already defines dedicated target, so `cmake --build <build dir> --target package_source` or even `make package_source`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,41 +279,47 @@ add_custom_target(uninstall
 #-----------------------------------------------------------------------------
 # "make dist" workalike
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pramsey:
Merged

PR #182, opened 2019-05-25
Clean up and update AppVeyor configuration

mloskot:
Install and use latest CMake instead of the one shipped on AppVeyor.
NMAKE is dead. Remove jobs using CMake generator for NMAKE.
Re-enable builds targeting 32-bit platform and VS 2015.

[skip travis]

-----

/cc @robe2

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,9 +2,13 @@ version: 1.0.{build}
 
 image: Visual Studio 2017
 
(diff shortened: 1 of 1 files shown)


mloskot:
Apparently, 10 minutes per build job is "super slow" hence unacceptable: https://trac.osgeo.org/geos/ticket/955

PR #192, opened 2019-06-06
GEOSPreparedContains performance improvements

dbaston:
This PR improves the performance of GEOSPreparedContains by about 30% for point inputs and adds some benchmarking code to the repository.

@dr-jts some of the tweaks here may be relevant for JTS. They're probably most easily reviewed in the individual commits, rather than by the whole PR diff.

### DIFF
--- a/benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
+++ b/benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
@@ -1,0 +1,114 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2019 Daniel Baston <dbaston@gmail.com>
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************/
+
+#include <geos/profiler.h>
+#include <geos_c.h>
+
+#include <geos/geom/Coordinate.h>
+#include <geos/geom/GeometryFactory.h>
+#include <geos/geom/Polygon.h>
+#include <geos/geom/util/SineStarFactory.h>
+
+#include <algorithm>
+#include <random>
+#include <vector>
+#include <memory>
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
+class GEOSPreparedContainsPerfTest {
+
+public:
+    void test(const GEOSGeometry* g, size_t num_points) {
+        using namespace geos::geom;
+
+        double xmin, xmax, ymin, ymax;
+
+        GEOSGeom_getXMin(g, &xmin);
+        GEOSGeom_getXMax(g, &xmax);
+        GEOSGeom_getYMin(g, &ymin);
+        GEOSGeom_getYMax(g, &ymax);
+
+        std::default_random_engine e(12345);
+        std::uniform_real_distribution<> xdist(xmin, xmax);
+        std::uniform_real_distribution<> ydist(ymin, ymax);
+
+        std::vector<Coordinate> coords(num_points);
+        std::generate(coords.begin(), coords.end(), [&xdist, &ydist, &e]() {
+            return Coordinate(xdist(e), ydist(e));
+        });
+
+        geos::util::Profile sw("GEOSPreparedContains");
+        sw.start();
+
+        size_t hits = 0;
+        auto prep = GEOSPrepare(g);
+        for (const auto& c : coords) {
+            auto seq = GEOSCoordSeq_create(1, 2);
+            GEOSCoordSeq_setX(seq, 0, c.x);
+            GEOSCoordSeq_setY(seq, 0, c.y);
+            auto pt = GEOSGeom_createPoint(seq);
+
+            if (GEOSPreparedContains(prep, pt)) {
+                hits++;
+            }
+
+            GEOSGeom_destroy(pt);
+        }
+
+        GEOSPreparedGeom_destroy(prep);
+
+        sw.stop();
+
+        std::cout << sw.name << ": " << hits << " hits from " << num_points << " points in " <<  sw.getTotFormatted() << std::endl;
+
+    }
+};
+
+int main(int argc, char** argv) {
+    if (argc != 3) {
+        std::cout << "perf_geospreparedcontins performs a specified number of point-in-polygon tests" << std::endl;
+        std::cout << "on randomly generated points from the bounding box of a single geometry provided" << std::endl;
+        std::cout << "in a file as WKT." << std::endl;
+        std::cout << std::endl;
+        std::cout << "Usage: perf_geospreparedcontins [wktfile] [n]" << std::endl;
+        return 0;
+    }
+
+    GEOSPreparedContainsPerfTest tester;
+
+    int n = std::atoi(argv[2]);
+    std::cout << "Performing " << n << " point-in-polygon tests." << std::endl;
+
+    std::string fname{argv[1]};
+    std::cout << "Reading shape from " << fname << std::endl;
+
+    std::ifstream f(fname);
+    std::stringstream buff;
+    buff << f.rdbuf();
+    f.close();
+
+    std::string wkt = buff.str();
+    buff.clear();
+
+    initGEOS(nullptr, nullptr);
+    GEOSGeometry* g = GEOSGeomFromWKT(wkt.c_str());
+    wkt.clear();
+
+    tester.test(g, n);
+
+    GEOSGeom_destroy(g);
+}
--- a/src/algorithm/RayCrossingCounter.cpp
+++ b/src/algorithm/RayCrossingCounter.cpp
@@ -75,21 +75,6 @@ RayCrossingCounter::locatePointInRing(const geom::Coordinate& point,
     return rcc.getLocation();
 }
 
-/*public static*/
-int
-RayCrossingCounter::orientationIndex(const geom::Coordinate& p1,
-                                     const geom::Coordinate& p2, const geom::Coordinate& q)
-{
-    // travelling along p1->p2, turn counter clockwise to get to q return 1,
-    // travelling along p1->p2, turn clockwise to get to q return -1,
-    // p1, p2 and q are colinear return 0.
-    double dx1 = p2.x - p1.x;
-    double dy1 = p2.y - p1.y;
-    double dx2 = q.x - p2.x;
-    double dy2 = q.y - p2.y;
-    return CGAlgorithmsDD::signOfDet2x2(dx1, dy1, dx2, dy2);
-}
-
 void
 RayCrossingCounter::countSegment(const geom::Coordinate& p1,
                                  const geom::Coordinate& p2)
@@ -138,7 +123,7 @@ RayCrossingCounter::countSegment(const geom::Coordinate& p1,
             ((p2.y > point.y) && (p1.y <= point.y))) {
         // For an upward edge, orientationIndex will be positive when p1->p2
         // crosses ray. Conversely, downward edges should have negative sign.
-        int sign = orientationIndex(p1, p2, point);
+        int sign = CGAlgorithmsDD::orientationIndex(p1, p2, point);
         if(sign == 0) {
             isPointOnSegment = true;
             return;
--- a/src/geom/prep/AbstractPreparedPolygonContains.cpp
+++ b/src/geom/prep/AbstractPreparedPolygonContains.cpp
@@ -109,25 +109,38 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
 {
     // Do point-in-poly tests first, since they are cheaper and may result
     // in a quick negative result.
-    //
-    // If a point of any test components does not lie in target,
-    // result is false
-    if (!requireSomePointInInterior || geom->getGeometryTypeId() != GEOS_POINT) {
-        bool isAllInTargetArea = isAllTestComponentsInTarget(geom);
-        if (!isAllInTargetArea) {
-            return false;
-        }
+    auto outermostLoc = getOutermostTestComponentLocation(geom);
+
+    if (outermostLoc == Location::EXTERIOR) {
+        // If a point of any test components does not lie in target,
+        // result is false
+        return false;
     }
 
-    // If the test geometry consists of only Points,
-    // then it is now sufficient to test if any of those
-    // points lie in the interior of the target geometry.
+    // Short-circuit for (Multi)Points
+    //
+    // For the Covers predicate, we can return true
+    // since no Points are on the exterior of the Polygon.
+    //
+    // For the Contains predicate, we need to test if any
+    // of those points lie in the interior of the target
+    // geometry.
+    //
     // If so, the test is contained.
     // If not, all points are on the boundary of the area,
     // which implies not contained.
-    if(requireSomePointInInterior && geom->getDimension() == 0) {
-        bool isAnyInTargetInterior = isAnyTestComponentInTargetInterior(geom);
-        return isAnyInTargetInterior;
+    if (geom->getDimension() == 0) {
+        if (requireSomePointInInterior && outermostLoc == Location::BOUNDARY) {
+            if (geom->getNumGeometries() > 1) {
+                return isAnyTestComponentInTargetInterior(geom);
+            } else {
+                // We only had one point, so if it was on the
+                // boundary, it is not contained.
+                return false;
+            }
+        } else {
+            return true;
+        }
     }
 
     // Check if there is any intersection between the line segments
--- a/src/geom/prep/AbstractPreparedPolygonContains.cpp
+++ b/src/geom/prep/AbstractPreparedPolygonContains.cpp
@@ -112,9 +112,11 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
     //
     // If a point of any test components does not lie in target,
     // result is false
-    bool isAllInTargetArea = isAllTestComponentsInTarget(geom);
-    if(!isAllInTargetArea) {
-        return false;
+    if (!requireSomePointInInterior || geom->getGeometryTypeId() != GEOS_POINT) {
+        bool isAllInTargetArea = isAllTestComponentsInTarget(geom);
+        if (!isAllInTargetArea) {
+            return false;
+        }
     }
 
     // If the test geometry consists of only Points,
--- a/src/geom/prep/AbstractPreparedPolygonContains.cpp
+++ b/src/geom/prep/AbstractPreparedPolygonContains.cpp
@@ -111,36 +111,15 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
     // in a quick negative result.
     auto outermostLoc = getOutermostTestComponentLocation(geom);
 
+    // Short-circuit for (Multi)Points
+    if (geom->getDimension() == 0) {
+        return evalPointTestGeom(geom, outermostLoc);
+    }
+
     if (outermostLoc == Location::EXTERIOR) {
         // If a point of any test components does not lie in target,
         // result is false
         return false;
-    }
-
-    // Short-circuit for (Multi)Points
-    //
-    // For the Covers predicate, we can return true
-    // since no Points are on the exterior of the Polygon.
-    //
-    // For the Contains predicate, we need to test if any
-    // of those points lie in the interior of the target
-    // geometry.
-    //
-    // If so, the test is contained.
-    // If not, all points are on the boundary of the area,
-    // which implies not contained.
-    if (geom->getDimension() == 0) {
-        if (requireSomePointInInterior && outermostLoc == Location::BOUNDARY) {
-            if (geom->getNumGeometries() > 1) {
-                return isAnyTestComponentInTargetInterior(geom);
-            } else {
-                // We only had one point, so if it was on the
-                // boundary, it is not contained.
-                return false;
-            }
-        } else {
-            return true;
-        }
     }
 
     // Check if there is any intersection between the line segments
@@ -200,6 +179,37 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
     }
 
     return true;
+}
+
+bool AbstractPreparedPolygonContains::evalPointTestGeom(const Geometry *geom,
+                                                      Location::Value outermostLoc) {
+    // If we had a point on the ourside of the polygon,
+    // we aren't covered or contained.
+    if (outermostLoc == Location::EXTERIOR) {
+        return false;
+    }
+
+    // For the Covers predicate, we can return true
+    // since no Points are on the exterior of the target
+    // geometry.
+    if (!requireSomePointInInterior) {
+        return true;
+    }
+
+    // For the Contains predicate, we need to test if any
+    // of those points lie in the interior of the target
+    // geometry.
+    if (outermostLoc == Location::INTERIOR) {
+        return true;
+    }
+
+    if (geom->getNumGeometries() > 1) {
+        // for MultiPoint, try to find at least one point
+        // in interior
+        return isAnyTestComponentInTargetInterior(geom);
+    }
+
+    return false;
 }
 
 //
--- a/src/geom/prep/BasicPreparedGeometry.cpp
+++ b/src/geom/prep/BasicPreparedGeometry.cpp
@@ -40,12 +40,20 @@ BasicPreparedGeometry::setGeometry(const geom::Geometry* geom)
 bool
 BasicPreparedGeometry::envelopesIntersect(const geom::Geometry* g) const
 {
+    if (g->getGeometryTypeId() == GEOS_POINT) {
+        return baseGeom->getEnvelopeInternal()->intersects(*(g->getCoordinate()));
+    }
+
     return baseGeom->getEnvelopeInternal()->intersects(g->getEnvelopeInternal());
 }
 
 bool
 BasicPreparedGeometry::envelopeCovers(const geom::Geometry* g) const
 {
+    if (g->getGeometryTypeId() == GEOS_POINT) {
+        return baseGeom->getEnvelopeInternal()->covers(g->getCoordinate());
+    }
+
     return baseGeom->getEnvelopeInternal()->covers(g->getEnvelopeInternal());
 }
 
--- a/src/geom/prep/PreparedPolygonPredicate.cpp
+++ b/src/geom/prep/PreparedPolygonPredicate.cpp
@@ -26,7 +26,6 @@
 #include <geos/algorithm/locate/PointOnGeometryLocator.h>
 #include <geos/algorithm/locate/SimplePointInAreaLocator.h>
 // std
-#include <algorithm> // std::max
 #include <cstddef>
 
 namespace geos {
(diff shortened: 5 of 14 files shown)


src/geom/prep/BasicPreparedGeometry.cpp
```diff
@@ -40,12 +40,20 @@ BasicPreparedGeometry::setGeometry(const geom::Geometry* geom)
 bool
 BasicPreparedGeometry::envelopesIntersect(const geom::Geometry* g) const
 {
+    if (g->getDimension() == 0) {
+        return baseGeom->getEnvelopeInternal()->intersects(*(g->getCoordinate()));
```
dr-jts: Does this handle a MultiPoint?

src/geom/prep/BasicPreparedGeometry.cpp
```diff
@@ -40,12 +40,20 @@ BasicPreparedGeometry::setGeometry(const geom::Geometry* geom)
...
+    }
+
     return baseGeom->getEnvelopeInternal()->intersects(g->getEnvelopeInternal());
 }
 
 bool
 BasicPreparedGeometry::envelopeCovers(const geom::Geometry* g) const
 {
+    if (g->getDimension() == 0) {
+        return baseGeom->getEnvelopeInternal()->covers(g->getCoordinate());
```
dr-jts: Same possible issue here - needs to handle a MultiPoint.

src/geom/prep/BasicPreparedGeometry.cpp
dbaston: I'll have to test what the fastest way is. Maybe just trying a `dynamic_cast` to `Point`.

src/geom/prep/AbstractPreparedPolygonContains.cpp
```diff
@@ -112,9 +112,11 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
     //
     // If a point of any test components does not lie in target,
     // result is false
-    bool isAllInTargetArea = isAllTestComponentsInTarget(geom);
-    if(!isAllInTargetArea) {
-        return false;
+    if (geom->getDimension() > 0) {
```
dr-jts: possible MultiPoint issue here too?

src/algorithm/RayCrossingCounter.cpp
```diff
@@ -138,7 +123,7 @@ RayCrossingCounter::countSegment(const geom::Coordinate& p1,
             ((p2.y > point.y) && (p1.y <= point.y))) {
         // For an upward edge, orientationIndex will be positive when p1->p2
         // crosses ray. Conversely, downward edges should have negative sign.
-        int sign = orientationIndex(p1, p2, point);
+        int sign = CGAlgorithmsDD::orientationIndex(p1, p2, point);
```
dr-jts: This is already in JTS.

src/geom/prep/AbstractPreparedPolygonContains.cpp
dbaston: Definitely.

src/geom/prep/BasicPreparedGeometry.cpp
dr-jts: The other possibility is to provide methods on Geometry which check Envelope intersection and containment.  That way Point geometries can optimize by avoiding creating Envelopes just for that check.  Essentially this converts the logic for checking for a Point geometry type into a virtual call.

This seems a bit cumbersome, but I can't think of any other way to avoid materializing an Envelope object when it isn't strictly needed.

dbaston: > The other possibility is to provide methods on Geometry which check Envelope intersection and containment.

That's a good idea.

dr-jts: I'm surprised this didn't trigger a test error.  Perhaps more test coverage is needed?

dr-jts: If envelope testing is moved into the Geometry class as a virtual, then that may open the way to eliminating envelopes on Points completely.  It seems like this would be a large potential memory savings?

Not sure this can happen in JTS as it stands, but GEOS has more flexibility.

dbaston: Another option is to make `Envelope` into more of an interface, and have a `PointEnvelope` that is backed by a single coordinate.

src/geom/prep/PreparedPolygonPredicate.cpp
```diff
@@ -91,6 +91,11 @@ bool
 PreparedPolygonPredicate::isAnyTestComponentInTargetInterior(
     const geom::Geometry* testGeom) const
 {
+    if (testGeom->getGeometryTypeId() == GEOS_POINT) {
+        // Avoid creating a vector to store a single point
+        return prepPoly->getPointLocator()->locate(testGeom->getCoordinate()) == geom::Location::INTERIOR;
+    }
+
     geom::Coordinate::ConstVect pts;
     geom::util::ComponentCoordinateExtracter::getCoordinates(*testGeom, pts);
```
dr-jts: Perhaps this should be replaced by a CoordinateSequenceFilter, which would avoid creating a vector completely.

src/geom/prep/PreparedPolygonPredicate.cpp
```diff
@@ -91,6 +91,11 @@ bool
 PreparedPolygonPredicate::isAnyTestComponentInTargetInterior(
     const geom::Geometry* testGeom) const
 {
+    if (testGeom->getGeometryTypeId() == GEOS_POINT) {
```
dr-jts: There's a few of these in `PreparedPolygonPredicate`, right?

Another alternative would be to create a visitor utility class based on `GeometryComponentFilter`, and avoid creating the list of coordinates entirely.

dr-jts: But maybe it's good enough to special-case Point, which is probably the high-value optimization.

dr-jts: I also realize now that this is doing too much work when testing Polygons with holes, since it tests each ring, but only needs to test the polygon shell.

dbaston: I like your filter idea and implemented it in 3cf6bb4.

dbaston: BTW, I'm curious what you think about changing the filter return type from `void` to `bool`, where `true` means "continue iterating" or something.

src/geom/prep/BasicPreparedGeometry.cpp
dr-jts: Wouldn't that still require an `PointEnvelope` object to be instantiated?  Unless I guess `Point` implemented `IEnvelope`...

dr-jts:
This certainly seems like a worthwhile improvement for GEOS.  It does have the effect of drifting it away from JTS somewhat though.

A slight mitigation for this might be to implement the idea of moving the envelope tests into the `Geometry` class (with an efficient override for `Point`).  That would surface the improvement more visibly, and would also let it be reused throughout the library where it makes sense.  And it would provide a better pattern for JTS to adopt when time permits.

Any thoughts on this @dbaston ?

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
```diff
@@ -0,0 +1,104 @@
...
+            return Coordinate(xdist(e), ydist(e));
+        });
+
+        geos::util::Profile sw("GEOSPreparedContains");
+        sw.start();
+
+        size_t hits = 0;
+        auto prep = GEOSPrepare(g);
+        for (const auto& c : coords) {
+            auto seq = GEOSCoordSeq_create(1, 2);
```
dr-jts: @dbaston Any thoughts on the perf hit from creating a new CoordSeq and Point each time? Is it possible to reuse the same CS and Point object, just resetting the XY each time?  Seems like a bit of a hack at the moment, but if it provides a decent improvement then it would be worth formalizing this pattern.  (This I think addresses the issue that PDAL had with using GEOS for PIP).

dbaston: @dr-jts I don't have numbers in front of me ATM, but the hit is significant. I don't think we allow changing a CoordinateSequence after it has been given to a Geometry (for one, wouldn't it make a lazy-constructed Envelope incorrect?). There are two ways I'd like to experiment with on this (both could possibly be adopted):

1) Create a new `CoordinateSequence` implementation that is backed by `std::array` for small values of `n`. This should compile to the equivalent of a single `Coordinate` while providing the hooks we need for various parts of our API (`size()`, filter methods, etc.)
2) Expose some new C API functions along the lines of `GEOSPreparedContainsXY(GEOSPreparedGeometry*, double x, double y)` that avoid creating Point geometries altogether.

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: > BTW, I'm curious what you think about changing the filter return type from `void` to `bool`, where `true` means "continue iterating" or something.

In general, in favour.  But it's a breaking API change, at least in JTS.  Probably not too much external use of this, though, so perhaps worth doing.

Note that `CoordinateSequenceFilter` has an isDone method that supports short-circuiting.  Not sure why I did it that way rather than a boolean return value (perhaps the idea was to make the done state more persistent, to allow easier checking by mulitple clients).  Is it worth continuing this pattern?

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dr-jts: There is the `Geometry.geometryChanged` method that allows invalidating the envelope after a coordinate change.   This pattern is supported in JTS, although it's only used in one place.  

Generally it's encouraged to treat Geometrys as immutable.  But if it's needed for performance this can probably be allowed.  And in this case the mutation can be hidden inside the GEOS API I think, so is less risky.  

Can you hack the performance test on this PR to see if this works, and get some idea if there's a perf improvement?

src/geom/prep/PreparedPolygonPredicate.cpp
dbaston: > Is it worth continuing this pattern?

Yes, calling `isDone()` seems more clear than assigning significance to the return value. I could add this to `CoordinateFilter`.

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dbaston: For that to work, the coordinate change would have to be initiated by the Geometry, right? (with a `filter` method, for example). In the C API we can only modify a coordinate by our handle on the `CoordinateSequence`. And the `CoordinateSequence` does not know which geometry owns it, if any.

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: Sure - nice that we can change GEOS API at will!  

Or... maybe wait until a use case emerges?  For this PR wouldn't it be better to add isDone to `GeometryComponentFilter`?  And then would need to change PreparedPolygonPredicate to use a filter rather than `ComponentCoordinateExtracter`.

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dr-jts: > For that to work, the coordinate change would have to be initiated by the Geometry, right? (with a `filter` method, for example).

Nope, this was designed to support external notification to a Geometry that its coordinates changed.

 In the C API we can only modify a coordinate by our handle on the `CoordinateSequence`. And the `CoordinateSequence` does not know which geometry owns it, if any.

My thinking was that for the prepared PIP use case the client could create a CS and the Point using it, and then reuse both by changing the ordinates of the CS.  Maybe not supported by the C API currently, but would this be possible in C++ now?  

Calling `geometryChanged` after setting the ordinates is the Right Thing To Do.  But really the goal of the changes in this PR are to avoid looking at the Point envelope entirely, right?!  So maybe not strictly necessary?

If this checks out as working, then could expose this in some way via C API.  Perhaps by returning a handle to a context created by C API, which keeps the Point and CS objects around until disposed of.

src/geom/prep/PreparedPolygonPredicate.cpp
dbaston: Well, the use case is here, where I had to essentially implement `isDone` in both of the `CoordinateFilters` I implemented in `PreparedPolygonPredicate`.

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dbaston: Oh, yes, you could make it work in C++ if you manually called `geometryChanged` after changing the `CoordinateSequence`. To do it in C you'd need to expose `geometryChanged` in the C API. My point was that the `CoordinateSequence` can't automatically notify the geometry that it was changed. It seems like a pattern that's better used within the library than as part of the public API.

dr-jts: Fair enough.  So is it possible to work up a C API call(s) which allow reusing Point and CS for say PIP calls?

src/geom/prep/PreparedPolygonPredicate.cpp
```diff
@@ -37,76 +38,79 @@ namespace prep { // geos.geom.prep
 //
 // protected:
 //
+struct AnyMatchingLocationFilter : public CoordinateFilter {
```
dr-jts: Isn't this doing a lot more work than the original use of `ComponentCoordinateExtracter`?  This will test **every** coordinate in the geometry, not just a single point from each component.

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: See my latest comment on this code.  I think using `CoordinateFilter` here is not efficient?

src/geom/prep/PreparedPolygonPredicate.cpp
dbaston: Yes. Fixed.

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dbaston: Yes, though I'd tackle that separately from this PR. I'm inclined to start by making point creation more efficient and looking at reuse strategies only if it remains a bottleneck.

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: And then really need an `GeometryComponentFilter.isDone()` short-circuit.  

Or at least a check for the short-circuit condition in the `filter()` method to avoid calling `locate()` if done.

dbaston: The short-circuit check is there, but would be better moved to `isDone`. Will do this.

dbaston: Hmm @dr-jts , maybe I don't understand `GeometryComponentFilter`. Why does a `GeometryCollection` apply the filter to itself? https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/GeometryCollection.java#L205

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dr-jts: Makes sense.  Great work on this as it is!

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: Actually _every_ Geometry subclass applies `GeometryComponentFilter` to itself first.  

Which is contrary to the `GeometryComponentFilter` doc.  Not sure why this is...  I suspect the idea was that it was easy to filter out complex types if needed.  But maybe this is confusing and should be eliminated.  

In that case I guess this change should use `GeometryFilter` to extract a candidate coordinate from each connected component.  But need to verify that this works correctly and doesn't miss anything.

This is muddying the waters a bit.. sorry bout that.

benchmarks/capi/GEOSPreparedContainsPerfTest.cpp
dbaston: Thanks. By the way, this gives a picture of point construction costs relative to the PIP call.

image

dr-jts: So Point construction is 5%, and running PIP is 25%?  Which means not much point in further optimization of Point construction?

dbaston: Well, 5% to create a coordseq, 3% to create a point, 5% to destroy a point....all in all, 13-14% total of total runtime.

The PIP test itself is 25% of total program runtime, so constructing a point is about half as expensive as the PIP test. I'd say there's room for improvement :)

src/geom/prep/AbstractPreparedPolygonContains.cpp
```diff
@@ -112,9 +112,11 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
     //
     // If a point of any test components does not lie in target,
     // result is false
-    bool isAllInTargetArea = isAllTestComponentsInTarget(geom);
-    if(!isAllInTargetArea) {
-        return false;
+    if (!requireSomePointInInterior || geom->getGeometryTypeId() != GEOS_POINT) {
```
dr-jts: This new test is kinda hard to comprehend, when combined with following negative test.  Am trying to think of how this can be rephrased to be .a bit more intuitive.  

At least a comment would help

dbaston: @dr-jts Agreed; I reworked it in a2030535758d04aa64e6

src/geom/prep/PreparedPolygonPredicate.cpp
```diff
@@ -37,71 +39,116 @@ namespace prep { // geos.geom.prep
...
+
+    algorithm::locate::PointOnGeometryLocator* pt_locator;
+    Location::Value max_loc;
+    bool found;
+
+    void filter_ro(const Geometry* g) override {
+        const Coordinate* pt = g->getCoordinate();
+        auto loc = static_cast<Location::Value>(pt_locator->locate(pt));
+
+        max_loc = std::max(max_loc, loc);
```
dr-jts: Not sure it's good to rely on numeric ordering for locations.

src/geom/prep/AbstractPreparedPolygonContains.cpp
```diff
@@ -109,23 +109,38 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
...
     // If so, the test is contained.
     // If not, all points are on the boundary of the area,
     // which implies not contained.
-    if(requireSomePointInInterior && geom->getDimension() == 0) {
-        bool isAnyInTargetInterior = isAnyTestComponentInTargetInterior(geom);
-        return isAnyInTargetInterior;
+    if (geom->getDimension() == 0) {
+        if (requireSomePointInInterior && outermostLoc == Location::BOUNDARY) {
+            if (geom->getNumGeometries() > 1) {
+                return isAnyTestComponentInTargetInterior(geom);
```
dr-jts: This is the only place this is called, so perhaps would be cleared to dedicate it to checking Points only, and name it appropriately?

tests/xmltester/tests/general/TestPreparedPolygonPredicate.xml
```diff
@@ -18,7 +18,21 @@
 </case>
 
 <case>
-  <desc>A/P - point equal to start point of polygon
+  <desc>A/P - point in polygon interior
```
dr-jts: Good improvements to the test coverage.  I will port these over for sure, in preparation for other improvements.

src/geom/prep/PreparedPolygonPredicate.cpp
dbaston: Yes, I had some hesitation on this. I changed it just now in fa665c2.

A cleaner implementation might be to continue w/`stx::max` but add a compile-time `static_assert` to verify that the locations are ordered in the expected way. I don't think that would be portable to Java, though.

src/geom/prep/AbstractPreparedPolygonContains.cpp
dbaston: Could be renamed, though there's some benefit to having a similar structure for the PreparedPolygonPredicate methods, IMO. Happy to change it if you want to propose something.

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: Maybe good enough to leave as is, and put a note in Location that values should be ordered.

Even better might be a comparison method on Location which highlights that the ordering is expected.

dbaston: Since fa665c2 it no longer depends on ordering.

src/geom/prep/PreparedPolygonPredicate.cpp
```diff
@@ -37,71 +38,118 @@ namespace prep { // geos.geom.prep
...
-{
-    geom::Coordinate::ConstVect pts;
-    geom::util::ComponentCoordinateExtracter::getCoordinates(*testGeom, pts);
-
-    for(std::size_t i = 0, ni = pts.size(); i < ni; i++) {
-        const geom::Coordinate* pt = pts[i];
-        const int loc = prepPoly->getPointLocator()->locate(pt);
-        if(geom::Location::EXTERIOR == loc) {
-            return false;
+struct AnyMatchingLocationFilter : public GeometryComponentFilter {
```
dr-jts: How about calling these `LocationNotMatchingFilter` and `LocationMatchingFilter`?

Shorter, and sorts better.

dr-jts: Could also combine into one filter, with a flag for the sense of the matching test.

dbaston: Done in eb34085

src/geom/prep/AbstractPreparedPolygonContains.cpp
dr-jts: Agreed there is benefit to similar structure for `PreparedPolygonPredicate` methods.

src/geom/prep/AbstractPreparedPolygonContains.cpp
```diff
@@ -109,23 +109,38 @@ AbstractPreparedPolygonContains::eval(const geom::Geometry* geom)
...
+    // of those points lie in the interior of the target
+    // geometry.
+    //
     // If so, the test is contained.
     // If not, all points are on the boundary of the area,
     // which implies not contained.
-    if(requireSomePointInInterior && geom->getDimension() == 0) {
-        bool isAnyInTargetInterior = isAnyTestComponentInTargetInterior(geom);
-        return isAnyInTargetInterior;
+    if (geom->getDimension() == 0) {
```
dr-jts: How about this code?  It puts the short-circuits first, to reduce depth of IF blocks.  Also names the test booleans rather than having them buried in IF conditions.  Easier to understand, I think.

```
    if (geom->getDimension() == 0) {
       // covers evaluation
        if (! requireSomePointInInterior)  return true;

       // contains evaluation
        boolean hasInteriorPoint = false;
        if (geom->getNumGeometries() == 1) {
                // Only  one point
                hasInteriorPoint = outermostLoc == Location::INTERIOR;
            }
        else {
           // for MultiPoint, check if there is at least one point in interior
            hasInteriorPoint = isAnyTestComponentInTargetInterior(geom);
       }
       return hasInteriorPoint;
    }
```

dr-jts: (code should be tested... I think it's right, but inverting all the logic is definitely a bit error-prone)

dr-jts: It might be even clearer to refactor this logic into a separate internal method, which is focussed on just handling points.  This makes the short-circuit for points very obvious.

Like this:

```
if (geom->getDimension() == 0) {
  return evalPointTarget(geom, outermostLoc);
}

boolean evalPointTarget(Geometry geom, Location outermostLoc)
{
       // covers evaluation
        if (! requireSomePointInInterior)  return true;

       // contains evaluation
        boolean hasInteriorPoint = false;
        if (geom->getNumGeometries() == 1) {
                // Only  one point
                hasInteriorPoint = outermostLoc == Location::INTERIOR;
            }
        else {
           // for MultiPoint, check if there is at least one point in interior
            hasInteriorPoint = isAnyTestComponentInTargetInterior(geom);
       }
       return hasInteriorPoint;
}
```

dbaston: I think this formulation subjects MultiPoints to a redundant PIP test if `outermostLoc == Location::INTERIOR`.

dbaston: Refactored in 51c2b64; hopefully this captures what you're after without adding the extra test.

dbaston:
Thank you @dr-jts for the thorough review and many improvements.

dr-jts:
And thanks @dbaston for your hard work and great ideas!

src/geom/prep/PreparedPolygonPredicate.cpp
```diff
@@ -37,71 +38,118 @@ namespace prep { // geos.geom.prep
...
+    Location::Value outermost_loc;
+    bool done;
+
+    void filter_ro(const Geometry* g) override {
+        const Coordinate* pt = g->getCoordinate();
+        auto loc = static_cast<Location::Value>(pt_locator->locate(pt));
+
+        if (outermost_loc == Location::UNDEF || outermost_loc == Location::INTERIOR) {
+            outermost_loc = loc;
+        } else if (Location::EXTERIOR) {
```
dr-jts: As @strk pointed out, this looks like a bug.  Shouldn't there be a mention of BOUNDARY here?

Would be good to find a test case that fails because of this.

And it would probably be nice to name this logic in a separate method (something like `locMax`).  Or move it into the Location class.

dbaston: I do see the bug @strk reported, but I don't think BOUNDARY needs to be mentioned. I'll try to find a test case.

dbaston: Fixed in 67d8a27 with test case. Would be fine to move in to `Location` but I would probably then revert the implementation back to `std::max`.

src/geom/prep/AbstractPreparedPolygonContains.cpp
dr-jts: Great.  Since the code doesn't make it fully clear that there is an ordering relation on Location, it would be better to define this explicitly in the Location class, and then use that here.  Perhaps a method `Location::Index`?  That would allow a simple comparison to be used in this method.

dr-jts: It would also be nice to run all the predicate unit tests through the Prepared code.  Not sure if there's a portable way to do this.  A quick-n-dirty way is just to hack the XMLTester code for each predicate to run both.

src/geom/prep/PreparedPolygonPredicate.cpp
dr-jts: Great. Since the code doesn't make it fully clear that there is an ordering relation on Location, it would be better to define this explicitly in the Location class, and then use that here. Perhaps a method Location::Index? That would allow a simple comparison to be used in this method.

It would also be nice to run all the predicate unit tests through the Prepared code. Not sure if there's a portable way to do this. A quick-n-dirty way is just to hack the XMLTester code for each predicate to run both.

One reason to do this is to validate that the predicate unit tests really have full coverage. If they don't trigger this bug that's an issue that needs to be fixed.

PR #193, opened 2019-06-07
Use C++11 std::chrono for Profiler

dbaston:
This avoids the need for compiler-specific tests and headers, and
anecdotally seems to provide more repeatable measurements.

It should remove a raft of warnings in MSVC builds.


mloskot:
Once I planned to replace TUT with Catch2 (see https://trac.osgeo.org/geos/changeset/4378).
Recently, Catch2 received micro-benchmark features (see https://github.com/catchorg/Catch2/pull/1616).
If TUT is ditched for Catch2 one day, it may be sensible to remake the profiler with the benchmark.

PR #196, opened 2019-06-14
Remove Puntal, Lineal, and Polygonal classes

dbaston:
These classes are used in only a handful of locations to check the
dimension of a Geometry. However, their use in the inheritance hierarchy
increases the class size of MultiPoint, MultiLineString, and
MultiPolygon by 8 bytes, and possibly imposes a runtime cost. This commit
provides alternative methods of checking Geometry dimension by adding 
`isMixedDimension`, `isPuntal`, `isLineal`, and `isPolygonal` to `Geometry.`

If others are OK with the idea of these changes, I will add tests/docstrings
to the new methods.


dr-jts:
The original goal in JTS was to allow methods to express geometry dimension constraints for inputs and outputs explicitly in method signatures.  This makes less sense in C++.  Replacing this with methods on Geometry makes good sense.  (And there's probably more places where the dimension metadata can be checked once this is done).

One simplification might be to have a single method `isStrictDimension(int)` instead of the 3 named methods (or provide this as well).  One thing this does is to make the test more data-driven, which might be handy in some uses.

dr-jts:
In keeping with my obsession with method sorting, perhaps these would be better named as:

`isDimensionMixed()`
`isDimensionStrict(int dim)`

And there may not be an actual use for `isDimensionMixed` if `isDimensionStrict` is provided.  It is sometimes necessary to know if a geometry is a `GeometryCollection`, usually because arbitrary collections of Polygons do not have the same topology constraints as a `MultiPolygon`.  But `isDimensionMixed` doesn't report on this.

dbaston:
The motivation in naming was to try to keep code similar to JTS, so `g instanceof Polygonal`  becomes `g->isPolygonal()`. I'm fine w/introducing a new concept of a "strict dimension" in GEOS but in the interest of minimizing divergence it would probably be good to do the same in JTS.

dr-jts:
Agreed.  I will work on this for JTS soon.  Can make this change in the meantime.

dbaston:
@dr-jts I added `isDimensionStrict` and removed `isDimensionMixed`. Are you OK w/keeping the 3 named methods (`isPuntal`, `isLineal`, `isPolygonal`) ?

dr-jts:
Sure, no harm for now.

PR #197, opened 2019-06-17
Return unique_ptr from WKTReader and WKBReader

dbaston:
This PR updates `WKTReader` and `WKBWriter` to return `unique_ptr` for the geometries they produce.

### DIFF
--- a/src/io/WKBReader.cpp
+++ b/src/io/WKBReader.cpp
@@ -241,25 +241,25 @@ WKBReader::readGeometry()
 
     switch(geometryType) {
     case WKBConstants::wkbPoint :
-        result.reset(readPoint().release());
+        result = readPoint();
         break;
     case WKBConstants::wkbLineString :
-        result.reset(readLineString().release());
+        result = readLineString();
         break;
     case WKBConstants::wkbPolygon :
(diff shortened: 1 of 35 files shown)


src/io/WKBReader.cpp
```diff
@@ -237,29 +237,29 @@ WKBReader::readGeometry()
         ordValues.resize(inputDimension);
     }
 
-    Geometry* result;
+    std::unique_ptr<Geometry> result;
 
     switch(geometryType) {
     case WKBConstants::wkbPoint :
-        result = readPoint();
+        result.reset(readPoint().release());
```
nyalldawson: This looks unusual - any reason why std::move couldn't work here instead?

nyalldawson:
You're on 🔥!!

src/io/WKBReader.cpp
dbaston: Hmm, I didn't think C++11 let me convert from a subclass `unique_ptr` to a base class `unique_ptr`. But I just tried it locally, and it seems to work. Explicit move isn't necessary since `readPoint()` is already an rvalue. Thanks for the catch!

PR #204, opened 2019-06-29
Add FixedSizeCoordinateSequence for Points

dbaston:
The latest release of GEOS (3.7) uses four heap allocations to construct a Point geometry through the C API. Excessive heap allocations have been [identified](https://github.com/PDAL/PDAL/issues/1735) as a weakness of GEOS by @hobu and others.

One of the four allocations was trivially avoided and was removed in aff7bcf25a4879d95243c5a60240a4ee0b4a06a6. This PR removes two of the remaining three.

The three heap allocations are as follows:

1) A `GeometryFactory` allocates all returned Geometries on the heap. This is unavoidable if we need to return the Geometry to a C API caller via an opaque pointer, and is not impacted by this PR. Workarounds are possible, such as providing XY variants of C API functions that avoid the need for the caller to manage a point, e.g. `GEOSPreparedContainsXY(GEOSPreparedGeometry* pg, double x, double y);`

2) A Point heap-allocates a `CoordinateArraySequence` and assigns the value of its first element to the provided x,y. Why is this heap-allocated? GEOS currently allows for multiple `CoordinateSequence` implementations and gives the `GeometryFactory` (through its member `CoordinateSequenceFactory`) control over which implementation is selected at runtime. However, prior to this PR, GEOS:

    - only has a single `CoordinateSequence` implementation, making polymorphism unnecessary.
    - regularly constructs a `CoordinateArraySequence` directly anyway, instead of going to the implementation provided by the `GeometryFactory`.

3) The `CoordinateArraySequence` has a member `std::vector`, which heap-allocates space to store a single `Coordinate`.

This PR eliminates #2 and #3 by:

1) Defining a new `FixedSizeCoordinateSequence` implementation that is backed by `std::array` instead of `std::vector` (removes heap allocation #3). This seems like a no-brainer, since it improves performance, reduces memory usage, and fits into the existing design of GEOS.
2) Uses `FixedSizeCoordianteSequence` for all `Point` geometries and introduces a C API method `GEOSGeom_createPointFromXY` that creates a Point from two doubles (removes heap allocation #2). This may be more controversial, since it goes against the design of allowing a `GeometryFactory` to control how `CoordinateSequence` objects are constructed. However, this design is already widely ignored. Allowing points to have a polymorphic `CoordinateSequence` could be reintroduced in the future if there is a reason to have GEOS construct geometries using some other library's memory. (In that case we would need to construct coordinate sequences through `GeometryFactory::getCoordinateSequenceFactory`.) But we don't have that ability now, so why pay the price?

To measure performance, I ran the `GEOSPreparedContainsPerfTest` under callgrind to perform 100,000 point-in-polygon calls against the boundary of Australia.

In master,

10.4% of the test time is taken constructing and destroying points

With this PR,

3.8% of the test time is taken constructing and destroying points.

As far as storage, a `FixedSizeCoordinateSequence<1>` is 40 bytes. A `CoordinateArraySequence` is 40 bytes plus the size of the underlying array, or a minimum of 64 bytes for a `Point`.

### DIFF
--- a/include/geos/geom/FixedSizeCoordinateSequence.h
+++ b/include/geos/geom/FixedSizeCoordinateSequence.h
@@ -116,6 +116,7 @@ namespace geom {
         void apply_rw(const CoordinateFilter* filter) override {
             std::for_each(m_data.begin(), m_data.end(),
                     [&filter](Coordinate &c) { filter->filter_rw(&c); });
+            dimension = 0; // re-check (see http://trac.osgeo.org/geos/ticket/435)
         }
 
     private:
--- a/include/geos/geom/FixedSizeCoordinateSequence.h
+++ b/include/geos/geom/FixedSizeCoordinateSequence.h
@@ -1,0 +1,129 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2019 Daniel Baston
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************/
+
+#ifndef GEOS_GEOM_FIXEDSIZECOORDINATESEQUENCE_H
+#define GEOS_GEOM_FIXEDSIZECOORDINATESEQUENCE_H
+
+#include <geos/geom/Coordinate.h>
+#include <geos/geom/CoordinateFilter.h>
+#include <geos/geom/CoordinateSequence.h>
+#include <geos/util.h>
+
+#include <algorithm>
+#include <array>
+#include <memory>
+#include <sstream>
+#include <vector>
+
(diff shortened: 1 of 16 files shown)


include/geos/geom/FixedSizeCoordinateSequence.h
```diff
@@ -0,0 +1,129 @@
...
+
+namespace geos {
+namespace geom {
+
+    template<size_t N>
+    class FixedSizeCoordinateSequence : public CoordinateSequence {
+    public:
+        explicit FixedSizeCoordinateSequence(size_t dimension_in = 0) : dimension(dimension_in) {}
+
+        std::unique_ptr<CoordinateSequence> clone() const override {
```
mloskot: IMHO, this should be declared `clone() const final`, unless there really is need to allow deriving new types from `FixedSizeCoordinateSequence` class.

dbaston: Done, thanks. I guess the benefit is to potentially allow compiler devirtualization?

PR #205, opened 2019-07-22
Remove LineSegment virtual destructor

dbaston:
LineSegment controls no resources, so we're not really concerned if its
destructor is called from a hypothetical derived class. Removing the
destructor removes the vtable and decreases the class size.


dbaston:
Needs a bit more thought, maybe keeping the destructor and making it protected.

PR #206, opened 2019-07-24
Port JTS #448

pramsey:
Fix Voronoi polygon generation by adding robust Triangle circumcentreDD


dbaston:
Adds a ~40% performance penalty in my quick test (`bin/perf_voronoi`). I wonder if the DD circumcenter can be used only when needed?

dr-jts:
> Adds a ~40% performance penalty in my quick test (`bin/perf_voronoi`). I wonder if the DD circumcenter can be used only when needed?

A few possibilities come to mind:
* I think the robust circumcentre code is only need for Delaunay triangles which are right isosceles triangles.  So perhaps a filter can be created to detect this condition (up to a tolerance) and fall back to the double-precision calc if not.
* perhaps some close inspection of the circumcentre calculation would reveal a way of creating an effective filter within the computation (e.g. if the `denom` is very large or the `numx` and `numy` are very close)
* maybe very close circumcentres could be detected and snapped together

pramsey:
@dbaston see if the little pre-filter condition I've added takes some of the pain away

dbaston:
> @dbaston see if the little pre-filter condition I've added takes some of the pain away

Yep, performance looks the same as `master` now. Thanks!

pramsey:
Pushed to master

src/triangulate/quadedge/QuadEdgeSubdivision.cpp
```diff
@@ -421,7 +421,10 @@ class
         Triangle triangle(triEdges[0]->orig().getCoordinate(),
                           triEdges[1]->orig().getCoordinate(), triEdges[2]->orig().getCoordinate());
         Coordinate cc;
-        triangle.circumcentre(cc);
+        if (triangle.isIsoceles())
```
dr-jts, line 424: @pramsey It looks like this heuristic doesn't work (mea culpa, since it was my suggestion).  We should remove it, I think.

tests/unit/triangulate/VoronoiTest.cpp
```diff
@@ -204,5 +219,19 @@ void object::test<9>
...
+
+    const char* expected =
+        "GEOMETRYCOLLECTION EMPTY";
+
+    runVoronoi(wkt, expected, 100);
```
dr-jts, line 233: The tolerance here should be much smaller for this test to actually work. Say 0.1, or even 0.

PR #208, opened 2019-08-05
Delaunay performance enhancements

dbaston:
Net benefit is approximately 20% for Delaunay, 2-3% for Voronoi.

- Avoid copying all triangles after construction
- Avoid inFrame checks if result will be ignored
- Avoid heap-allocating std::vector
- Avoid conversions from unique_ptr to raw pointers

### DIFF
--- a/include/geos/geom/CoordinateArraySequence.h
+++ b/include/geos/geom/CoordinateArraySequence.h
@@ -56,6 +56,10 @@ public:
     /// Construct an empty sequence
     CoordinateArraySequence();
 
+    /// Construct sequence moving from given Coordinate vector
+    CoordinateArraySequence(std::vector<Coordinate> && coords,
+                            std::size_t dimension = 0);
+
     /// Construct sequence taking ownership of given Coordinate vector
     CoordinateArraySequence(std::vector<Coordinate>* coords,
                             std::size_t dimension = 0);
--- a/include/geos/triangulate/quadedge/QuadEdge.h
+++ b/include/geos/triangulate/quadedge/QuadEdge.h
@@ -102,6 +102,7 @@ private:
     QuadEdge* next;			  // A reference to a connected edge
     void*   data;
     bool isAlive;
+    bool visited;
 
     /**
      * Quadedges must be made using {@link makeEdge},
@@ -165,11 +166,21 @@ public:
      * @return `true` if this edge has not been deleted.
      */
     inline bool
-    isLive()
+    isLive() const
     {
         return isAlive;
     }
 
+    inline bool
+    isVisited() const
+    {
+        return visited;
+    }
+
+    inline void
+    setVisited(bool v) {
+        visited = v;
+    }
 
     /** \brief
      * Sets the connected edge
--- a/include/geos/triangulate/quadedge/QuadEdgeSubdivision.h
+++ b/include/geos/triangulate/quadedge/QuadEdgeSubdivision.h
@@ -99,6 +99,7 @@ private:
     Vertex frameVertex[3];
     geom::Envelope frameEnv;
     std::unique_ptr<QuadEdgeLocator> locator;
+    bool visit_state_clean;
 
 public:
     /** \brief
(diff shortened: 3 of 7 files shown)


mloskot:
Looks like refactoring can deliver actual optimisation :-) Great effort!

dbaston:
Thanks! This is on top of a previous PR, so Delaunay is now about 40% faster than in 3.7.

dr-jts:
Nice work!

Is the inFrame check applicable to JTS as well?

dbaston:
> Is the inFrame check applicable to JTS as well?

Yes, although it doesn't account for much on its own, maybe a couple of percent. The big wins (here and in previous PR) are from avoiding gratuitous copies, which JTS does not appear to be doing in the first place.

dbaston:
At this point I'm most suspicious of `visitTriangles` / `fetchTrianglesToVisit`. I'm surprised in GEOS to see how much time is spent checking to see whether a `QuadEdge` has been visited, and there's likely a faster way to do this. Also not sure that we need to maintain a stack of `QuadEdge`s to visit, since (in GEOS) we have a list of them all in the first place. (Actually, two lists of all of them, but that's another story)

dr-jts:
> At this point I'm most suspicious of `visitTriangles` / `fetchTrianglesToVisit`. I'm surprised in GEOS to see how much time is spent checking to see whether a `QuadEdge` has been visited, and there's likely a faster way to do this. 

The other way to do this is a `visited` flag on `QuadEdge`.  That's actually in the JTS code, but commented out.  I can't remember why - perhaps I thought reducing space was more important than the (lower in Java) cost of checking inclusion in a `HashSet`.  Perhaps it's worth revisiting this...  

> Also not sure that we need to maintain a stack of `QuadEdge`s to visit, since (in GEOS) we have a list of them all in the first place. 

As does JTS.  Perhaps it would be simpler and faster to just scan the list of `QuadEdge`s rather than using a stack.  The order of triangle building wouldn't be quite as nice, but not sure that matters.  It might also be that in JTS I was trying to allow for concurrent visitors, but that's a minor use case, and even less applicable to GEOS.

dr-jts:
> > The other way to do this is a visited flag on QuadEdge.
> 
> Disadvantage is that we need to reset the flags before each visiting session, and then we have a thread-safety issue too. (As you say though, a minor use case.)

Gaining performance seems like a good tradeoff for thread-safety.  And could always provide a thread-safe alternative if needed.
 
> Since we have the `QuadEdges` in an array, an alternative that would work in GEOS is to create a side array of `visited` flags. You can use pointer arithmetic to find the entry in the side array for any `QuadEdge` in constant time, w/o heap allocation. Not saying it's a good idea, but it's an option.

Would this support reverse lookup from an arbitrary `QuadEdge`?  (Which is the logic needed for the triangle traversal. )

dbaston:
> Would this support reverse lookup from an arbitrary QuadEdge? (Which is the logic needed for the triangle traversal. )

Yes; that's the pointer arithmetic part. But I went ahead and added the `visited` bool to each `QuadEdge` anyway; the class is far from thread-safe as it is. The performance difference is surprising; another ~35% or so. 

Other leads didn't make too much difference. Removing the stack doesn't save a whole lot (and the order it enforces turns out to be important on the Voronoi side). Heap allocations when constructing a QuadEdge don't turn out to be too bad. They can be removed for modest performance gain but not worth it IMO. Inlining some of the common predicates (`rightOf`, `inCircle`) actually made things worse.

Overall though, with a random input of 100,000 points we're now at the following timings (ms)

| GEOS ver  | 3.7  | 3.8 |
| -------- | ---- | --- |
| Delaunay | 1080 | 354 |
| Voronoi  | 1050 | 528 |

dr-jts:
A very nice bump in performance!  Definitely worth dropping the support for thread-safety (which as noted could be provided by additional code if needed in the future).

Not sure offhand why the Voronoi requires the stack-based traversal, but I'll investigate and document this requirement.

It will be interesting to compare this against JTS, and against [Delaunator](https://github.com/delfrrr/delaunator-cpp).  And would also be interesting to see if more sweeping changes could provide even more performance (such as using an alternative data structure to the somewhat complex `QuadEdge`, and perhaps an alternative algorithm to incremental insertion.).

PR #209, opened 2019-08-25
Add envelope test to Distance::segmentToSegment

dbaston:
Trac https://trac.osgeo.org/geos/ticket/584
JTS https://sourceforge.net/p/jts-topo-suite/code/652

### DIFF
--- a/tests/unit/operation/distance/DistanceOpTest.cpp
+++ b/tests/unit/operation/distance/DistanceOpTest.cpp
@@ -554,6 +554,17 @@ void object::test<20>()
     ensure(geos::geom::LineSegment(b0, b1).distance(seq->getAt(1)) < 1e-8);
 }
 
+// JTS testDisjointCollinearSegments
+template<>
+template<>
+void object::test<21>()
+{
+    auto g1 = wktreader.read("LINESTRING (0.0 0.0, 9.9 1.4)");
+    auto g2 = wktreader.read("LINESTRING (11.88 1.68, 21.78 3.08)");
+
+    ensure_equals(g1->distance(g2.get()), 1.9996999774966246);
+}
+
 // TODO: finish the tests by adding:
 // 	LINESTRING - *all*
 // 	MULTILINESTRING - *all*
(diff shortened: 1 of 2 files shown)


tests/unit/operation/distance/DistanceOpTest.cpp
```diff
@@ -554,6 +554,17 @@ void object::test<20>()
...
 
+// JTS testDisjointCollinearSegments
+template<>
+template<>
+void object::test<21>()
+{
+    auto g1 = wktreader.read("LINESTRING (0.0 0.0, 9.9 1.4)");
+    auto g2 = wktreader.read("LINESTRING (11.88 1.68, 21.78 3.08)");
+
+    ensure_equals(g1->distance(g2.get()), 1.9996999774966246);
```
dbaston: @dr-jts I'm confused by the corresponding JTS test: https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/org/locationtech/jts/operation/distance/DistanceTest.java#L46

The correct value is 1.999, no?

dr-jts: That's because the `GeometryFactory` being used in that test file has scale factor = 1 (integer precision).  So the inputs are actually round off to:
```
LINESTRING (0 0, 10 1)
LINESTRING (12 2, 22 3)
```
which produces the value in the test. 

Yes, this is confusing.  That's a very old test file, and it's sub-optimal in a few ways.  For one, i don't see any reason to use a rounding precision model - it should be changed to be full precision.  And it would be nice to refactor at least one of the tests into separate tests.  Also, it seems likely that the unit test is not providing full coverage of the DistanceOp computation (although that will require more thought to resolve).

I'll fix the immediate problem soon, anyway.

dbaston: Ah, I completely missed the scale factor. Thanks! I was just trying to add the test case that was introduced at in r652, and was hung up on why the result wasn't what I'd expect.

You can see what's hit with the GEOS test suite here: https://codecov.io/gh/libgeos/geos/src/81451a44c31eb5045b6026942051312130886bfe/src/algorithm/Distance.cpp

And yes, it looks like there are some gaps.

PR #213, opened 2019-08-27
Improve performance of unary union for many short lines

dbaston:
This PR adds a benchmark for the case of unioning many line segments through unary union, and then improves the performance of that benchmark by roughly 90-95% for the case of 200 lines.


dbaston:
Failing test is `issue-geos-392.xml`, will have to dig into it more. Result is valid, and difference between actual and expected results is EMPTY, yet they fail an equality test. Fun!

dbaston:
@dr-jts I'm tempted to just change the expected result of `issue-geos-392.xml`. Both results are valid (in `master` and in this PR). This test case doesn't pass w/JTS, so I don't have a clear reason to prefer one over the other. Any thoughts?

dr-jts:
> @dr-jts I'm tempted to just change the expected result of `issue-geos-392.xml`. Both results are valid (in `master` and in this PR). This test case doesn't pass w/JTS, so I don't have a clear reason to prefer one over the other. Any thoughts?

Can you look at both outputs in the JTS TestBuilder and see where the differences are?   The latest TestBuilder includes a Diff.diffSegments function which can help here.  Also check that the noding is correct - can use the Noding.findNodePoints function to display nodes (intersections) in the linework.  The output should really be fully noded.

dbaston:
I can't build the latest TestBuilder at the moment...I'm certain it's something on my end. 
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single (make-assembly) on project jts-tests: Execution make-assembly of goal org.apache.maven.plugins:maven-assembly-plugin:2.6:single failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-assembly-plugin:2.6:single: java.lang.ExceptionInInitializerError: null
```

Noding.findNodePoints returns `MULTIPOINT EMPTY` for both version of the output. I'll post an XML of the two outputs for reference. `A` is from `master` and `B` is from this PR. (extension changed because GH "doesn't support" XML) [diff.txt](https://github.com/libgeos/geos/files/3581755/diff.txt)

dr-jts:
I looked at the two geometries in the TestBuilder.  They look very, very close.. Unfortunately the TestBuilder doesn't expose the `equalsExact` with a tolerance, so I can't test that.  But I suspect that would report them as equal.  Can you try that?

PR #215, opened 2019-09-05
Port JTS PR 468

pramsey:
https://github.com/locationtech/jts/pull/468

Refactor line intersection code into Intersection class.
Use new class for all line intersection computation
Unroll input conditioning for performance
Add tests to verify accuracy of conditioned double-precision intersection algorithm

### DIFF
--- a/src/algorithm/LineIntersector.cpp
+++ b/src/algorithm/LineIntersector.cpp
@@ -21,7 +21,7 @@
 #include <geos/algorithm/LineIntersector.h>
 #include <geos/algorithm/Distance.h>
 #include <geos/algorithm/Orientation.h>
-#include <geos/algorithm/HCoordinate.h>
+#include <geos/algorithm/Intersection.h>
 #include <geos/algorithm/NotRepresentableException.h>
 #include <geos/geom/Coordinate.h>
 #include <geos/geom/PrecisionModel.h>
@@ -846,7 +846,7 @@ LineIntersector::intersection(const Coordinate& p1,
                               Coordinate& intPtOut) const
 {
 
-    intersectionWithNormalization(p1, p2, q1, q2, intPtOut);
+    intersectionSafe(p1, p2, q1, q2, intPtOut);
 
     /*
      * Due to rounding it can happen that the computed intersection is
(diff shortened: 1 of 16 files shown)


src/algorithm/LineIntersector.cpp
```diff
@@ -948,81 +908,17 @@ LineIntersector::isInSegmentEnvelopes(const Coordinate& pt) const
 
 /*private*/
 void
-LineIntersector::normalizeToEnvCentre(Coordinate& n00, Coordinate& n01,
-                                      Coordinate& n10, Coordinate& n11, Coordinate& normPt) const
+LineIntersector::intersectionSafe(const Coordinate& p1, const Coordinate& p2,
```
dbaston: While you're in here, maybe consider simplifying signatures to just return a Coordinate.

pramsey: And ignore the wisdom of our forecoders? Don't we want to return-by-reference? sizeof(void) < sizeof(Coordinate) ?

dbaston: Only if it's actually being copied though. My understanding is that return-value optimization makes it unlikely that it will be copied, and the standard requires no copy from C++17 onward. Do I have it right @mloskot ?

strk: Please remember we're targetting C++11 as the minimum, so don't change signature based on C++17 onward features.

mloskot: I think, `return nearestEndpoint(p1, p2, q1, q2);` is canonical here, https://en.cppreference.com/w/cpp/language/copy_elision

Algunenano: I don't think this is a C++17 based change. RAII is a common pattern in modern C++, it's just that compilers (and apparently now mandated by the standard) started getting cleverer and avoid copying returning objects, so the object will be constructed only once.

I think that the change Dan is proposing will really benefit moving GEOS into a more modern C++, which is a good thing.

pramsey: I've added a commit to change the call pattern. See if you like it. It's nice that it's closer to the JTS pattern, except that we have to retain our fun setNull, isNull stuff for Coordinate.

dbaston: Looks good, thanks for indulging me. Not that this PR needs to address it, I wonder if it make sense for the default Coordinate constructor to initialize to "null".

(Another way to address the `isNull` ugliness is to backport C++17 `std::optional`, like we did with the C++14 `std::make_unique`. But last time I looked at that it didn't seem worth it to bring it in.)

Algunenano: > Another way to address the isNull ugliness is to backport C++17 std::optional, like we did with the C++14 std::make_unique

Another less conventional one would be to implement `operator bool` as `!isNull()` so you can do:
```
auto intersection = a.intersection(b);
if (!intersection) {
    throw Error;
}
// Use intersection
```

Anyhow, I prefer std::optional since it allows you not have to deal with errors internally, that is, you can make it impossible to have an invalid/null coordinate. But the issue of being C++17 is there

PR #217, opened 2019-09-10
Add multi-ordinate coordseq setters and getters to C API

dbaston:
From the dumb-but-useful department, this PR adds the following CAPI functions:

```
GEOSCoordSeq_setXY
GEOSCoordSeq_setXYZ
GEOSCoordSeq_getXY
GEOSCoordSeq_getXYZ
```

This allows for marginally less verbose client code, and reduces the time to create GEOS geometries by, unsurprisingly, about 50%. This is rarely a bottleneck, but still reduces total time of a PostGIS query like `SELECT ST_NPoints(ST_Collect(ST_Centroid(geom))) FROM basins_lev05;` by about 15%, so might as well.

This adds a lot of copy-paste to the already out-of-control `geos_ts.cpp`. There was some talk of simplifying that with, e.g., https://github.com/dbaston/libgeos/commit/e271b8dd467c359d090dc221c083effd0fbe2a26. Maybe worth revisiting that separately.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,8 @@ Changes in 3.8.0
   - CAPI: GEOSMakeValid (#952, Even Rouault)
   - CAPI: GEOSPolygonize_valid (#727, Dan Baston)
   - CAPI: GEOSCoverageUnion (Dan Baston)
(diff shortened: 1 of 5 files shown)


dbaston:
closed in  4bc853900102ced

PR #218, opened 2019-09-11
Remove copy-paste error handling in C API

dbaston:
This PR defines some templates that execute a chunk of code (expressed as a lambda), capture any exceptions, and pass them to the context's error handler.

This allows over 3,000 lines of copy-paste to be removed from `geos_ts_c.cpp`, improving readability/maintainability, and ensures a consistent return value for any error occurring within a function.

There is a very small increase in library file size; on my system, `libgeos_c.a` goes from 390K to 402K. Inspecting the libraries with `nm -C` shows that two of the lambdas (`GEOSWKBWriter_writeHEX_R` and `GEOSWKBWriter_write_r`) are not inlined.

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -3,8 +3,9 @@
  *
  * C-Wrapper for GEOS library
  *
- * Copyright (C) 2010-2012 Sandro Santilli <strk@kbt.io>
  * Copyright (C) 2005-2006 Refractions Research Inc.
+ * Copyright (C) 2010-2012 Sandro Santilli <strk@kbt.io>
+ * Copyright (C) 2016-2019 Daniel Baston <dbaston@gmail.com>
  *
  * This is free software; you can redistribute and/or modify it under
  * the terms of the GNU Lesser General Public Licence as published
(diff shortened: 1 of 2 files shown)


capi/geos_ts_c.cpp
```diff
@@ -329,14 +330,76 @@ gstrdup(std::string const& str)
...
+    } catch (...) {
+        handle->ERROR_MESSAGE("Unknown exception thrown");
+    }
+
+    return errval;
+}
+
+// Execute a lambda, using the given context handle to process errors.
+// Return nullptr on error.
+template<typename F, typename std::enable_if<!std::is_void<decltype(std::declval<F>()())>::value, F>::type* = nullptr>
```
dbaston: Compile error in MSVC 2019, C++20 strict:
`D:\a\1\s\capi\geos_ts_c.cpp(359,84): error C2064:  term does not evaluate to a function taking 0 arguments [D:\a\1\s\build.cxx20\geos_c.vcxproj]`

Log: https://dev.azure.com/libgeos/geos/_build/results?buildId=64

dbaston: Possibly related to https://developercommunity.visualstudio.com/content/problem/726778/this-snippet-compiles-totally-fine-on-godbolt-msvc.html

hobu: Follow-up shows they implemented a fix for this in MSVC https://developercommunity.visualstudio.com/comments/744771/view.html

dbaston: Thanks @hobu . It seems though that your suggestion for better error messages (https://trac.osgeo.org/geos/ticket/989) would tip the scales towards a macro-based solution instead?

hobu: Your template solution is nice and clean. I think it should be merged. geos_ts_c.cpp is a mess of copypasta right now.

PR #220, opened 2019-09-13
Add Coordinate::distanceSquared

dbaston:
This can be used to shave some time off minimum distance calculations.

Shaves about 10% off the test suite time for me.

### DIFF
--- a/include/geos/algorithm/CentralEndpointIntersector.h
+++ b/include/geos/algorithm/CentralEndpointIntersector.h
@@ -140,12 +140,12 @@ private:
     findNearestPoint(const geom::Coordinate& p,
                      const std::vector<geom::Coordinate>& pts) const
     {
-        double minDist = std::numeric_limits<double>::max();
+        double minDistSq = std::numeric_limits<double>::max();
         geom::Coordinate result = geom::Coordinate::getNull();
         for(std::size_t i = 0, n = pts.size(); i < n; ++i) {
-            double dist = p.distance(pts[i]);
-            if(dist < minDist) {
-                minDist = dist;
+            double distSq = p.distanceSquared(pts[i]);
+            if(distSq < minDistSq) {
+                minDistSq = distSq;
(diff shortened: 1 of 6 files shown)


strk:
I'm pretty sure I remember having seen, years ago, a user report
which made the squared distance overflow. Please keep this in mind,
in case you can think of a way to handle such case.

dr-jts:
> I'm pretty sure I remember having seen, years ago, a user report which made the squared distance overflow. Please keep this in mind, in case you can think of a way to handle such case.

I don't see how overflow can be more of a problem using dist^2 than dist.  Distance is always computed by first computing the distance^2, and then taking the square root.  So if the distance^2 overflows, the square root will be erroneous as well.

Algunenano:
> I don't see how overflow can be more of a problem using dist^2 than dist. Distance is always computed by first computing the distance^2, and then taking the square root. So if the distance^2 overflows, the square root will be erroneous as well.

A guild scenario, but I'm not sure if it's possible with square root, `(a^2 + b^2)` is computed using 80bit registers and the result overflows a 64b double; then calculating the sqrt before returning could avoid the overflow (provided the cpu used a 80bit sqrt operation).

In any case, that'd be architecture and CPU dependant so I wouldn't consider it important.

dr-jts:
@dbaston is much of the performance boost due to the switch to `std::array` from `std"vector`?

dbaston:
> @dbaston is much of the performance boost due to the switch to `std::array` from `std::vector`?

I don't think it was, but I could be wrong. Sorry for not documenting it at the time.

strk:
> I don't see how overflow can be more of a problem using dist^2 than dist.  Distance is always computed by first computing the distance^2, and then taking the square root.  So if the distance^2 overflows, the square root will be erroneous as well.  

You're right, I guess the case in my memory was about a sum
of distances^2 vs. a sum of net distances (area computation?)

PR #223, opened 2019-09-14
Add DefaultCoordinateSequenceFactory

dbaston:
The previous default, `CoordinateArraySequenceFactory`, always creates a
heap-allocating `CoordinateArraySequence`, even for small sizes. This
commit defines a `DefaultCoordinateSequenceFactory` that delegates to
`FixedSizeCoordinateSequence` for length <= 5. This is big enough to cover
some common cases like conversion of line segments, triangles, and
boxes into geometries.


dbaston:
Applied in bdc41b10bb5f883c3379c4ce09093f5c382aef93

PR #224, opened 2019-09-16
WIP: Add CMake doxygen error check #988

nilason:
This PR adds doxygen error check for CMake builds with a new target (make doxygen-checked), similar to what’s implemented for autotools.

Implemented with only CMake code it’s supposed to work on any platform, but should be tested.

This is a mirror of https://git.osgeo.org/gitea/geos/geos/pulls/95. Opened the PR here too for testing purposes.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -28,14 +28,28 @@ if(BUILD_DOCUMENTATION)
 
     set(srcdir ${CMAKE_CURRENT_SOURCE_DIR})
     set(VERSION ${GEOS_VERSION_FULL})
-    set(DOXYGEN_LOGFILE ${CMAKE_CURRENT_BINARY_DIR}/doxygen_log)
-    configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in
-                   ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile
-                   @ONLY)
+    set(DOXYGEN_LOGFILE ${CMAKE_CURRENT_BINARY_DIR}/doxygen.log)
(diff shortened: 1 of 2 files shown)


doc/CMakeLists.txt
```diff
@@ -28,14 +28,29 @@ if(BUILD_DOCUMENTATION)
...
+        ${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in
+        ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile
+        @ONLY)
     unset(srcdir)
     unset(VERSION)
-    unset(DOXYGEN_LOGFILE)
 
     add_custom_target(docs
-            COMMAND ${DOXYGEN_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile)
+        COMMAND ${DOXYGEN_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile
```
dbaston: What do you think about registering this as a test, like this?

```
    add_custom_target(docs
        COMMAND ${DOXYGEN_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile
        DEPENDS geos
        BYPRODUCTS ${DOXYGEN_LOGFILE})

    add_test(NAME test_docs
             COMMAND ${CMAKE_COMMAND}
                 -D DOXYGEN_LOGFILE=${DOXYGEN_LOGFILE}
                 -P "${CMAKE_CURRENT_BINARY_DIR}/${CHECK_ERROR_SCRIPT}")
```

nilason: Makes sense, a test is after all what it is. Updated as recommended.
Thanks!

PR #225, opened 2019-09-16
Port JTS cascaded union w/ buffer fallback

pramsey:
JTS has implemented a slight update to the cascaded union, that includes testing for edge moves to avoid a special failure case, and a fallback to buffer(0) where unaryunion fails. This improves robustness quite a bit.


strk:
Any test that can be enabled due to this ? There are some that are disabled, and possibly some robustness issues filed on trac to check...

Algunenano:
I must say that all the memory handling (mixing raw pointers and memory deletion with smart pointers)  makes this code (in general, not just the PR) scary as hell. Any chance on uniformizing it and avoid raw pointer usage at all (unless proven beneficial performance wise)?

src/operation/union/CascadedPolygonUnion.cpp
```diff
@@ -230,179 +231,35 @@ CascadedPolygonUnion::unionSafe(geom::Geometry* g0, geom::Geometry* g1)
...
+    geom::Geometry* unioned = unionOp.doUnion();
+    geom::Geometry* justPolys = restrictToPolygons(unioned);
+    delete unioned;
+    return justPolys;
 }
 
-std::unique_ptr<geom::Geometry>
-CascadedPolygonUnion::restrictToPolygons(std::unique_ptr<geom::Geometry> g)
+geom::Geometry*
+CascadedPolygonUnion::restrictToPolygons(geom::Geometry* g)
```
dbaston: Going from managed to raw pointers feels like a move in the wrong direction...

src/operation/union/OverlapUnion.cpp
```diff
@@ -0,0 +1,267 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2019 Daniel Baston <dbaston@gmail.com>
```
dbaston: copy-paste error

dbaston:
> I must say that all the memory handling (mixing raw pointers and memory deletion with smart pointers) makes this code (in general, not just the PR) scary as hell. Any chance on uniformizing it and avoid raw pointer usage at all (unless proven beneficial performance wise)?

The mix of raw and managed is a bit on me...I've been trying to incrementally move toward having everything managed rather than doing one massive commit that modified half of every single file.

pramsey:
I initially started with a bias towards "managed" pointers and found, due to the mixture in other parts of the code base, I ended up thinking just as hard about memory life cycles and ownership (should I release() here? should I get()? where does this go out of scope?) as I did with unmanaged pointers, and where I ended up was the situation I found most understandable, personally.

src/operation/union/OverlapUnion.cpp
```diff
@@ -0,0 +1,267 @@
...
+        return GeometryCombiner::combine(g0, g1).release();
+    }
+
+    std::vector<Geometry*> disjointPolys;
+
+    std::unique_ptr<Geometry> g0Overlap = extractByEnvelope(overlapEnv, g0, disjointPolys);
+    std::unique_ptr<Geometry> g1Overlap = extractByEnvelope(overlapEnv, g1, disjointPolys);
+
+    // std::out << "# geoms in common: " << intersectingPolys.size() << std::endl;
+    std::unique_ptr<Geometry> theUnion(unionFull(g0Overlap.get(), g1Overlap.get()));
```
dbaston: If this throws, `disjointPolys` will be leaked, no?

src/operation/union/OverlapUnion.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+/* private */
+Geometry*
+OverlapUnion::unionBuffer(const Geometry* geom0, const Geometry* geom1)
+{
+    Geometry* copy0 = geom0->clone().release();
+    Geometry* copy1 = geom1->clone().release();
+    std::vector<Geometry*> geoms = {copy0, copy1};
+    const GeometryFactory* factory = copy0->getFactory();
+    std::unique_ptr<GeometryCollection> gColl(factory->createGeometryCollection(&geoms));
```
dbaston: This could result in the `GeometryCollection` constructor trying to delete a pointer to stack memory. Can either create a `GeometryCollection` from `std::vector<std::unique_ptr<Geometry>>` ("new way") or heap-allocate `geoms` ("old way")

src/operation/union/OverlapUnion.cpp
pramsey: Pretty sure they leak no matter what 😂

src/operation/union/OverlapUnion.cpp
```diff
@@ -0,0 +1,267 @@
...
+        bool
+        isGeometryChanged() const override  { return false; }
+
+        void
+        filter_ro(const CoordinateSequence& seq, std::size_t i) override
+        {
+            if (i <= 0) return;
+
+            // extract LineSegment
+            Coordinate p0, p1;
```
dbaston: Should be able to do `Coordinate& p0 = seq.getAt(i-1)` to avoid a copy.

strk:
> I initially started with a bias towards "managed" pointers and found, due to the mixture in other parts of the code base, I ended up thinking just as hard about memory life cycles and ownership (should I release() here? should I get()? where does this go out of scope?) as I did with unmanaged pointers, and where I ended up was the situation I found most understandable, personally. 

I think you should consider the "next developer" here.
Your train of thoughts would not need to be taken again by
the next developer if you show your intentions via .release() or
.get() ...

Algunenano:
> I initially started with a bias towards "managed" pointers and found, due to the mixture in other parts of the code base, I ended up thinking just as hard about memory life cycles and ownership (should I release() here? should I get()? where does this go out of scope?) as I did with unmanaged pointers, and where I ended up was the situation I found most understandable, personally.

I think that comes mostly from the mixing of both types. I don't have much (almost any at all) experience with GEOS code base, but I feel like using shared_ptr in a bunch of places would make things way more clear.

For example (again, I don't have a lot of knowledge of the codebase) if `getGeometryN` returned a shared_ptr<const Geometry> you don't need to do the whole `->clone().release()` which is done in multiple places, you either use the const pointer or, if you need a copy call ->clone() and use that unique_ptr (no need to call release; why use `release` in a pointer that is being deleted straightaway?).

This would require `std::vector<std::unique_ptr<Geometry>> geometries` to be `std::vector<std::shared_ptr<Geometry>> geometries`, but in general I believe it'd make the API and calls cleaner.

Anyhow, this is beyond the scope of the PR and I fee like demands more code on my part, not just comments.

dbaston:
`shared_ptr` is a pretty big change, and not one I think that the GEOS is well-positioned to deal with at the moment. (Example: if a geometry is shared between two collections, what happens when it is modified?) On the other hand, `unique_ptr` has no overhead and captures what GEOS is generally doing anyway (coordinate sequences are exclusively owned by geometries, geometries are exclusively owned by collections, etc.)

Algunenano:
> Example: if a geometry is shared between two collections, what happens when it is modified?

You forbid modification by using `shared_ptr<const T>`; then when you need to change it you to clone it (getting a shared_ptr<T> or a unique_ptr<T>), edit and replace it in the array (safe for the other geometry). This way you just do lazy copying only when needed, and the memory will go away once all users stop referencing it (no need to keep track as long as you avoid circular references).

dbaston:
Definitely worth exploring, @Algunenano .

src/operation/union/CascadedPolygonUnion.cpp
pramsey: Returned to previous state

src/operation/union/OverlapUnion.cpp
pramsey: disjointPolys now managed

pramsey:
With luck this does what we want, basically.

pramsey:
> Any test that can be enabled due to this ? There are some that are disabled, and possibly some robustness issues filed on trac to check...

None of the existing failure/ xml files passes with the new code, but none of them are really supposed to use cascaded union.

pramsey:
Applied in 9e79f3db..6bac4f45

PR #226, opened 2019-09-20
get MSVC 2015+ healthy

hobu:
This gets the cmake-based builds with -DBUILD_SHARED_LIBS=ON healthy

https://trac.osgeo.org/geos/ticket/975


include/geos/geom/Location.h
```diff
@@ -30,7 +30,8 @@ namespace geom { // geos::geom
  * For a description of the DE-9IM, see the
  * [OpenGIS Simple Features Specification for SQL](http://www.opengis.org/techno/specs.htm").
  */
-GEOS_DLL enum class Location : char {
+#undef Location
```
dbaston: What is this `undef` needed for?

dbaston:
Does it make sense to update the AzP file to build the shared library?

include/geos/geom/Location.h
hobu: killed debugging lint in 2387764a226199beb7d5b9bdc5c2cf81960b4f1c

hobu:
I have no idea how and where stuff gets merged, but I think this is ready.

dbaston:
Looks like @pramsey committed it as 1df959cc8a39807e63a09bff

PR #227, opened 2019-09-23
Optimize GEOSIntersects/GEOSIntersection

dbaston:
This PR optimizes performance of the `GEOSIntesects`/`GEOSIntersection` benchmark introduced in 4701553c4b8554f4f1e. With my test data it increases benchmark performance by 35% relative to `master` (45% relative to the 3.7 branch) on a standard CMake Release build.

Gains come from lots of small changes, hence the size of this PR. (Individual commits have a reasonable scope, though.) There are no changes to logic or algorithms. The benefits come, in roughly decreasing order, from

1) Changing the backing collection of `EdgeIntersectionList` from `std::set` to `std::vector` and lazily removing duplicates/sorting before iteration (1200ea5)
2) Removing single-object heap allocations from `SimpleMCSweepLineIntersector`
3) Inlining trivial methods of `ToplologyLocation`, `Label`, `Depth`, and `Envelope`, and others
4) Removing usage of `dynamic_cast` in `STRtree`


include/geos/geom/Coordinate.inl
```diff
@@ -157,6 +139,14 @@ operator!=(const Coordinate& a, const Coordinate& b)
     return ! a.equals2D(b);
 }
 
+INLINE size_t
+Coordinate::HashCode::operator()(const geos::geom::Coordinate &c) const {
+    size_t h = std::hash<double>{}(c.x);
+    h ^= std::hash<double>{}(c.y) << 1;
```
Algunenano: Can you explain why you are shifting the hash here (and the c.z too)? Does that eliminate a bit of information?

dbaston: Just following the example at https://en.cppreference.com/w/cpp/utility/hash. There may be a better way to combine these, but I don't think our performance is terribly sensitive to the hash function.

Algunenano: :ok_hand: Thank you, I was just curious.

dr-jts:
Nice work!

dbaston:
Rebased & merged

PR #229, opened 2019-09-25
Add missing parts of doxygen generated documentation

nilason:
A lot of the nicely documented code never get generated at all or poorly by doxygen. The reason is primarily missing descriptions of namespaces, as well as some classes, but also by missing brief tags.

This PR addresses this issue in two commits. Unfortunately,  I could't come up with a better approach for adding namespace description than adding package/namespace header files (in the same pattern as for other similar cases, e.g. `geos/geosAlgorithm.h`).

### DIFF
--- a/include/geos/algorithm/ConvexHull.h
+++ b/include/geos/algorithm/ConvexHull.h
@@ -45,7 +45,7 @@ class GeometryFactory;
 namespace geos {
 namespace algorithm { // geos::algorithm
 
-/**
+/** \brief
  * Computes the convex hull of a Geometry.
  *
  * The convex hull is the smallest convex Geometry that contains all the
--- a/include/geos/algorithm/Distance.h
+++ b/include/geos/algorithm/Distance.h
@@ -26,7 +26,7 @@
 namespace geos {
 namespace algorithm { // geos::algorithm
 
-/**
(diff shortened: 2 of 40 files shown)


dbaston:
Introducing new header files to document the namespaces seems a bit overkill, but fixing all of the `\brief` stuff is a nice improvement.

nilason:
A closer look at the existing namespace header files, reveals that their present use is in fact mainly for documentation, with a few exceptions (some are for example included in XMLTester.cpp).
My guess is that somewhere on the road it was decided to include only what is needed, if they ever were used for bunch-include in a more general way.
As I see it, either it may be accepted to add these five files, or I squeeze in the namespace docs in a random file where it is declared.
Alternatively, ALL more-or-less unused namespace files may be baked into one file, for instance geos/geom.h. The latter is beyond the scope of this PR though.

nilason:
Sounds good to me, dropped the commit with the added files. May be merged, if you will.

I can have a go for it, dropping the namespace files (at least see what it leads to), I'll put up a new PR. That is, if you're not eager for doing it yourself.

dbaston:
ok, done with a4465831d3fefdc1840ab000868e121ecb9769f3. Thanks!

PR #237, opened 2019-10-08
Reproduce #994

Algunenano:
Reproduces https://trac.osgeo.org/geos/ticket/994


dbaston:
Resolved in  3a9ce5422c929b5. Thank you for extracting the test geometry.

Algunenano:
Thanks a lot, Dan. I can confirm it's fixed now.

PR #238, opened 2019-10-08
Some issues found by scan-build

Algunenano:
Used: `scan-build make`

If you are interested I can try to address the rest of the issues.

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -6268,8 +6268,8 @@ extern "C" {
 
         double length;
         if(GEOSLength_r(extHandle, g, &length) != 1) {
-			return -1.0;
-		};
+            return -1.0;
+        };
         return GEOSProject_r(extHandle, g, p) / length;
     }
 
@@ -6280,8 +6280,8 @@ extern "C" {
     {
         double length;
         if (GEOSLength_r(extHandle, g, &length) != 1) {
-			return 0;
-		}
+            return 0;
+        }
         return GEOSInterpolate_r(extHandle, g, d * length);
     }
 
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -6266,7 +6266,7 @@ extern "C" {
                             const Geometry* p)
     {
 
-        double length;
+        double length = NAN;
         GEOSLength_r(extHandle, g, &length);
         return GEOSProject_r(extHandle, g, p) / length;
     }
@@ -6276,7 +6276,7 @@ extern "C" {
     GEOSInterpolateNormalized_r(GEOSContextHandle_t extHandle, const Geometry* g,
                                 double d)
     {
-        double length;
+        double length = 0;
         GEOSLength_r(extHandle, g, &length);
         return GEOSInterpolate_r(extHandle, g, d * length);
     }
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -6266,8 +6266,10 @@ extern "C" {
                             const Geometry* p)
     {
 
-        double length = NAN;
-        GEOSLength_r(extHandle, g, &length);
+        double length;
+        if(GEOSLength_r(extHandle, g, &length) != 1) {
+			return -1.0;
+		};
         return GEOSProject_r(extHandle, g, p) / length;
     }
 
@@ -6276,8 +6278,10 @@ extern "C" {
     GEOSInterpolateNormalized_r(GEOSContextHandle_t extHandle, const Geometry* g,
                                 double d)
     {
-        double length = 0;
-        GEOSLength_r(extHandle, g, &length);
+        double length;
+        if (GEOSLength_r(extHandle, g, &length) != 1) {
+			return 0;
+		}
         return GEOSInterpolate_r(extHandle, g, d * length);
     }
 
--- a/src/algorithm/MinimumBoundingCircle.cpp
+++ b/src/algorithm/MinimumBoundingCircle.cpp
@@ -300,7 +300,7 @@ Coordinate
 MinimumBoundingCircle::pointWithMinAngleWithSegment(std::vector<Coordinate>& pts, Coordinate& P, Coordinate& Q)
 {
     double minAng = std::numeric_limits<double>::max();
-    const Coordinate* minAngPt = nullptr;
+    const Coordinate* minAngPt = &pts[0];
 
     for(const auto& p : pts) {
         if(p == P) {
--- a/src/io/WKTReader.cpp
+++ b/src/io/WKTReader.cpp
@@ -79,8 +79,8 @@ WKTReader::getCoordinates(StringTokenizer* tokenizer)
     Coordinate coord;
     getPreciseCoordinate(tokenizer, coord, dim);
 
-	std::vector<Coordinate> v;
-	v.push_back(coord);
+    std::vector<Coordinate> v;
+    v.push_back(coord);
 
     nextToken = getNextCloserOrComma(tokenizer);
     while(nextToken == ",") {
--- a/src/io/WKTReader.cpp
+++ b/src/io/WKTReader.cpp
@@ -89,7 +89,7 @@ WKTReader::getCoordinates(StringTokenizer* tokenizer)
         nextToken = getNextCloserOrComma(tokenizer);
     }
 
-    return std::move(coordinates);
+    return coordinates;
 }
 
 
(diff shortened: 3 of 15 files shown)


capi/geos_ts_c.cpp
```diff
@@ -6266,7 +6266,7 @@ extern "C" {
                             const Geometry* p)
     {
 
-        double length;
+        double length = NAN;
```
dbaston: We recently got rid of the NAN macros, so if NAN is useful here maybe use `std::numeric_limits<double>::quiet_NaN()` or whatever it is? I don't think the default value is actually important because I don't think GEOSLength is actually capable of producing an error.

Algunenano: This isn't using `getLength()`, it's using `GEOSLength_r` which can produce errors if the handle or the geometry are NULL.

I've used NAN here because length is then used as a denominator in a function. The other option would be to check the output of `GEOSLength_r` and return an error somehow in `GEOSProjectNormalized_r`, `GEOSInterpolateNormalized_r`; but I don't know what would a proper value be in that case (0?, NAN?).

src/io/WKTReader.cpp
```diff
@@ -89,7 +89,7 @@ WKTReader::getCoordinates(StringTokenizer* tokenizer)
         nextToken = getNextCloserOrComma(tokenizer);
     }
 
-    return std::move(coordinates);
+    return coordinates;
```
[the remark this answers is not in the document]
Algunenano: AFAIK no, the problem is that GCC4.8 isn't clever enough to cast CoordinateArraySequence into its base clase, a CoordinateSequence, and `std::move` is a cast so it's doing that for you. Returning a static_cast<function_return_type> should be enough too.

In any case, why complicate the code with unique_ptr instead of using references and return value optimization?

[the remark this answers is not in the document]
dbaston: `std::unique_ptr<CoordinateSequence>` is not a base class of `std::unique_ptr<CoordinateArraySequence>`, though. The code returns a pointer because GEOS `CoordinateSequences `are always handled as pointers to heap memory because they're polymorphic.

capi/geos_ts_c.cpp
dbaston: Yes, better to check the return of `GEOSLength_r` and propagate the error value of `GEOSProject_r` / `GEOSInterpolateNormalized_r`.

src/io/WKTReader.cpp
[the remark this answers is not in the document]
Algunenano: As far as I can see, `getCoordinates(tokenizer)` is only used in `readLineStringText` and `readLinearRingText`, none of which need the smart pointer at all. If you change `getCoordinates` to return a `CoordinateSequence` or a `CoordinateArraySequence` (no pointers) you can change `readLineStringText` into:
```
std::unique_ptr<LineString>
WKTReader::readLineStringText(StringTokenizer* tokenizer)
{
    return geometryFactory->createLineString(getCoordinates(tokenizer));
}
```

By adding a constructor of `createLineString^ based on move and not a pointer, which will gain ownership of the coordinate sequence:
```
std::unique_ptr<LineString> createLineString( CoordinateSequence&& coordinates) const;
``` 
Which replaces:
```
std::unique_ptr<LineString> GeometryFactory::createLineString(CoordinateSequence::Ptr && newCoords)
```

That way you remove the need of using smart pointers (which IMO is complicating things quite a bit, as you need multiple std::move calls). Any modern compiler should be able to generate more efficient code by avoiding unnecessary copies / moves.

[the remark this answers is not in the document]
dbaston: Eventually you need a pointer, because geometries own a pointer to their coordinates. It can't be a value, because `CoordinateSequence` is abstract. Try it if you want?

capi/geos_ts_c.cpp
Algunenano: Looking at the code it seems that the sane thing to do would be to return -1 in `GEOSProjectNormalized_r` (the return error for GEOSProject_r) and 0 in `GEOSInterpolateNormalized_r` (the return error for GEOSInterpolate_r), right?

dbaston: Yes, that's what I was trying to say.

src/io/WKTReader.cpp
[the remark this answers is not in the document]
dbaston: BTW, `readLineStringText` can already be simplified to `return geometryFactory->createLinearRing(getCoordinates(tokenizer));`. The two-line formulation is just the style of whomever wrote this originally.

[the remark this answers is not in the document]
Algunenano: I see the problem now, I didn't realize CoordinateSequence was an abstract class. So it's either pointers or c++17 variants as far as i can see.

Algunenano:
There is still 3 warnings left. The idea is to first clean these warnings and then add a new CI target that runs scan-build so it's run automatically.

src/algorithm/MinimumBoundingCircle.cpp
```diff
@@ -299,23 +299,23 @@ MinimumBoundingCircle::pointWitMinAngleWithX(std::vector<Coordinate>& pts, Coord
 Coordinate
```
dbaston: Was the dereference even necessary in the first place, or should the function just return `const Coordinate*` ?

Algunenano: You are right, in fact now that I'm looking at it there is a way less intrusive way to handle this which is to just initialize minAngPt to a pointer to the first element, instead of nullptr.

Algunenano: You are absolutely right. All the other functions do the same, should I change the 3 of them?

Algunenano: Now that I look into it, this would involve major changes in the API, changing `extremalPts` to an array of references / pointers and so on, so I'd rather do it in a different PR and leave this one with the copy for now.

dbaston: Sounds good. While you're in there, the other function arguments should probably be `const Coordinate &` instead of mutable...

dbaston:
@Algunenano are you still working here, or should this branch be merged (I'll squash out the back-and-forth commits first, unless you want to)

Algunenano:
> @Algunenano are you still working here, or should this branch be merged (I'll squash out the back-and-forth commits first, unless you want to)

I'm still working on it when I can. There are still a few (2-3?, I don't recall exactly) issues I want to tackle and then I want to try and add the check to the CI.

Algunenano:
I'm currently working on 2 missing things:
- Some warnings in ttmath, which I'm not sure how to fix it or ignore it. I've tried updating the library to its current trunk but the warnings are still there.
- Create a build in Travis running with scan-build. Right now I'm fighting against C++11 flags, which for some reasons it's not getting automatically and the build fails.

dbaston:
> * Some warnings in ttmath, which I'm not sure how to fix it or ignore it. I've tried updating the library to its current trunk but the warnings are still there.

Yes, I also spent some time looking at how to conditionally disable some warnings in this header. It seemed like there should be a way to do it by getting CMake to treat it as a "system" header, but I didn't get it to work. For scan-build, does it make sense to just create a suppressions file?

> * Create a build in Travis running with scan-build. Right now I'm fighting against C++11 flags, which for some reasons it's not getting automatically and the build fails.

I think the longer-term vision is to use AzP, so you might try making the changes there instead.

Algunenano:
> Yes, I also spent some time looking at how to conditionally disable some warnings in this header. It seemed like there should be a way to do it by getting CMake to treat it as a "system" header, but I didn't get it to work. For scan-build, does it make sense to just create a suppressions file?

I'll give a try to `#pragma GCC system_header` and see how it goes, that should get rid of all warnings related to that file and might help scan-build too.

> I think the longer-term vision is to use AzP, so you might try making the changes there instead.

For now I'm using Travis because I already know it and had it activated in my fork, but I'll consider switching once I get everything working.

Algunenano:
scan-build appears to ignore my demands of using `system_header`, but t seems assigning default constructors to ttmath::Big members is enough to please it, so I'll go with that route instead.

Algunenano:
Ok, with this I get a clean build now and there should be a Travis task (https://travis-ci.com/libgeos/geos/jobs/254533462) that only passes if scan-build doesn't throw any warning.

Algunenano:
@dbaston I'd say this is ready for final review now. Let me know if I should improve anything to get it merged.

PR #240, opened 2019-10-15
Add deprecated macro - proposal (addressing #843)

nilason:
This PR add a macro framework for deprecation warnings and disabeling of deprecated definitions, addressing [#843](https://trac.osgeo.org/geos/ticket/843). Consider it a proposal.

Builds and passes tests with both cmake and autotools for me locally. This introduces ~150 warnings, a majority of which is from test cases.

Note - following definitions are disabled by deprecation settings on initial commit for this PR, as they were deprecated before v3.0:
- PrecisionModel(double newScale, double newOffsetX, double newOffsetY)
- PrecisionModel::getOffsetY()
- PrecisionModel::getOffsetX()

**Description**

Deprecated declarations can be attributed with a basic deprecation macro such as `GEOS_DEPRECATED`, which issues a compilation warning. Using a versioned form e.g. `GEOS_DEPRECATED_VERSION_3_1` (in this case the deprecation version is GEOS major version 3 and the minor version is 1) a warning is issued only if `GEOS_DISABLE_DEPRECATED_BEFORE_OR_EQUAL_TO` is set to a version higher than the deprecation version.

By default `GEOS_DISABLE_DEPRECATED_BEFORE_OR_EQUAL_TO` is set to `0x020000`, e.g. <3.0, however it may be re-defined at compilation. The version is defined as a hex in the form `0xMMmmpp`, where MM=major, mm=minor, pp=patch).

Deprecation macros with version checking (e.g. `GEOS_DEPRECATED_VERSION_3_1`) only issues warnings and will not cause build failure.

A *hard* deprecation can be achieved with a `GEOS_DEPRECATED_SINCE(major, minor)` conditional clause. In this case the build will fail, if using a deprecated declaration predating `GEOS_DISABLE_DEPRECATED_BEFORE_OR_EQUAL_TO`.

Defining `GEOS_NO_DEPRECATED_WARNINGS` at compilation will suppress all warnings and prevent eventual build failures. Alternatively, setting `GEOS_DISABLE_DEPRECATED_BEFORE_OR_EQUAL_TO=0x000000` will also prevent build failure, but not suppress warnings.

### DIFF
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -150,6 +150,7 @@ typedef int (*GEOSDistanceCallback)(const void *item1, const void* item2, double
  ***********************************************************************/
 
 #include <geos/export.h>
+#include <geos/deprecated.h>
 
 /*
  * Register an interruption checking callback
@@ -169,12 +170,14 @@ extern void GEOS_DLL GEOS_interruptCancel();
  *     initialize using GEOS_init_r() and set the message handlers using
  *     GEOSContext_setNoticeHandler_r and/or GEOSContext_setErrorHandler_r
  */
+GEOS_DEPRECATED_VERSION_X_3_5("initialize using GEOS_init_r() and set the message handlers using GEOSContext_setNoticeHandler_r and/or GEOSContext_setErrorHandler_r")
 extern GEOSContextHandle_t GEOS_DLL initGEOS_r(
                                     GEOSMessageHandler notice_function,
                                     GEOSMessageHandler error_function);
 /*
  * @deprecated in 3.5.0 replaced by GEOS_finish_r.
  */
+GEOS_DEPRECATED_VERSION_X_3_5("replaced by GEOS_finish_r")
 extern void GEOS_DLL finishGEOS_r(GEOSContextHandle_t handle);
 
 extern GEOSContextHandle_t GEOS_DLL GEOS_init_r();
@@ -465,6 +468,7 @@ extern GEOSGeometry GEOS_DLL *GEOSBufferWithStyle_r(GEOSContextHandle_t handle,
 
 /* These functions return NULL on exception. Only LINESTRINGs are accepted. */
 /* @deprecated in 3.3.0: use GEOSOffsetCurve instead */
+GEOS_DEPRECATED_X("use GEOSOffsetCurve instead")
 extern GEOSGeometry GEOS_DLL *GEOSSingleSidedBuffer_r(
 	GEOSContextHandle_t handle,
 	const GEOSGeometry* g, double width, int quadsegs,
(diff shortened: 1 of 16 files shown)


dbaston:
Given that GEOS says its C++ API is unstable, I wonder if the entire idea of deprecation is a bit over-cautious. We add, remove, and change C++ signatures regularly without keeping older signatures in a deprecated form. That's just me with my cowboy hat though, curious what others think.

But, if others feel a need to keep methods around and mark them as deprecated, I think a simple solution like the one you linked in the ticket is preferable. And whenever we adopt C++14 we can just use [`[[deprecated]]`](https://en.cppreference.com/w/cpp/language/attributes/deprecated).

pramsey:
Has there been an outbreak of deprecations that I was not notified about?

nilason:
@pramsey It' contagious as well, so beware!

First of all, I'd like to state that I take no prestige in implementing the whole package (initial PR commit), in fact I was initially just aiming for the simpler solution you referred to @dbaston. Then for the fun of it and curiosity, I couldn't stop myself. Hold your hat cowboy, here comes the Viking! :)
The simple solution is there, with 2 or 3 macros, just issues warnings. Then there is the rest of it, which is not that much, really, and only to be used if so wanted. Nevertheless, I can remove that part ("the rest") if you wish.
I have no personal interest in a stable C++ API, I'm happy with the C API. But what struck me working through doxygen docs: it is remarkable how very long-ago deprecated stuff remain in spite of unstable C++ API. My primary goal with this PR was to assist to better controlled spring-cleaning now and then.

strk:
I do like having more facilities. And stopping to internally use deprecated functions is a nice thing to do (I'm thinking of the many wrapper functions going form pointers to ref or vice-versa).

I agree we don't need to be too careful about just dropping C++ APIs (we don't _need_ a deprecation warning) but we do have deprecation comments (mostly because we still use those APIs internally) and we do have a (single?) C-API deprecation, so... I'm ok with this change.

dbaston:
It just seems like it's harmless until we need to maintain it....and then we need to dive into 500 lines of macros. Which I really don't want to do. Why not just do https://stackoverflow.com/a/21265197 ?

Algunenano:
Since dropping C++ APIs is not important, why don't you just remove those functions and their usage internally? It is less code thus less maintenance.

dbaston:
I mean, do you want to be trying to decipher [this](https://github.com/libgeos/geos/pull/240/files#diff-f7d9da87a4dea8ec61d1f10cf6e84253R190) when there's a problem that only shows up on @robe2's system? I don't.

strk:
> Since dropping C++ APIs is not important, why don't you just remove those functions and their usage internally? It is less code thus less maintenance.

Because nobody sent a PR to drop those functions ? :)

Note that this PR also has a deprecation warning for C-API, which we'll never drop

pramsey:
YAGNI, I'm not convinced this solves a pressing problem, and adds code in so doing.

nilason:
Well, to be fair it there’s roughly 200 lines of macro, and the compiler detection (and most critical) part of that (~70 lines including some comments) is borrowed from the source stated in the file header and it is not modified (only stripped off of things unrelated to deprecation). I can’t and won’t take any credit for that. For this reason in particular I dare assume it is well tested, and if troubles comes, we are probably not alone.
For the rest of the macro lines, of which none is especially complicated, about a 2/3 is repetitive code.
The remaining almost 280 lines are documentation and the file header. If we drop the version checking parts the number of lines of macro will be halved, and most of the documentation will go as well.

What I want to say with this statistical, boring summary (awfully sorry for that) is although it seems to be a massive chunk of undecipherable code, it is not the case.

@pramsey you may be right this doesn't solve a pressing problem, but I do think that in the end there will be less code and more importantly, the right code.

nilason:
Just out of sheer curiosity and only partially related to this PR, have you given any thoughts on setting C++14 as default or when that might be concidered?  I'm aware such a move will exclude gcc4.8.

dbaston:
Last discussed about a year ago: https://lists.osgeo.org/pipermail/geos-devel/2018-December/008768.html

nilason:
Right, thanks for the link! I suppose the situation didn't change fundamentally since then.

pramsey:
I don't thing we're going to take this at this time.

PR #241, opened 2019-10-18
Improve Delaunay/Voronoi performance

dbaston:
Overall performance improvement of about 25% observed on benchmark. Runtime is now 7x faster than version 3.7.

- Introduce a QuadEdgeQuartet structure to store four QuadEdges. This
reduces the number of memory allocations needed to create the QuadEdges,
and allows them to refer to each other by their relative memory
addresses. (This follows the Graphics Gems implementation.) For now, we
still require QuadEdgeQuartets to have stable addresses, so we store
them in a std::deque. If this requirement can be relaxed, better
performance may come from using a std::vector.
- Reduce the size of a Vertex and a QuadEdge by eliminating unnecessary
virtual methods and associated vtable storage.
- Minor optimization to Vertex::isCCW

### DIFF
--- a/benchmarks/ClassSizes.cpp
+++ b/benchmarks/ClassSizes.cpp
@@ -40,6 +40,9 @@
 #include <geos/constants.h>
 #include <iostream>
 #include <geos/geomgraph/index/SweepLineEvent.h>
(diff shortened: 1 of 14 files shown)


dbaston:
committed as 8b433f0887633e3

PR #243, opened 2019-10-23
Return Point::coordinates to the heap

dbaston:
This re-allows the type of the CoordinateSequence to be polymorphic,
which is necessary for empty points to have an empty CoordinateSequence.

Fixes #1001

PR #245, opened 2019-10-24
Avoid returning size-1 CoordinateSequence from empty point

dbaston:
Fixes https://trac.osgeo.org/geos/ticket/1001 without reintroducing a heap allocation as #243 does.

### DIFF
--- a/src/geom/Point.cpp
+++ b/src/geom/Point.cpp
@@ -41,18 +41,20 @@ using namespace std;
 namespace geos {
 namespace geom { // geos::geom
 
-const static FixedSizeCoordinateSequence<0> emptyCoords;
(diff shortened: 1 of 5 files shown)


src/geom/Point.cpp
```diff
@@ -41,6 +41,7 @@ using namespace std;
 namespace geos {
 namespace geom { // geos::geom
 
+const static FixedSizeCoordinateSequence<0> emptyCoords;
```
sir-sigurd: Its `getDimension()` always returns `3`. But 2D and 3D empty points are the different things :slightly_smiling_face:

dbaston: I guess the 3D one is even emptier than the 2D one. Should be fixed now.

PR #246, opened 2019-11-06
Improve performance of GEOSPreparedContains

dbaston:
A few low-level changes that are surprisingly significant. Performance gains of 25-35% in my benchmark.

### DIFF
--- a/include/geos/index/intervalrtree/SortedPackedIntervalRTree.h
+++ b/include/geos/index/intervalrtree/SortedPackedIntervalRTree.h
@@ -19,6 +19,7 @@
 #include <geos/index/intervalrtree/IntervalRTreeNode.h>
 #include <geos/index/intervalrtree/IntervalRTreeBranchNode.h>
 #include <geos/index/intervalrtree/IntervalRTreeLeafNode.h>
+#include <geos/util/UnsupportedOperationException.h>
 
 // forward declarations
 namespace geos {
@@ -84,7 +85,13 @@ public:
      *
      * @throw IllegalStateException if the index has already been queried
      */
-    void insert(double min, double max, void* item);
+    void insert(double min, double max, void* item) {
+        if(root != nullptr) {
+            throw util::UnsupportedOperationException("Index cannot be added to once it has been queried");
+        }
+
+        leaves.emplace_back(min, max, item);
+    }
 
     /**
      * Search for intervals in the index which intersect the given closed interval
--- a/src/algorithm/locate/IndexedPointInAreaLocator.cpp
+++ b/src/algorithm/locate/IndexedPointInAreaLocator.cpp
@@ -56,6 +56,8 @@ IndexedPointInAreaLocator::IntervalIndexedGeometry::init(const geom::Geometry& g
         addLine(line->getCoordinatesRO());
     }
 
+    index = index::intervalrtree::SortedPackedIntervalRTree(segments.size());
+
     for(geom::LineSegment& seg : segments) {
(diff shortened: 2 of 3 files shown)


dr-jts:
Nice work.

I have had a few further ideas about improving performance in this area in JTS:
* Rework SPRtree to use an array of internal node bound values, rather than a tree of objects (am using the same approach in the Hilbert Packed Rtree, soon to be commited)  
* Possibly eliminate the interval tree altogether and just use the segment index already constructed for detecting segment intersection (with a query using an envelope which is just a line crossing at the query point)
* Can perhaps change to using HPRtree instead of STRtree for segment intersection detection - although performance testing is inconclusive on this so far.

dr-jts:
Is the benchmark code in the repo?

dbaston:
This is also a case where GEOS is paying a pretty high price for abstraction in a tree that (I think) is only used in one place. A leaf node effectively occupies 64 bytes (min and max ordinate values, plus a `LineSegment` (48 bytes)) when it really only needs to be 16-24 bytes (pointers to the two coordinates that define the interval, plus a flag that identifies their relative positions).

Another path to explore (maybe you have already) would be a segment tree.

dbaston:
> Is the benchmark code in the repo?

Yes, run `make perf_geospreparedcontains`. Call it with a WKT file that represents a polygon, and a number of random points to use for the PIP test. e.g. `bin/perf_geospreparedcontains australia.txt 100000`

dbaston:
At this point, runtime appears completely dominated (~ 90% ?) by the sort of the leaf nodes at the start of tree construction. That's surprising to me.

dr-jts:
> At this point, runtime appears completely dominated (~ 90% ?) by the sort of the leaf nodes at the start of tree construction. That's surprising to me.

That does seem a bit surprising.  Does that mean there's not much performance to be gained from changing the tree memory structure?  (Although could save on memory, I guess)

dr-jts:
> This is also a case where GEOS is paying a pretty high price for abstraction in a tree that (I think) is only used in one place. A leaf node effectively occupies 64 bytes (min and max ordinate values, plus a `LineSegment` (48 bytes)) when it really only needs to be 16-24 bytes (pointers to the two coordinates that define the interval, plus a flag that identifies their relative positions).

Agreed, the SPIRtree is overly general considering its use.  Would be good to simplify it.

> Another path to explore (maybe you have already) would be a segment tree.

Haven't looked at segment tree idea.  Isn't it a dynamic data structure?  And might have the same implementation questions as this tree?

dbaston:
We could save a lot of memory by tailoring the tree to this application, and storing a pointer to two coordinates, plus some flag that identifies their relative orientation. That would be 24 bytes/node (unless we could find a clever way to hide the flag). Right now we have values for left and right ordinates (16 bytes), plus a pointer (8 bytes) to a `LineSegment` (48 bytes) = 72 bytes/node.

dbaston:
> Does that mean there's not much performance to be gained from changing the tree memory structure? (Although could save on memory, I guess)

That's what the profiling indicates, but sometimes the profiling is misleading.

> Haven't looked at segment tree idea. Isn't it a dynamic data structure? And might have the same implementation questions as this tree?

I don't know that much about it, it just often comes up in resources about stabbing queries. It should be less vulnerable to bad performance from segments of widely varying sizes. But given that the PIP performance seems so dominated by tree construction, even for tens or hundreds of thousands of point queries, I'm skeptical of getting better overall performance from a tree with more complex construction phase.

PR #249, opened 2019-11-28
Avoid MonotoneChain self-compare in SimpleMCSweepLineIntersector

dbaston:
Improves runtime of GEOSisValid for polygons by about 40%, or specifically:

- from 3.7s to 2.0s for the 5,489 polygons in the HydroBASINS level 5 dataset
- from 610ms to 360ms for the individual polygon components of the boundary of Australia in the GADM country dataset.


dbaston:
To some extent I think this is papering over expensive recursion in `MonotoneChainEdge::computeIntersectsForChain` that may also be hit outside of the self-compare case. I have some ideas about improving that more generally, but this is a much quicker fix.

dr-jts:
Good find, and a nice performance boost.  Also applicable to JTS - I will test the impact on performance there.

dr-jts:
This does make a difference in JTS, for some datasets.  Very data-dependent - most significant for large geometries.

dbaston:
Applied as 7ed0c824b76e212

PR #251, opened 2019-11-30
Fix CMake build without tests

mwtoews:
See https://trac.osgeo.org/geos/ticket/1006


dbaston:
Committed as `fbb18046` (I amended to add "closes" line to commit message). Thanks!

PR #253, opened 2019-12-02
Address clang warnings

Algunenano:
They come from building with scan-build with -Wall and -Wextra and clang version 9.0.0.

### DIFF
--- a/include/geos/geom/FixedSizeCoordinateSequence.h
+++ b/include/geos/geom/FixedSizeCoordinateSequence.h
@@ -37,7 +37,11 @@ namespace geom {
         std::unique_ptr<CoordinateSequence> clone() const final override {
             auto seq = detail::make_unique<FixedSizeCoordinateSequence<N>>(dimension);
             seq->m_data = m_data;
+#if defined __GNUC__ && ((__GNUC__ == 4) && (__GNUC_MINOR__ == 8))
             return std::move(seq); // move needed for gcc 4.8
+#else
+            return seq;
+#endif
         }
 
         const Coordinate& getAt(size_t i) const final override {
--- a/include/geos/geom/PrecisionModel.h
+++ b/include/geos/geom/PrecisionModel.h
@@ -163,6 +163,8 @@ public:
     /// destructor
     ~PrecisionModel(void);
 
+    // copy operator
+    const PrecisionModel& operator=(const PrecisionModel& pm);
 
     /// The maximum precise value representable in a double.
     //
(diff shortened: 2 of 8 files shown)


include/geos/geom/FixedSizeCoordinateSequence.h
```diff
@@ -37,7 +37,11 @@ namespace geom {
         std::unique_ptr<CoordinateSequence> clone() const final override {
             auto seq = detail::make_unique<FixedSizeCoordinateSequence<N>>(dimension);
             seq->m_data = m_data;
+#if defined __GNUC__ && ((__GNUC__ == 4) && (__GNUC_MINOR__ == 8))
```
dbaston: This is pretty gross. I thought the remove was in fact required by the C++11 standard (and not required in C++14?)

include/geos/geom/PrecisionModel.h
```diff
@@ -163,6 +163,8 @@ class GEOS_DLL PrecisionModel {
     /// destructor
     ~PrecisionModel(void);
```
dbaston: Pretty sure we can just un-declare the destructor instead of declaring the assignment operator here.

include/geos/geom/FixedSizeCoordinateSequence.h
Algunenano: Yeah, looking at the build failures it seems that old compilers can't auto cast between the declared and the returned type. I remember I did some trick in my previous PR to handle a similar situation, so I'll try to go that route instead.

dbaston: Here's the reference I was looking for: http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1579

If you're strictly following C++11, it's the older compilers that are correct, and newer compilers that are applying a non-standard extension. I'm surprised that clang generates an error with `-std=c++11`.

Algunenano: I don't see any `-std=c++11` in the autotools build. Maybe that's the way to go as long as the project is based on c++11.

Algunenano: Seems like the main issue is that I don't get --std=c++11 with clang with the current autotools macro, but I get it with the one we have for Postgis, so I'll open a new PR to start with that and build only tackle whatever warnings are left after that.

Algunenano:
Superseeded by https://github.com/libgeos/geos/pull/254. I'll open a different PR after that is addressed as it should remove several of the warnings seen here.

Algunenano:
BTW, thanks a lot for the comments @dbaston. It's always great working with you.

dbaston:
Likewise!

PR #254, opened 2019-12-02
Autotools: Add the std=c++11 to the flags automatically

Algunenano:
Since I was there it also removes several macros that are no longer used.


Algunenano:
Does this change require any improvements to be considered?

dbaston:
I'm not using autotools so not in a position to evaluate it. Maybe @strk could check it out?

strk:
Merged as f5131cf06e2204de0176fae7fe2245548cad04ad

Algunenano:
Thanks!

PR #255, opened 2019-12-06
Additional performance enhancements to GEOSisValid

dbaston:
This PR contains a number of small enhancements that modestly improve the performance of `GEOSisValid`. Improvement relative to master is on the order of 10% for several datasets.

It also adds a benchmark for testing `GEOSisValid` that can be easily extended to test any operation, until such time as a JTSOp equivalent can be built for GEOS.


src/operation/valid/IsValidOp.cpp
```diff
@@ -426,12 +426,16 @@ IsValidOp::checkNoSelfIntersectingRing(EdgeIntersectionList& eiList)
 void
 IsValidOp::checkHolesInShell(const Polygon* p, GeometryGraph* graph)
 {
+    auto nholes = p->getNumInteriorRing();
```
dr-jts: Same optimization can be applied to `checkHolesNotNested`

PR #256, opened 2019-12-06
Use spatial index to check shell nesting in IsValidOp

dbaston:
This PR introduces an `IndexedNestedShellTester` class, modelled after
`IndexedNestedRingTester`, to check shell nesting as part of MultiPolygon
validity checking. Relevant methods are yanked out of `IsValidOp`
then updated to use indexes to locate rings relative to each other 
(STRtree) and to check point containment within a ring
(IndexedPointOnAreaLocator).

You need the right geometry for this to matter.

This commit **reduces GEOSisValid runtime for a test geometry (GADM
boundary of Australia) from about 85 seconds to less than 1 second**. It
does not appreciably change runtime for TIGER counties, HydroBASINS
Level 5, or a sample parcel dataset.

### DIFF
--- a/include/geos/algorithm/locate/IndexedPointInAreaLocator.h
+++ b/include/geos/algorithm/locate/IndexedPointInAreaLocator.h
@@ -104,6 +104,10 @@ public:
      */
     IndexedPointInAreaLocator(const geom::Geometry& g);
 
+    const geom::Geometry&  getGeometry() const {
+        return areaGeom;
+    }
+
     /** \brief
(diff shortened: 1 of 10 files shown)


dr-jts:
Wow, nice performance bump.  

This is relevant to MultiPolygons with large numbers of element Polygons, right?

dbaston:
> This is relevant to MultiPolygons with large numbers of element Polygons, right?

Yes. Australia sees such a huge benefit because it has so many islands that have to be tested against the highly-complex shell for the mainland.

PR #258, opened 2019-12-13
Install inl files if DISABLE_GEOS_INLINE is OFF

gergondet:
Currently, if `DISABLE_GEOS_INLINE` is `OFF` (the default value) then the `.inl` files are not installed along which causes an issue for consumers down the line.

This PR fixes that behaviour by installing the `.inl` files if required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,6 +309,13 @@ install(DIRECTORY
   "${CMAKE_CURRENT_BINARY_DIR}/include/geos"
   DESTINATION include
   FILES_MATCHING PATTERN "*.h")
+if(NOT DISABLE_GEOS_INLINE)
+  install(DIRECTORY
+    "${CMAKE_CURRENT_LIST_DIR}/include/geos"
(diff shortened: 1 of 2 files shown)


dbaston:
If inlines are enabled, then the `.inl` files are effectively `.cpp` files, no?

gergondet:
> If inlines are enabled, then the .inl files are effectively .cpp files, no?

Yes. But it doesn't change the fact that they need to be included for the installed version to work.

The headers look like this:
```cpp
/* some declaration */

#ifdef GEOS_INLINE
#include "SomeHeader.inl"
#endif
```

Note that they used to be installed but this was removed by the modern CMake patch: https://github.com/libgeos/geos/commit/1aedfca497d6de23bd3a857191da10b98b6b01f0#diff-46193e2e675adcde81f1d095d8b3e1dbL34 maybe @mloskot can give more information about that.

dbaston:
Sorry, I got confused by the double-negative of `DISABLE_GEOS_INLINE=OFF`.

dbaston:
Also cherry-picked to 3.8 as fe6d9701a5923

PR #263, opened 2019-12-18
Docs update

nilason:
This PR addresses two independent issues with doxygen documentations:

- adds documentation to `IndexedFacetDistance`

- removes the doxygen tags `PERL_PATH` and `MSCGEN_PATH` from Doxyfile.in. The tags are with Doxygen 1.8.16 obsolete, and have not been used in GEOS docs anyway. They will most probably not be missed with earlier versions either.

I'm not sure though, whether the IndexedFacetDistance class description statement that the class is thread-safe also is true for GEOS.

Update:
Additionally, I corrected some bad comment indentation and added some missing brief tags.

### DIFF
--- a/include/geos/operation/distance/IndexedFacetDistance.h
+++ b/include/geos/operation/distance/IndexedFacetDistance.h
@@ -24,19 +24,78 @@
 namespace geos {
 namespace operation {
 namespace distance {
+
+/// \brief Computes the distance between the facets (segments and vertices)
+/// of two [Geometrys](\ref geom::Geometry) using a Branch-and-Bound algorithm.
+///
+/// The Branch-and-Bound algorithm operates over a traversal of R-trees built
+/// on the target and the query geometries.
+///
+/// This approach provides the following benefits:
+///
+/// - Performance is dramatically improved due to the use of the R-tree index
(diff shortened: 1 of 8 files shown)


include/geos/operation/distance/IndexedFacetDistance.h
```diff
@@ -24,19 +24,80 @@
...
+///   and the pruning due to the Branch-and-Bound approach
+/// - The spatial index on the target geometry is cached which allow reuse in
+///   an repeated query situation.
+///
+/// Using this technique is usually much more performant than using the
+/// brute-force \ref geom::Geometry::distance(const Geometry* g) const when one
+/// or both input geometries are large, or when evaluating many distance
+/// computations against a single geometry.
+///
+/// This class is thread-safe.
```
nilason: As copied from JTS. But is this the case for GEOS too?

dbaston: Maybe? But I don't think we need to make that claim.

nilason: Ok, I removed that line.

dbaston:
Appreciated as always, @nilason . Thanks!

PR #264, opened 2019-12-26
Allow re-use of IndexedFacetDistance by exposing creator, destroyer, and distance method to c api

nyalldawson:
If deemed the correct approach to use, I'll go ahead and add unit tests...

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -21,6 +21,7 @@
 #include <geos/io/WKBReader.h>
 #include <geos/io/WKTWriter.h>
 #include <geos/io/WKBWriter.h>
+#include <geos/operation/distance/IndexedFacetDistance.h>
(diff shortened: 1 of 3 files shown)


dbaston:
Hi Nyall,

Thanks for putting this together. I was talking with @dr-jts about this a week or two ago. What do you think about the idea of storing a lazily-initialized `IndexedFacetDistance` within a `PreparedGeometry` and then exposing this at the CAPI as `GEOSPreparedDistance` ? (or maybe `GEOSPreparedBoundaryDistance`, which would cover the case of distance between a test point and a polygon boundary). This would allow us to keep the C API a little higher-level and leave us free to implement various distance optimizations on prepared geometries in the future.

nyalldawson:
@dbaston ok, I'll take a look...

pramsey:
Trying a new approach?

nyalldawson:
This is redundant now

PR #266, opened 2020-01-07
Minor debug code fixes

kartynnik:
- Fixes a compile-time error when `GEOS_DEBUG` is defined.
- Fixes redefining `DEBUG_INTERSECT` with 0 regardless of its value.


dbaston:
Thanks!

PR #267, opened 2020-01-09
Provide location for GEOSNode failure

dbaston:
This PR modifies `IteratedNoder` to provide an intersection coordinate for a failed noding, e.g.

`TopologyException: Iterated noding failed to converge after 6 iterations (near -8.064e+06 4.99333e+06)`

It also resolves a memory leak for the error case.

cc @mappingvermont


mappingvermont:
Oh wow! Ask (a vague question on SO) and ye shall receive! Thanks so much @dbaston!

PR #268, opened 2020-01-10
Rename geos::M_PI to less intrusive MATH_PI 

mkrivos:
That don't collide with standard #include <cmath> define.
#undef M_PI public headers is not good idea IMHO.


dbaston:
Makes sense to me. Would need to rename all usages, though.

mkrivos:
Done.

dbaston:
Committed as 089f87756b5230b61c34648c3126b8dca9b98c32, thank you.

PR #269, opened 2020-01-15
Fix memory leak in GEOSLineMerge

dbaston:
Fixes https://trac.osgeo.org/geos/ticket/1010

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -1893,9 +1893,9 @@ extern "C" {
             LineMerger lmrgr;
             lmrgr.add(g);
 
-            std::vector<LineString*>* lines = lmrgr.getMergedLineStrings();
+            auto lines = lmrgr.getMergedLineStrings();
 
-            auto out = gf->buildGeometry(lines->begin(), lines->end());
+            auto out = gf->buildGeometry(std::move(lines));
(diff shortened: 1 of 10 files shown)


Algunenano:
I haven't looked at the ramifications around the changes (the move implementation depending on the size doesn't seem necessary in my head, but I don't know the code so I trust Dan), but I can confirm this fixes the leak in GEOSLineMerge we are seeing in Postgis unit tests.

dbaston:
Yes, this was a rather lean commit message for so many changes. I'll improve it when I fix the MSVC issue.

> the move implementation depending on the size doesn't seem necessary in my head

I think you're referring to `buildGeometry` ? The semantics of this method are to return the simplest form possible, so for `size() == 1` we just return the input and for `size() > 1` we construct a collection of the proper type.

Algunenano:
> I think you're referring to buildGeometry ? 

Yes, detecting empty, 1 geom, or 1+ geoms seemed excessive, but it makes sense if it's done to use the "lowest" type possible.

dbaston:
Looks like I botched the merge. Applied to master as 24f22e720ebcd3ed12d1651e539f44f774f3c5ef and 1837d3b74477ccd5624fe578a582f455bab310d1

PR #270, opened 2020-01-16
remove whitespace at the end of capi version

azhi:
Previously version included git commit hash at the end.
Git commit hash was [removed](https://github.com/libgeos/geos/commit/7699310f0bc7b3f2a6f300e72164d3fab3e76646#diff-8dd96440c9832c59176dcaa7c346b982R3549), but whitespace was left over.
It breaks ruby ffi-geos library that tries to parse capi version using regex :).

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -1931,7 +1931,7 @@ extern "C" {
     const char* GEOSversion()
     {
         static char version[256];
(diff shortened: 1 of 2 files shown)


dbaston:
Thanks! Not sure if you're handy with C++ regex, but I think there's also a need to add a test showing that `GEOSversion()` follows the expected format. There have been a few cases of unexpected changes in this format breaking downstream projects.

PR #274, opened 2020-02-12
Use !std::isfinite() to exclude nan and infinite values

rouault:
- std::isnan() is useless when !std::isfinite() is tested
- std::isnan() || std::isinf() can be replaced by !std::isfinite()


rouault:
Build failures seem unrelated to my changes: ```Cannot add PPA: 'ppa:~ubuntu-toolchain-r/ubuntu/test'```

dbaston:
Thanks!

PR #275, opened 2020-02-13
Fix a few CLang 9 -Wextra warnings

rouault:
- Unused private member --> removed
- Unused method (likely a debugging one) -> protected by #if GEOS_DEBUG
- Unused variables in a destructor -> removed

### DIFF
--- a/include/geos/noding/NodingIntersectionFinder.h
+++ b/include/geos/noding/NodingIntersectionFinder.h
@@ -60,8 +60,7 @@ public:
         interiorIntersection(geom::Coordinate::getNull()),
         intersectionCount(0),
         isCheckEndSegmentsOnly(false),
-        findAllIntersections(false),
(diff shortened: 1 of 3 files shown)


include/geos/noding/NodingIntersectionFinder.h
```diff
@@ -60,8 +60,7 @@ class NodingIntersectionFinder: public SegmentIntersector {
         interiorIntersection(geom::Coordinate::getNull()),
         intersectionCount(0),
         isCheckEndSegmentsOnly(false),
-        findAllIntersections(false),
-        keepIntersections(true)
```
dbaston: Probably best resolved at https://github.com/libgeos/geos/blob/master/src/noding/NodingIntersectionFinder.cpp#L149, but I guess no harm removing it in the meantime.

PR #278, opened 2020-02-13
Add a geos::detail::down_cast<To>(From*) function

rouault:
(Ported from what has been used since a few years in GDAL, which was
added per https://trac.osgeo.org/gdal/ticket/6720)

This function can be used to replace most patterns where a static_cast
or dynamic_cast is done, and we are sure that the result is of the target
type (except programming bugs)

In particular, the following common pattern
```
assert(dynamic_cast<Target*>(foo));
Target* t = static_cast<Target*>(foo);
```
can now been replaced equivalently by
```
Target* t = detail::down_cast<Target*>(foo);
```
The assertion is only done in GEOS_DEBUG mode by down_cast.

This is even safer as it avoids the risk of potentially side effect,
in the assert() as was found in RelateComputer.cpp that had
```
            assert(dynamic_cast<RelateNode*>(nodes.addNode(ei.coord)));
            RelateNode* n = static_cast<RelateNode*>(nodes.addNode(ei.coord));
```

### DIFF
--- a/include/geos/util.h
+++ b/include/geos/util.h
@@ -32,6 +32,7 @@
 //#include <geos/util/math.h>
 
 #include <memory>
+#include <type_traits>
 
 //
 // Private macros definition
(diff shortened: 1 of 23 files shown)


dbaston:
Looks like a good idea to me. Only nit would be to put it in the `geos::detail` namespace, like the `make_unique` templates defined in the same header.

rouault:
>  to put it in the `geos::detail`

I considered it, but geos::detail::down_cast<> seemed to verbose to me.

dbaston:
GEOS code is in `geos` namespace already, so the calls would be `detail::down_cast<>`. There's some discussion of the "why" in #178.

rouault:
> so the calls would be `detail::down_cast<>`

ok, I just force pushed a change doing that

dbaston:
Thanks!

PR #279, opened 2020-02-13
Add final keyword in class declarations wherever possible

rouault:
Two advantages:
- Helps the compiler to generate better code by de-virtualizing
  some calls
- Better document the inheritance diagram

Also tag with a /*non-final*/ comment classes that are later derived.

Candidates found with 'grep ": public" . | grep final"

### DIFF
--- a/include/geos/algorithm/NotRepresentableException.h
+++ b/include/geos/algorithm/NotRepresentableException.h
@@ -32,7 +32,7 @@ namespace algorithm { // geos::algorithm
  * @version 1.4
  * @see HCoordinate
  */
-class GEOS_DLL NotRepresentableException: public util::GEOSException {
+class GEOS_DLL NotRepresentableException final: public util::GEOSException {
 public:
     NotRepresentableException();
     NotRepresentableException(std::string msg);
--- a/include/geos/algorithm/SimplePointInRing.h
+++ b/include/geos/algorithm/SimplePointInRing.h
@@ -31,7 +31,7 @@ class CoordinateSequence;
 namespace geos {
 namespace algorithm { // geos::algorithm
 
-class GEOS_DLL SimplePointInRing: public PointInRing {
+class GEOS_DLL SimplePointInRing final: public PointInRing {
 public:
     SimplePointInRing(geom::LinearRing* ring);
     ~SimplePointInRing() override = default;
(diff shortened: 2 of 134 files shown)


dbaston:
Curious if you measured a performance difference on any of the benchmarks? If not, I can give it a try later. I've had this change in the back of my mind for a while, and am inclined to think it's a good idea, but I have some hesitation about applying it without measurement because of the potential to break any external C++ code that extends a GEOS class.

rouault:
> Curious if you measured a performance difference on any of the benchmarks?

No. I wouldn't expect anything dramatic. More a good practice than anything

> because of the potential to break any external C++ code that extends a GEOS class.

Isn't that the purpose of them to define -DUSE_UNSTABLE_GEOS_CPP_API ;-) ? More seriously, yes, there is a small risk. I'd say let's try and revert the problematic parts if/when people yell. Are you aware of external code to GEOS that would use the C++ API and extend core clases ?

dbaston:
If it successfully enables more devirtualization of `CoordinateArraySequence::getAt`, it ought to be measurable if not dramatic. I'll give it a try.

Agree that committing and partially reverting on request seems reasonable. I'm not aware of external code extending the C++ API.

hobu:
> Are you aware of external code to GEOS that would use the C++ API and extend core clases ?

I think we should assume this is quite likely. I don't think we should `final` the class hierarchy. GEOS already acts as bad citizens for C++ developers.

dbaston:
I tried two benchmarks (perf_voronoi and perf_intersection) and both perform about 5% worse under the `final` branch. Not what I would have expected. This is on a release build with gcc 7.4. Haven't tried other compilers, or looking at at actual devirtualization w/callgrind.

rouault:
> perform about 5% worse

ok, that kills the whole idea :-( Probably that the devirtualization results in some inlining that has adverse effect on the instruction cache.

For the record, which optimization level and CPU did you use?

Closing.

dbaston:
Looks like CMake is using `-O3` for release.

CPU info is
```
vendor_id	: GenuineIntel
cpu family	: 6
model		: 60
model name	: Intel(R) Core(TM) i7-4700MQ CPU @ 2.40GHz
stepping	: 3
microcode	: 0x27
cpu MHz		: 2340.394
cache size	: 6144 KB
physical id	: 0
```

PR #280, opened 2020-02-13
tests: geos_unit: fix missing include

metux:
Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>


pramsey:
Same Q as #281, what's the actual error, because cstdlib is already there...

PR #282, opened 2020-02-15
Avoid adding small segment from some buffer operations

mwtoews:
This bug, [described on Trac](https://trac.osgeo.org/geos/ticket/743), has been around since probably the first releases of GEOS, and is also in JTS master.

Some tests needed to be modified, as there are sometimes fewer points with the revised logic.

This change will visible downstream, for instance shapely uses a default quadsegs to 16, which has this rounding error for buffered points. Other geometries and quadseg combinations may yield fewer coordinates.

Fixes [#743](https://trac.osgeo.org/geos/ticket/743)

### DIFF
--- a/src/operation/buffer/OffsetSegmentGenerator.cpp
+++ b/src/operation/buffer/OffsetSegmentGenerator.cpp
@@ -286,7 +286,7 @@ OffsetSegmentGenerator::addFillet(const Coordinate& p, double startAngle,
 
     double currAngle = initAngle;
     Coordinate pt;
-    while(currAngle < totalAngle) {
+    while(currAngle < totalAngle - 1e-12) { // avoid adding tiny segment
         double angle = startAngle + directionFactor * currAngle;
         pt.x = p.x + radius * cos(angle);
         pt.y = p.y + radius * sin(angle);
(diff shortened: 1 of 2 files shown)


dr-jts:
Seems like a good fix.  I will apply it to JTS as well.

mwtoews:
> Seems like a good fix. I will apply it to JTS as well.

Thanks, I had a quick look at JTS to see the same logic in `OffsetSegmentGenerator.java`, but I'm out of my league when it comes to Java development.

mwtoews:
And yup, this bug had been around for 17 years ([seen here](https://github.com/libgeos/geos/blob/7e662562d8485b11d0fdba7273079123b8101269/source/operation/buffer/BufferLineBuilder.cpp#L342)), but has only been bothering me for 5 years.

dr-jts:
JTS has logic to remove very small segments, but only for curves generated for non-point geometries.  The point buffer generation uses the same code to generate "fillets", but for a full circle.  This is why there is not check for short segments in the fillet code itself.

An alternative to your patch is to switch to using a better generator for circles, which would just generate the circle with exactly the desired number of points.  I might consider implementing that instead (just because it's more direct logic, and stays out of the way of the code dropping short segments).

mwtoews:
@dr-jts feel free to supersede this PR with something better!

Algunenano:
I remember fixing something like this in Postgis with a simpler solution: instead of calculating the increment and adding it in each iteration (which increases errors) you can keep the first and last point in the angle (already known) and calculate each intermediate point as n/segments * angle. It's going to require some extra ops per iteration but the error doesn't accumulate.

dr-jts:
See [JTS PR 526](https://github.com/locationtech/jts/pull/526). Note it uses the approach suggested by @Algunenano.

pramsey:
Closing this in favour of #296

PR #300, opened 2020-03-30
Remove redundant EditorConfig properties

gszy:
The `insert_final_newline` and `trim_trailing_whitespace` properties do not have to be enabled for `CMakeLists.txt` as they are already enabled for all files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -37,8 +37,6 @@ end_of_line = crlf
 [CMakeLists.txt]
 indent_style = space
 indent_size = 2
-trim_trailing_whitespace = true
-insert_final_newline = true
 
 # CMake modules
 [*.cmake]


strk:
I merged this change in the branch. Not sure why GitHub fails to figure this out.
Better use Gitea for PRs ...

dbaston:
@strk again, you didn't merge this, you rebased it.

strk:
On Tue, Mar 31, 2020 at 06:48:05AM -0700, Dan Baston wrote:
> @strk again, you didn't merge this, you rebased it.

Or cherry-picked. To me "merge" is a general term, I should have used
"accept" ("I accepted it", "I pushed it to the official repository").

What happened is someone sent a change request (patch) and I applied it.

In pre-git times we used to be the "committers" and we'd attribute the
patch to the author. How to do this with git is to be decided on (no,
I'm still not sold on "merge commits")

PR #302, opened 2020-04-08
Enforce utf-8 charset when building astyle

gergondet:
I encountered a build issue while attempting to build geos with MSVC on a PC with a Chinese locale. I found this [comment](https://github.com/libusb/libusb/issues/207#issuecomment-288664124) on a similar problem that points to a deeper explanation of the issue and a solution :)

This PR enables the `/source-charset:utf-8` for the `astyle` target in CMake which solves this particular issue.


pramsey:
Merged, thank you!

PR #303, opened 2020-04-15
Use JTS doubledouble implementation instead of ttmath

pramsey:
The ttmath library is quite large and full of fancy stuff like ASM and so on, and all we seem to need is doubledouble. Also ttmath has some issues on big-endian systems. Also, this implementation seems to be 5-10% faster than ttmath, and I haven't exactly fine tuned it. I think with some added unit tests from JTS we should consider dropping ttmath entirely, and maybe even back-porting this into 3.8, since there is no regression change at all (!!). (Maybe, at least so far testing only on my machine.)


include/geos/math/DD.h
```diff
@@ -0,0 +1,126 @@
...
+        friend DD operator/ (const DD &lhs, const DD &rhs);
+        friend DD operator/ (const DD &lhs, double rhs);
+
+        static DD determinant(const DD &x1, const DD &y1, const DD &x2, const DD &y2);
+        static DD determinant(double x1, double y1, double x2, double y2);
+        static DD abs(const DD &d);
+        static DD pow(const DD &d, int exp);
+        static DD trunc(const DD &d);
+
+        bool isNaN() const;
```
dbaston, line 92: Many of these methods seem like good candidates for inlining (either by defining them where they're declared, or by using a `.inl` file like much of the older work)

include/geos/math/DD.h
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * \class DD
+ *
+ * \brief
+ * Wrapper for DoubleDouble higher precision mathematics
+ * operations.
+ */
+class GEOS_DLL DD {
+    private:
+        double SPLIT = 134217729.0; // 2^27+1, for IEEE double
```
dbaston, line 32: This can be `constexpr` and will reduce `sizeof(DD)`

include/geos/math/DD.h
pramsey: Mostly they don't get called except in the tests. Most of the work happens in the arithmetic operators and selfAdd/selfMultiply.

dr-jts: Seems like just about all the methods should be inlineable, no?  Would be interesting to see the performance diff.

PR #304, opened 2020-04-16
Port JTS DD and bring as an option to 3.8

pramsey:
Because 3.8 is a stable branch, this is a big change, but 3.8 is currently broken on big endian architectures, so this patch fixes that, by allowing the DD implementation to replace ttmath for *just those architectures* on the 3.8 stable branch.


pramsey:
References https://trac.osgeo.org/geos/ticket/1024

PR #314, opened 2020-04-28
Add envelope check to DistanceOp::computeMinDistance

dbaston:
Reduces `GEOSDistance` runtime by 70% on a test set of polygons.

### DIFF
--- a/src/operation/distance/DistanceOp.cpp
+++ b/src/operation/distance/DistanceOp.cpp
@@ -447,7 +447,7 @@ DistanceOp::computeMinDistance(
 
         Envelope segEnv0(p00, p01);
 
-        if (segEnv0.distance(lineEnv1) > minDistance) {
+        if (segEnv0.distanceSquared(lineEnv1) > minDistance*minDistance) {
             continue;
         }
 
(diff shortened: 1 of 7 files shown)


dr-jts:
Nice!  That works in JTS as well - will make the change.

dr-jts:
@dbaston I also tried adding a short-circuit after the first `for` statement, checking the distance from the segment envelope to the other line envelope.  That seemed to produce a further 50% reduction in my test data.  What do you think?

dbaston:
Confirmed, @dr-jts . Now looking at overall 90% improvement in `GEOSDistance`.

dr-jts:
Excellent.

It's worth noting that using `IndexedFacetDistance` on the US-Europe test is still 3x faster than `DistanceOp` with the new optimizations.  But this certainly makes the improvement with indexing a lot less significant.

src/operation/distance/DistanceOp.cpp
```diff
@@ -447,7 +447,7 @@ DistanceOp::computeMinDistance(
 
         Envelope segEnv0(p00, p01);
 
-        if (segEnv0.distance(lineEnv1) > minDistance) {
+        if (segEnv0.distanceSquared(lineEnv1) > minDistance*minDistance) {
```
dr-jts: Is this more performant?

dbaston: Modestly but measurably.

PR #318, opened 2020-05-01
Minor IndexedFacetDistance optimizations

dbaston:
- avoid heap allocations, virtual calls
- avoid distance calcs against zero-length segments
- check envelope distance before calculating segment distance

Net gain ~15%.


pramsey:
👍

dbaston:
Committed after rebase

PR #321, opened 2020-05-10
Added shortcut to computeOverlay

doskabouter:
Can add more cases (union,difference) if accepted

### DIFF
--- a/src/operation/overlay/OverlayOp.cpp
+++ b/src/operation/overlay/OverlayOp.cpp
@@ -729,6 +729,23 @@ OverlayOp::computeOverlay(OverlayOp::OpCode opCode)
     const Envelope* env = nullptr;
     const Envelope* env0 = getArgGeometry(0)->getEnvelopeInternal();
     const Envelope* env1 = getArgGeometry(1)->getEnvelopeInternal();
+
+    if (opCode == opINTERSECTION)
+    {
+        const Geometry* geom0 = getArgGeometry(0);
+        const Geometry* geom1 = getArgGeometry(1);
+        if (geom0->isRectangle() && env0->contains(env1))
+        {
+            resultGeom = geom1->clone().release();
+            return;
+        }
+        if (geom1->isRectangle() && env1->contains(env0))
+        {
+            resultGeom = geom0->clone().release();
+            return;
+        }
+    }
+
     Envelope opEnv;
     if(resultPrecisionModel->isFloating()) {
         // Envelope-based optimization only works in floating precision


dr-jts:
Seems reasonable.  Missed this optimization in JTS, but would be good to add it.

doskabouter:
Well, it's coming from postgis. Do you think it's better added overthere?
I also want to add the union, difference and symdifference.
And check if the preparedgeom can benefit too

dr-jts:
Best to add this check in GEOS, and then all clients benefit.  

Adding the other overlay operators would be good.

`PreparedPolygon` has the rectangle check in JTS, so should have that in GEOS too. Or it can be ported.

dr-jts:
Actually this patch violates the contract of overlay.  Overlay promises that the output will be fully noded.  So it's necessary to fully process the inputs always.

Clients which don't care about this have to implement this check themselves.

doskabouter:
bummer

doskabouter:
Is there a better place in geos where this could be added? One which doesn't need to have the result fully noded?

dr-jts:
Well, there's already [`RectangleIntersection`](https://github.com/libgeos/geos/blob/master/src/operation/intersection/RectangleIntersection.cpp).  

Do you have a use case for this over and above that?  Not sure I see much use for optimizing the other overlay methods for rectangles.

doskabouter:
That is indeed another option, but (seeing the time my query took when I added st_intersection) I concluded the USE_RECTANGLE_INTERSECTION is not defined in normal build. I assumed that this was in a define because it wasn't good enough or something)

Use case is doing an intersectionwith with x:-180..180 y -85.5..85.5 on my data to get rid of errors when converting antarctica from wgs to mercator

dr-jts:
Not sure why `RectangleIntersection` isn't wired up by default. Perhaps it doesn't produce valid geometry in some cases?

Can you just call `RectangleIntersection` directly? Or are you using the C API?  Or just do this check in your own code - it's not very much code.

There is going to be a new Overlay codebase showing up soonish, which should address this problem.  So not keen on making big changes before then.

doskabouter:
Nice to hear about the new overlay, hope it's going to have some nice tricks to speedup things!

I can't use any direct functionality from libgeos, I'm using it through postgis.
Wouldn't mind trying to add some stuff to postgis, but since I'm a windows guy and the latest documentation for compiling postgis on windows is 7 years old it's going to be quite a challenge...

dr-jts:
The new overlay should be quite a bit faster for clipping to rectangles.

If you're calling this from PostGIS the usual pattern for speeding up intersection is this:  https://postgis.net/2014/03/14/tip_intersection_faster/

There is also [`ST_CipByBox2D`](http://postgis.net/docs/manual-dev/ST_ClipByBox2D.html), but it is not guaranteed to produce valid output.  It will probably be nice to update that with the new overlay, which should then guarantee valid output.

doskabouter:
I indeed know about that trick about calculating the relation and act accordingly, but I find it's getting messy really fast, especially if that src-geom is already heavily manipulated and the destination geom needs other calculations before consuming the result.

Other option I could do is testing envelope containment in the query itself but in the (or my at least) ideal world this should be abstracted away by the underlying platforms (postgis or libgeos) so I can concentrate on *what* I will get as results instead of trying to figure out *how* to steer postgres/postgis in the fastest direction.

Anyways, I'll close the issue as this solution is indeed not the correct one. 

Thanks for clarifying things, and if possible can you still take a look at [my other pullrequest](https://github.com/libgeos/geos/pull/145) ?

dr-jts:
> I indeed know about that trick about calculating the relation and act accordingly, but I find it's getting messy really fast, especially if that src-geom is already heavily manipulated and the destination geom needs other calculations before consuming the result.

To be clear, that pattern is actually doing more than just optimizing rectangle clipping - it improves performance for clipping to arbitrary polygons as well.  But I agree that it would be nicer to have this built in to PostGIS.

> Other option I could do is testing envelope containment in the query itself but in the (or my at least) ideal world this should be abstracted away by the underlying platforms (postgis or libgeos) so I can concentrate on _what_ I will get as results instead of trying to figure out _how_ to steer postgres/postgis in the fastest direction.

Agreed - it would be better if PostGIS just makes things go fast without a lot of fancy coding from the user.  But it also probably needs to let you be more specific about what you want to do (e.g. in this case clip to a rectangle but don't bother changing geometry which lies within the rectangle).

For background, what does the data you are clipping look like (i.e. geometry type, # vertices)

> Anyways, I'll close the issue as this solution is indeed not the correct one.

Could you open a PostGIS issue requesting performance improvements for clipping to rectangles?  I am thinking that the best way to expose this is via improving the `ST_ClipToRect2D` function. But it would be good to have a place to discuss that.  

> Thanks for clarifying things, and if possible can you still take a look at [my other pullrequest](https://github.com/libgeos/geos/pull/145) ?

Yes, will try and re-look at that soon.

doskabouter:
> 
> For background, what does the data you are clipping look like (i.e. geometry type, # vertices)
> 
Data are (multi) polygons, admin boundaries of the world. # vertices: most of them not too high but there were some nasty outliers (norway,canada). Don't have an exact number but could very well be in the 100000's
> 
> Could you open a PostGIS issue requesting performance improvements for clipping to rectangles? I am thinking that the best way to expose this is via improving the `ST_ClipToRect2D` function. But it would be good to have a place to discuss that.
> 
Will try

> Yes, will try and re-look at that soon.

Thanks!

PR #324, opened 2020-06-19
Add some tests for C API functions

dbaston:
Lost commits discovered when switching to new laptop...

### DIFF
--- a/tests/unit/capi/GEOSBoundaryTest.cpp
+++ b/tests/unit/capi/GEOSBoundaryTest.cpp
@@ -1,0 +1,38 @@
+#include <tut/tut.hpp>
+// geos
+#include <geos_c.h>
+
+#include "capi_test_utils.h"
+
+namespace tut {
+//
+// Test Group
+//
+
+struct test_geosboundary_data : public capitest::test_handlers {};
+
+typedef test_group<test_geosboundary_data> group;
+typedef group::object object;
+
+group test_geosboundary("capi::GEOSBoundary");
+
+template<>
+template<>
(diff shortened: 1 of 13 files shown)


tests/unit/capi/GEOSBoundaryTest.cpp
```diff
@@ -0,0 +1,38 @@
...
+typedef test_group<test_geosboundary_data> group;
+typedef group::object object;
+
+group test_geosboundary("capi::GEOSBoundary");
+
+template<>
+template<>
+void object::test<1>()
+{
+    GEOSGeometry* input = GEOSGeomFromWKT("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))");
```
Algunenano, line 23: Not directly related to this, but have you considered using some kind of shared pointer to handle the destruction automatically so there is no need to call `GEOSGeom_destroy` at the end?

In other projects I've been using something as described in https://stackoverflow.com/a/26276805/13038612: 
```
template <typename T> using deleted_unique_ptr = std::unique_ptr<T, std::function<void(T *)>>;
```

And then using it like:
```
deleted_unique_ptr<Type> name(constructor, destructor_function)
```

So destructor_function(name) is called when it goes out of scope.

dbaston: Yes, that's what I do in my own projects, but for these C API tests I'd prefer to keep it simple (even if that means verbose)

PR #326, opened 2020-06-24
CMake: Remove NAMELINK_SKIP for C++ lib install

mwtoews:
This reverses part of @pramsey's  a71e095d304d13c071e5144f128fe37680e6c47b which (on some platforms) creates a symblink to the shared library:

    libgeos.so -> libgeos.so.3.9.0

With NAMELINK_SKIP, the libgeos.so symblink is not created. This means it's impossible to link to this library using `-lgeos` (which `geos-config --cclibs` does).

This PR allows the symbolic link to be created, similar to what the autotools version does (except that GEOS' Autotools SONAME version filename format is different, libgeos.so -> libgeos-3.9.0.so)

And yes, I'm aware that dynamic linking to the C++ API is discouraged.


pramsey:
Just to confirm, you're sure this behaviour is what the autotools does? Because I fiddled with cmake for quite a while to try and ape the autotools behaviour, which is not exactly standard (because of the "discouragement").

mwtoews:
Sure, here is what happens with the current master with an autotools build and install:
```
$ ./configure --prefix=/tmp/geos-autotools && make -j && make install
...
$ ls -gnG /tmp/geos-autotools/lib/
total 137292
-rwxr-xr-x 1 34125248 Jun 26 12:18 libgeos-3.9.0.so
-rw-r--r-- 1 98816712 Jun 26 12:18 libgeos.a
-rw-r--r-- 1  5003850 Jun 26 12:18 libgeos_c.a
-rwxr-xr-x 1      992 Jun 26 12:18 libgeos_c.la
lrwxrwxrwx 1       19 Jun 26 12:18 libgeos_c.so -> libgeos_c.so.1.14.0
lrwxrwxrwx 1       19 Jun 26 12:18 libgeos_c.so.1 -> libgeos_c.so.1.14.0
-rwxr-xr-x 1  2617736 Jun 26 12:18 libgeos_c.so.1.14.0
-rwxr-xr-x 1      948 Jun 26 12:18 libgeos.la
lrwxrwxrwx 1       16 Jun 26 12:18 libgeos.so -> libgeos-3.9.0.so
```
and with a CMake build and install before this PR:
```
$ mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX=/tmp/geos-cmake .. && make -j && make install
...
$ ls -gnG /tmp/geos-cmake/lib/
total 8612
drwxr-xr-x 3    4096 Jun 26 12:20 cmake
lrwxrwxrwx 1      14 Jun 26 12:20 libgeos_c.so -> libgeos_c.so.1
lrwxrwxrwx 1      19 Jun 26 12:20 libgeos_c.so.1 -> libgeos_c.so.1.14.0
-rw-r--r-- 1  862760 Jun 26 12:20 libgeos_c.so.1.14.0
-rw-r--r-- 1 7949056 Jun 26 12:20 libgeos.so.3.9.0
```
after this PR, `libgeos.so -> libgeos.so.3.9.0` would be also created, similar to autotool's `libgeos.so -> libgeos-3.9.0.so` above.

mwtoews:
Oh, and I should also say that the autotools build can link to the C++ API using either:

 - `geos-config --libs` returns `-L/usr/local/lib -lgeos-3`, which links to `libgeos-3.9.0.so`
 - `geos-config --cclibs` returns `-L/usr/local/lib -lgeos`, which links to `libgeos.so`

meanwhile the current cmake build cannot establish dynamic links directly with `libgeos.so.3.9.0`, so while attempting to link an application with `-L/tmp/geos-cmake/lib -lgeos`, the end result is that it will link to Ubuntu's system `libgeos-3.6.2.so => /usr/lib/x86_64-linux-gnu/libgeos-3.6.2.so`, which is a surprise.

I have some upcoming post-install tests nearly ready for a PR that checks this kind of behaviour.

mwtoews:
@pramsey any more thoughts on this change to CMake? I'm looking to get #346 to run post-install tests, where this issue gets caught up, e.g. https://travis-ci.com/github/libgeos/geos/jobs/439098095#L4276 where the C app compiles and links OK, but the C++ app compiles, but links to a system GEOS library as a bit of a surprise.

pramsey:
Seems fine to me, ready to merge?

PR #328, opened 2020-06-27
Add tools/cppcheck.sh and make it a AZP job

rouault:
And fixes related warnings

Adapted from script used by GDAL, PROJ, QGIS

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -258,3 +258,14 @@ stages:
         ctest -V --output-on-failure -C $BUILD_TYPE
       displayName: 'Build C++20'
       condition: contains(variables['CXXSTD'], '20')
+
+  - job: Code_quality_checks
+    pool:
+      vmImage: 'ubuntu-18.04'
+    steps:
+    - script: |
+        set -e
+        sudo -E apt-get update
+        sudo -E apt-get -yq --no-install-suggests --no-install-recommends install cppcheck
+        ./tools/cppcheck.sh
+      displayName: 'cppcheck'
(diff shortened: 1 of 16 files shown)


rouault:
Any interest from the core team for that contribution, before I rebase it to latest master ?

pramsey:
Seems OK to me? Wonder if we can thin out the AZP matrix so it runs in less that fifty billion hours...?

rouault:
> Wonder if we can thin out the AZP matrix so it runs in less that fifty billion hours...?

a bit of a different topic, but for the Linux builds, you could potentially just test the lower and higher gcc & clang versions, and hope that in-between ones are OK...

rouault:
Rebased on top of latest master

rouault:
Note: the CI failures are unrelated to my changes. The same ones occur with master

Algunenano:
> Any interest from the core team for that contribution, before I rebase it to latest master ?

Not a core contributor, but I think having this kind of tooling integrated in the CI it's great, in many situations it forces the code to be simpler (and more modern / C++11-esque) and can detect bugs that otherwise are hard to see.

The one thing I'm not convinced about, but it's totally personal preference, is the `postfixOperator`. I prefer i++ vs ++i, but I'm happy either way if something detects it for me (although ideally it'd be great to have this integrated as part of a the linter / clang format).

rouault:
> is the `postfixOperator`

the rationale given by this one is that the post-fix operator involves more work in its implementation than the pre-fix one. For the post-fix one, you need to create a copy of an object (for incrementation not on base types). So there's a tiny performance gain in using the pre-fix one.

pramsey:
I'm going to take it in, let the CI zoo continue to grow :)

rouault:
@pramsey I'm confused: this pull request is supposed to have been merged, but I can't see any trace of it in master (here on github I mean). Can't find cppcheck.sh in https://github.com/libgeos/geos/tree/master/tools ...

pramsey:
I think I mashed a merge button maybe, instead of doing the manual two-step to push to the osgeo repo. Sorry about that, it's done now.

rouault:
> Sorry about that

thanks! (To be clear my rant wasn't about an occasional mess up, but against the process that makes it possible)

PR #332, opened 2020-08-03
Fix windows compilation issues.

abellgithub:
So, this is confusing, but here's the explanation:

IF you create a class that you're dllexport'ing, the compiler tries to instantiate all the implicit functions that the class needs at the time that it sees the class. This seems to happen even if you never use those functions or instantiate the class. This is to allow the version in the DLL to be used by the dynamic linker.

Anyway, you can't copy or assign a class with vector<unique_ptr<foo>>, so the compiler complains when it attempts to instantiate the implicit functions that are part of the class.  These changes stop the compiler from trying to create the implicit assign operator and copy ctor.

### DIFF
--- a/include/geos/operation/overlayng/IntersectionPointBuilder.h
+++ b/include/geos/operation/overlayng/IntersectionPointBuilder.h
@@ -87,8 +87,8 @@ public:
 
     std::vector<std::unique_ptr<geom::Point>> getPoints();
 
-
-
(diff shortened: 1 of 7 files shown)


pramsey:
This explains the same pattern that shows up in other parts of the code base, a piece of weird knowledge we are repeatedly rediscovering, apparently.

PR #334, opened 2020-09-02
Fix segfault in SimplePointInAreaLocator caused by casting MultiPolyg…

sumeerbhola:
…on to Polygon

References https://trac.osgeo.org/geos/ticket/1047
References https://github.com/cockroachdb/cockroach/issues/53254

---
This change is [Reviewable](https://reviewable.io/reviews/libgeos/geos/334)

### DIFF
--- a/src/algorithm/locate/SimplePointInAreaLocator.cpp
+++ b/src/algorithm/locate/SimplePointInAreaLocator.cpp
@@ -67,15 +67,17 @@ SimplePointInAreaLocator::locateInGeometry(const Coordinate& p, const Geometry* 
     }
 
     if (geom->getNumGeometries() == 1) {
-        return locatePointInPolygon(p, dynamic_cast<const Polygon*>(geom->getGeometryN(0)));
-    } else {
-        for (size_t i = 0; i < geom->getNumGeometries(); i++) {
-            const Geometry* gi = geom->getGeometryN(i);
-
-            auto loc = locateInGeometry(p, gi);
-            if(loc != Location::EXTERIOR) {
-                return loc;
-            }
+        auto poly = dynamic_cast<const Polygon*>(geom->getGeometryN(0));
+        if (poly) {
+            return locatePointInPolygon(p, poly);
+        }
+        // Else it is a collection with a single element. Will be handled below.
+    }
+    for (size_t i = 0; i < geom->getNumGeometries(); i++) {
+        const Geometry* gi = geom->getGeometryN(i);
+        auto loc = locateInGeometry(p, gi);
+        if(loc != Location::EXTERIOR) {
+            return loc;
         }
     }
 
--- a/tests/unit/algorithm/PointLocatorTest.cpp
+++ b/tests/unit/algorithm/PointLocatorTest.cpp
@@ -5,6 +5,7 @@
 #include <tut/tut.hpp>
 // geos
 #include <geos/io/WKTReader.h>
+#include <geos/algorithm/locate/SimplePointInAreaLocator.h>
 #include <geos/algorithm/PointLocator.h>
 #include <geos/geom/PrecisionModel.h>
 #include <geos/geom/GeometryFactory.h>
@@ -56,6 +57,14 @@ runPtLocator(Location expected, const Coordinate& pt,
     ensure_equals(loc, expected);
 }
 
+void
+runSimplePtLocator(Location expected, const Coordinate& pt,
+             const std::string& wkt)
+{
+    GeomPtr geom(reader.read(wkt));
+    Location loc = geos::algorithm::locate::SimplePointInAreaLocator::locate(pt, geom.get());
+    ensure_equals(loc, expected);
+}
 
 //
 // Test Cases
@@ -109,6 +118,16 @@ void object::test<5>
 {
     runPtLocator(Location::INTERIOR, Coordinate(0, 0),
                  "MULTIPOINT ((1 1), (0 0))");
+}
+
+// 6 - TestPointLocator Point inside GeometryCollection containing MultiPolygon, using SimplePointInAreaLocator.
+template<>
+template<>
+void object::test<6>
+()
+{
+    runSimplePtLocator(Location::INTERIOR, Coordinate(0, 0),
+                       "GEOMETRYCOLLECTION (MULTIPOLYGON (((-1 -1, 1 -1, 1 1, -1 1, -1 -1))))");
 }
 
 } // namespace tut


src/algorithm/locate/SimplePointInAreaLocator.cpp
```diff
@@ -67,15 +67,17 @@ SimplePointInAreaLocator::locateInGeometry(const Coordinate& p, const Geometry*
...
-        for (size_t i = 0; i < geom->getNumGeometries(); i++) {
-            const Geometry* gi = geom->getGeometryN(i);
-
-            auto loc = locateInGeometry(p, gi);
-            if(loc != Location::EXTERIOR) {
-                return loc;
-            }
+        auto poly = dynamic_cast<const Polygon*>(geom->getGeometryN(0));
+        if (poly) {
+            return locatePointInPolygon(p, dynamic_cast<const Polygon*>(geom->getGeometryN(0)));
```
otan, line 72: maybe use `poly` as the second argument?

tests/unit/algorithm/PointLocatorTest.cpp
```diff
@@ -111,5 +120,15 @@ void object::test<5>
                  "MULTIPOINT ((1 1), (0 0))");
 }
 
+// 6 - TestPointLocator Point inside GeometryCollection containing MultiPolygon, using SimplePointInAreaLocator.
```
otan, line 123: maybe worth adding a clarification it is a single Polygon inside a MULTIPOLYGON

dr-jts:
In JTS at least, spatial predicates do not support GEOMETRYCOLLECTIONs as inputs. So this is a case of erroneous input. But should not be allowed to cause a segfault, for sure.

dr-jts:
The idea of generalizing `SimplePointInAreaLocator` to handle any kind of collection isn't a bad one. But the fact that this allows `intersects` to handle a geometry like
```
GEOMETRYCOLLECTION (MULTIPOLYGON (((9223372036854776000 -9223372036854776000, 9223372036854776000 9223372036854776000, -9223372036854776000 9223372036854776000, -9223372036854776000 9223372036854776000, 9223372036854776000 -9223372036854776000))))
```
(as per the original ticket [GEOS-1047](https://trac.osgeo.org/geos/ticket/1047)) is just a lucky accident.  You should review your upstream code and ensure it meets the contract for GEOS spatial predicates.

otan:
> You should review your upstream code and ensure it meets the contract for GEOS spatial predicates.

This is for a database (i.e. CockroachDB), where the users can specify the kinds of shapes coming in. We're just the messenger. This affects PostGIS as well (https://trac.osgeo.org/geos/ticket/1042).

We don't control the shapes; this was one we happened to bump into whilst randomized testing a variety of different unions that led to this shape. By chance, it also affects intersects.

However, the seg fault as it stands is definitely not a good situation for us as it is not easily recoverable.

dr-jts:
I also note that the original [CockroachDB issue](https://github.com/cockroachdb/cockroach/issues/53254) involved computing an aggregate union of two geometry collections (one heterogeneous).  This seems to be supported in PostGIS, but   not sure how GEOS is being used to perform that.

dr-jts:
> > You should review your upstream code and ensure it meets the contract for GEOS spatial predicates.
> 
> This is for a database (i.e. CockroachDB), where the users can specify the kinds of shapes coming in. We're just the messenger. This affects PostGIS as well (https://trac.osgeo.org/geos/ticket/1042).
> 
> We don't control the shapes; this was one we happened to bump into whilst randomized testing a variety of different unions that led to this shape. By chance, it also affects intersects.
> 
> However, the seg fault as it stands is definitely not a good situation for us as it is not easily recoverable.

If you're using a library, it's your responsibility to ensure that you are not providing inputs to the library which are out of contract for it.

As said, I agree that the lib should not segfault on bad inputs.  But the answer is not necessarily to hack a "fix" deep inside the codebase with no knowledge of how that might affect other code in the library.

otan:
> As said, I agree that the lib should not segfault on bad inputs. But the answer is not necessarily to hack a "fix" deep inside the codebase with no knowledge of how that might affect other code in the library. The library should detect known situations which are not handled and report that via standard error mechanism.

It may be more helpful if you gave a pointer as to where / what you prefer changed in this situation.

It sounds like if you want GEOMETRYCOLLECTION spatial predicates to not work entirely, it will break existing cases. This already works with MULTIPOLYGON inside GEOMETRYCOLLECTION if the MULTIPOLYGON has >= 2 shapes within it. Should that be changed going forwards? Sounds like that would break existing PostGIS cases as well?

dr-jts:
> It may be more helpful if you gave a pointer as to where / what layer you prefer changed in this situation.

I'm not sure at the moment. In JTS GeometryCollection input to predicates are simply disallowed.  I'm not sure why this isn't implemented in GEOS.

> It sounds like if you want GEOMETRYCOLLECTION spatial predicates to not work entirely, it will break existing cases. This seems like it already works with MULTIPOLYGON inside GEOMETRYCOLLECTION if the MULTIPOLYGON has >= 2 shapes within it. Sounds like that would break existing PostGIS cases as well?

Are there existing PostGIS unit tests that rely on this?

If a GC with a single MultiPolygon works, that is just a happy accident. But I suppose it could be made explicit behaviour.  I'd prefer that was done more explicitly, however, not just as a lucky side-effect of a sub-module of the overlay algorithm.  (I.e. explicitly code for unwrapping GCs containing a single element which is itself not a GC).

dr-jts:
Whatever the way this is handled, there needs to be a unit test covering this case.  Otherwise this risks a regression failure if and when the predicate code is modified or replaced.

src/algorithm/locate/SimplePointInAreaLocator.cpp
sumeerbhola: Done

sumeerbhola:
> there needs to be a unit test covering this case

This PR includes a change to `PointLocatorTest.cpp`. Should there be more tests?

dr-jts:
> > there needs to be a unit test covering this case
> 
> This PR includes a change to `PointLocatorTest.cpp`. Should there be more tests?

Yes, but perhaps they should not be part of this PR.  If the ultimate goal is to have GEOS support GeometryCollection inputs to predicates, then there needs to be tests at that level.

This will need to be based on a decision by the GEOS team about how to handle this.  Or not - currently GC inputs sometimes work, sometimes probably return the wrong answer, and sometimes cause topology exceptions.  To my mind this means they basically aren't handled in a meaningful way.

strk:
Thanks all, this was merged in master as 07e60468a9a155c63b06fa1e24db7c113f03004a  - backports are in progress

PR #335, opened 2020-09-06
io: fix parsing of WKT with EMPTY components

otan:
* Allow WKT to parse EMPTY within MULTIPOINT elements.
* Fix a bug where getNextToken() would not skip forward empty spaces.

Refs: https://github.com/cockroachdb/cockroach/issues/53997


pramsey:
Thanks!

PR #336, opened 2020-09-16
Compiler warning fixes

rouault:
A few unexciting compiler warnings found when building with cmake & gcc 5.5


pramsey:
Lovely, thank you Even!

PR #340, opened 2020-10-15
DOC: clarify that GEOSProject returns -1 on exception (C API)

jorisvandenbossche:
With PyGEOS we forgot to check this, so I think would be good to add this to the doc comment.


strk:
Thanks, this was merged as https://github.com/libgeos/geos/commit/4f099194d449b987cf9349f65731fce89891152f

jorisvandenbossche:
Thanks for the quick merge!

PR #343, opened 2020-11-04
Remove no-op code from QESubdiv::remove

dr-jts:
As discussed in #342, the use of `erase` to remove detached `QuadEdges` is incorrect, since it causes `QuadEdge` addresses to change.  Instead, it is fine to simply leave them in the `quadEdges` container marked as non-live.


dr-jts:
@strk feel free to merge this.

strk:
Merged as 76d7586ea53b2f9796bd4257d1d997fe05f11703

PR #355, opened 2020-11-29
WIP: New STRtree implementation

dbaston:
This PR contains a new STRtree implementation (placeholder name!) that continues work started by @pramsey in the `SimpleSTRtree`. Like that tree, it disentangles itself from the `AbstractSTRtree` which allows more implementation flexibility and less virtual call overhead. It goes a bit further than the `SimpleSTRtree` in reducing memory fragmentation and usage, by precomputing the final size of the tree (all leaf and branch nodes) and setting aside that space in a contiguous `std::vector`. This allows us to build the tree such that the children of a branch node are always contiguous in memory, which brings about faster access but also reduced memory requirements for the branch node itself (we only need to store pointers to the first and last child; we don't need to allocate and store a vector of pointers to all children.) If we know the number of items to include in the tree in advance, the entire tree construction uses only a single heap allocation.

The construction algorithm was the same as that used in `STRtree` and `SimpleSTRtree`, so the generated trees should have been equivalent to the ones produced by those classes. However, I relaxed the requirement that the child nodes of a given branch occur in sorted order. (This allows us to avoid using `std::sort` and only do a partial sort using `std::nth_element`.) This could be problematic, since it could mean that equally-valid outputs could differ between platforms (polygon component order, etc.).

It can be used as a drop-in replacement wherever a `SpatialIndex` is required. I have not (yet?) implemented the nearest-neighbor pieces that are outside the `SpatialIndex` interface.

In benchmarking, it reduces the time of a polygon union operation by about 4-5% (relative to `SimpleSTRtree`) which makes this about 35 - 45% faster than `SimpleSTRtree` in that application.

It is templated to store any `ItemType*` but maybe this could be relaxed to store value types as well.


pramsey:
This is a masterpiece of concision.

pramsey:
Without the nearest-neighbour bits, it's less useful, as you don't get to plug it into indexed facet distance, etc. There's also a piece in the AbstractSTRtree that spits out the items in index order which is used by the cascaded union code, but I feel like it would be nice to rip that out of cascaded union in favour of a hilbert sort of the inputs prior to the binary union recursion.

pramsey:
I remain very excited by this.

dbaston:
Just hacking away a bit more so that you can also declare `TemplateSTRtree<std::unique_ptr<Geometry>> t`.

dbaston:
Performance for `GEOSPreparedContains` is about 35-40% better on this branch relative to master, if the tree is queried many times:

```
➜  cmake-build-release git:(master) ✗ bin/perf_geospreparedcontains ~/data/australia.txt 1000000
Performing 1000000 point-in-polygon tests.
Reading shape from /home/dan/data/australia.txt
GEOSPreparedContains: 328315 hits from 1000000 points in 1,829,257 usec
```

```
➜  cmake-build-release git:(baston-tree) ✗ bin/perf_geospreparedcontains ~/data/australia.txt 1000000
Performing 1000000 point-in-polygon tests.
Reading shape from /home/dan/data/australia.txt
GEOSPreparedContains: 328315 hits from 1000000 points in 1,138,055 usec
```

But worse if it is queried few times:

```
➜  cmake-build-release git:(master) ✗ bin/perf_geospreparedcontains ~/data/australia.txt 100
Performing 100 point-in-polygon tests.
Reading shape from /home/dan/data/australia.txt
GEOSPreparedContains: 34 hits from 100 points in 148,534 usec

```
```
➜  cmake-build-release git:(baston-tree) ✗ bin/perf_geospreparedcontains ~/data/australia.txt 100
Performing 100 point-in-polygon tests.
Reading shape from /home/dan/data/australia.txt
GEOSPreparedContains: 34 hits from 100 points in 192,491 usec
```

pramsey:
That seems counter-intuitive... does the timing include the tree build? I assume not?

dbaston:
> That seems counter-intuitive... does the timing include the tree build? I assume not?

It is included, because the "preparing" in a prepared geometry happens on the first query.

`master` here is actually not using an `STRtree`, but a hand-coded 1D equivalent of it where each node has exactly two children. So it's not too surprising that the `STRtree` build process, which is clearly written with the 2D case in mind, would do worse. Maybe I can tighten the gap, though.

dbaston:
Performance is now 12% faster than master for the 100x query and 45% faster for the 1,000,000x query.

PR #356, opened 2020-11-29
Add Envelope::fromXY

dbaston:
Per discussion in https://lists.osgeo.org/pipermail/geos-devel/2020-November/009882.html

### DIFF
--- a/include/geos/geom/Envelope.h
+++ b/include/geos/geom/Envelope.h
@@ -117,6 +117,8 @@ public:
     static bool intersects(const Coordinate& p1, const Coordinate& p2,
                            const Coordinate& q);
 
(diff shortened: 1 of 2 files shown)


dr-jts:
Ok yes, this is one way to add a different parameter order.  But you have to know this method exists and that you need to use it - which seems to have the same cognitive cost as simply using the current parameter order in the constructor!  I guess it does help to highlight the unconventional order of the constructor params though.

dbaston:
Yeah, not sure changing the parameter order is worth it. I do think having a constructor that assumes sorted inputs `(xmin, xmax)` vs `(x1, x2)` is worthwhile.

PR #357, opened 2020-11-29
Remove default arguments for Coordinate constructor

dbaston:
The previously defined constructor allowed a Coordinate to be constructed
from only a single value, i.e. `Coordinate(3)` would produce `(3, 0, NaN)`.
It's not clear what the use for this is, and it allowed further
complications like `Envelope(3)` compiling by implicitly constructing the
coordinate above.

This commit adds a no-argument constructor to preserve the behavior of
`Coordinate()` producing `(0, 0, NaN)`.

### DIFF
--- a/include/geos/geom/Coordinate.h
+++ b/include/geos/geom/Coordinate.h
@@ -91,7 +91,9 @@ public:
 
     bool isNull() const;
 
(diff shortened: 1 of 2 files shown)


dr-jts:
Good fix.

PR #358, opened 2020-11-30
Validate complex UnaryUnion test with OverlayNG

mwtoews:
Closes [#1046](https://trac.osgeo.org/geos/ticket/1046)


pramsey:
Thanks!

dr-jts:
This test was actually already present:  https://github.com/libgeos/geos/blob/master/tests/xmltester/tests/robust/overlay/TestOverlay-shapely-829.xml

Would be nice to avoid duplication?

pramsey:
Is it? The coordinates look quite different.

dr-jts:
Yes, you're right.  My bad.

dr-jts:
Yes, I suggest removing the file added in this PR.  

I think it's nicer to start pruning the random assortment of tests in `issues` if they exist elsewhere.   There's probably more dups to `robust/overlay` tests - should really clean them up.

pramsey:
Contra that I kind of like the issues files because they map back to ticket numbers. Can see if an old ticket has a test easily. And vice versa, can research "what was the problem this file was about".

pramsey:
Removed dupe file at 401aee79

dr-jts:
> Contra that I kind of like the issues files because they map back to ticket numbers. Can see if an old ticket has a test easily. And vice versa, can research "what was the problem this file was about".

That's a good point.  

It would be nice if the issue filenames indicated what operation or issue they are testing.

PR #359, opened 2020-11-30
Use Polygon constructor that avoids internal delete 

pramsey:
The constructor that purely uses unique_ptr inputs seems to both avoid bulk copy and also the internal delete of the vector containing the rings. Failing that, a whole extra Polygon constructor will be needed.

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -4411,16 +4411,21 @@ extern "C" {
         try {
             using geos::geom::LinearRing;
 
-            auto vholes = geos::detail::make_unique<std::vector<LinearRing*>>(nholes);
-
+            /*
+            * First check that we can cast these generic
+            * geometry pointers into linearrings. Then
+            * convert them to unique_ptr to pass to the constructor
+            * so we can get a no-copy transfer of the data
+            */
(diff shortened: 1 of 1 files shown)


capi/geos_ts_c.cpp
```diff
@@ -4411,24 +4411,36 @@ extern "C" {
...
             LinearRing* nshell = dynamic_cast<LinearRing*>(shell);
             if(! nshell) {
                 handle->ERROR_MESSAGE("Shell is not a LinearRing");
                 return NULL;
             }
             const GeometryFactory* gf = handle->geomFactory;
 
-            return gf->createPolygon(nshell, vholes.release());
+            /* Create unique_ptr version for constructor */
+            std::vector<std::unique_ptr<LinearRing>> vholes(nholes);
```
rouault, line 4437: Unless I'm wrong the current code will have vholes with nholes first elements at nullptr, followed by the content of tmpholes.
You'd probably want to do instead
```suggestion
            std::vector<std::unique_ptr<LinearRing>> vholes;
            vholes.reserve(nholes);
```

pramsey: yep, you are right

PR #363, opened 2020-12-09
Store SegmentNode values directly in SegmentNodeList

dbaston:
SegmentNodeList went to great pains to create SegmentNodes with stable
addresses so that pointers to the created SegmentNodes could be passed
to callers. However, the returned pointers were never used. This commit
simplifies SegmentNodeList by storing SegmentNodes directly in nodeMap.

Improves overlay performance by maybe 5%.

### DIFF
--- a/include/geos/noding/Makefile.am
+++ b/include/geos/noding/Makefile.am
@@ -1,5 +1,5 @@
 #
-# This file is part of project GEOS (http://trac.osgeo.org/geos/) 
+# This file is part of project GEOS (http://trac.osgeo.org/geos/)
 #
 SUBDIRS = \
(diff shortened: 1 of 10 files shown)


pramsey:
Sounds good. Looking forward to getting 3.9 out so we can land all these bits and pieces.

dr-jts:
I have been working on a fast Line clipping algorithm that turned out to need access to the SegmentNodes.  So never say never...  This is fine to go ahead now, but *might* need revisiting sometime.

dbaston:
Added some polishing; net gain in overlay test (union of BC watersheds) is 12%.

Issue #353, opened 2020-11-28
Build error in ARM64 Mac system


sigmafelix:
Hello,
I am trying to build `geos` in my ARM64 Mac system, but I am getting an error when the `make` process gets to make a dynamic library:

```
% sudo /usr/local/bin/cmake .
-- GEOS: Using default build type: Release
-- GEOS: Run-time output: /Users/isong/Downloads/geos/bin
-- GEOS: Archives output: /Users/isong/Downloads/geos/lib
-- GEOS: Version 3.9.0dev
-- GEOS: C API Version 1.14.0
-- GEOS: JTS port 1.16.0
-- GEOS: Require C++11
-- GEOS: Developer mode enabled
-- GEOS: Configured 'dist' target
-- GEOS: Configured 'distcheck' target
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/isong/Downloads/geos
isong@Insangs-MacBook-Pro geos % sudo /usr/local/bin/make . 
isong@Insangs-MacBook-Pro geos % sudo /usr/local/bin/make  
[  0%] Linking CXX shared library lib/libgeos.dylib
duplicate symbol 'vtable for geos::noding::BasicSegmentString' in:
    CMakeFiles/geos.dir/src/inlines.cpp.o
    CMakeFiles/geos.dir/src/noding/BasicSegmentString.cpp.o
duplicate symbol 'typeinfo name for geos::noding::BasicSegmentString' in:
    CMakeFiles/geos.dir/src/inlines.cpp.o
    CMakeFiles/geos.dir/src/noding/BasicSegmentString.cpp.o
duplicate symbol 'typeinfo for geos::noding::BasicSegmentString' in:
    CMakeFiles/geos.dir/src/inlines.cpp.o
    CMakeFiles/geos.dir/src/noding/BasicSegmentString.cpp.o
ld: 3 duplicate symbols for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [CMakeFiles/geos.dir/build.make:5383: lib/libgeos.3.9.0.dylib] Error 1
make[1]: *** [CMakeFiles/Makefile2:817: CMakeFiles/geos.dir/all] Error 2
make: *** [Makefile:182: all] Error 2
```

Does anyone have an idea for the error?
Thank you.

pramsey:
[...]

PR #365, opened 2020-12-15
Disable additional inlining definitions when building for Apple ARM64

tzakharko:
Fixes https://github.com/libgeos/geos/issues/353

The "fix" is simply to ignore the includes on the affected platforms, as it already done on Windows

### DIFF
--- a/src/inlines.cpp
+++ b/src/inlines.cpp
@@ -35,6 +35,9 @@
 // the stubs correctly at link time by itself
 #if !defined(__CYGWIN__)
 
+// Same as above for Apple ARM64 platforms
+#if !(defined(__APPLE__) && defined(__aarch64__))
+
 // Undefine GEOS_INLINE so that .inl files
 // will be ready for an implementation file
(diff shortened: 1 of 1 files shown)


src/inlines.cpp
```diff
@@ -35,6 +35,9 @@
 // the stubs correctly at link time by itself
 #if !defined(__CYGWIN__)
 
+// Same as above for Apple ARM64 platforms
+#if !defined(__APPLE__) && !defined(__aarch64__)
```
macdrevx, line 39: Might it be better to say:

```
#if !(defined(__APPLE__) && defined(__aarch64__))
```

so that the change only applies to ARM64 Macs?

tzakharko: Thanks for pointing this out! It should have been a `||` instead of `&&`, that's what I get doing things in a rush and not paying attention. 

I have now changed the line as you suggest as it's definitely more readable.

pramsey:
Thanks!

PR #369, opened 2020-12-19
Add inlines relevant to overlay performance

dbaston:


dr-jts:
Good job on inlining. Especially OverlayLabel - a perfect use case for it.

PR #370, opened 2020-12-19
Avoid two CoordinateSequence copies in OverlayNG

dbaston:


src/operation/overlayng/EdgeNodingBuilder.cpp
```diff
@@ -128,8 +129,8 @@ EdgeNodingBuilder::createEdges(std::vector<SegmentString*>* segStrings)
         // Record that a non-collapsed edge exists for the parent geometry
         hasEdges[info->getIndex()] = true;
         // Allocate the new Edge locally in a std::deque
-        std::unique_ptr<CoordinateSequence> ssPts = ss->getCoordinates()->clone();
-        edgeQue.emplace_back(ssPts.release(), info);
+        NodedSegmentString* nss = detail::down_cast<NodedSegmentString*>(ss);
+        edgeQue.emplace_back(nss->releaseCoordinates(), info);
```
pramsey, line 133: This noded segment string will get deleted in node() after being used in create edges. I assume that the release() in releaseCoordinates() nulls out the pointer so we don't end up with a double free in the NodedSegmentString destructor?

src/operation/overlayng/OverlayEdgeRing.cpp
```diff
@@ -41,8 +41,9 @@ OverlayEdgeRing::OverlayEdgeRing(OverlayEdge* start, const GeometryFactory* geom
     , locator(nullptr)
     , shell(nullptr)
 {
-    computeRingPts(start, ringPts);
-    computeRing(ringPts, geometryFactory);
+    auto ringPts = detail::make_unique<CoordinateArraySequence>();
```
pramsey, line 44: Clever, looks good to me.

pramsey (approved):

src/operation/overlayng/EdgeNodingBuilder.cpp
dbaston: Good question, and I think the answer is "yes" https://en.cppreference.com/w/cpp/memory/unique_ptr/release

PR #373, opened 2020-12-22
Some MonotoneChain improvements

dbaston:
This PR makes `Quadrant` into an `enum class`, allowing us to specify a smaller type (`char`) and carry it around for for free on `MonotoneChain` objects (because of struct packing.) It also has a type-safety benefit. The stored `Quadrant` is taken advantage of to speed up envelope initialization for a chain, and could possibly be used later in the overlap checking. The PR also avoids individual heap-allocations of the chains in `MonotoneChainBuilder`.

Net benefit is minor, about 4% of a union operation. So maybe not worth writing, but unfortunately by the time you find that out it's already written. On the other hand, the 3-4% changes can and do add up.

### DIFF
--- a/benchmarks/ClassSizes.cpp
+++ b/benchmarks/ClassSizes.cpp
@@ -38,6 +38,7 @@
 #include <geos/noding/NodedSegmentString.h>
 #include <geos/profiler.h>
 #include <geos/constants.h>
+#include <geos/index/chain/MonotoneChain.h>
 #include <iostream>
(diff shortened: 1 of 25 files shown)


dr-jts:
Nice!  Agreed, all these minor changes will definitely add up.  And things like this are the preferred pattern in C++, so it's nice to use the full capabilities of the language.

dbaston:
Closing in favor of #374

PR #374, opened 2020-12-22
MonotoneChain handling performance

pramsey:
Overlays start with an MCIndexedNoder, so making that faster should affect all overlays quite a lot. 

Current behaviour heap allocates a lot of wee MonotoneChain objects, and takes copies of vectors of pointers to those objects. With a little change in workflow, we can place the MonotoneChain objects directly into a vector that is stack-allocated in the object that actually uses the MonotoneChains (either the MCIndexNoder or MCIndexSegmentSetMutualIntersector).

For the Vancouver Island watershed union, this change provides a 9% speed-up.

### DIFF
--- a/include/geos/noding/MCIndexSegmentSetMutualIntersector.h
+++ b/include/geos/noding/MCIndexSegmentSetMutualIntersector.h
@@ -22,13 +22,14 @@
 
 #include <geos/noding/SegmentSetMutualIntersector.h> // inherited
 #include <geos/index/chain/MonotoneChainOverlapAction.h> // inherited
+#include <geos/index/chain/MonotoneChain.h> // inherited
+#include <geos/index/strtree/SimpleSTRtree.h> // inherited
 
 namespace geos {
 namespace index {
 class SpatialIndex;
 
 namespace chain {
-class MonotoneChain;
 }
 namespace strtree {
 //class STRtree;
@@ -55,14 +56,22 @@ namespace noding { // geos::noding
 class MCIndexSegmentSetMutualIntersector : public SegmentSetMutualIntersector {
 public:
 
-    MCIndexSegmentSetMutualIntersector();
+    MCIndexSegmentSetMutualIntersector()
+        : monoChains()
+        , index(new index::strtree::SimpleSTRtree())
+        , indexCounter(0)
+        , processCounter(0)
+        , nOverlaps(0)
+        , indexBuilt(false)
+    {}
 
-    ~MCIndexSegmentSetMutualIntersector() override;
+    ~MCIndexSegmentSetMutualIntersector() override
+    {};
 
     index::SpatialIndex*
     getIndex()
     {
-        return index;
+        return index.get();
     }
 
     void setBaseSegments(SegmentString::ConstVect* segStrings) override;
@@ -96,7 +105,7 @@ public:
 
 private:
 
-    typedef std::vector<std::unique_ptr<index::chain::MonotoneChain>> MonoChains;
+    typedef std::vector<index::chain::MonotoneChain> MonoChains;
     MonoChains monoChains;
 
     /*
@@ -104,7 +113,7 @@ private:
      * envelope (range) queries efficiently (such as a index::quadtree::Quadtree
      * or index::strtree::STRtree).
      */
-    index::SpatialIndex* index;
+    std::unique_ptr<index::SpatialIndex> index;
     int indexCounter;
     int processCounter;
     // statistics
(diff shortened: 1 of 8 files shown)


src/noding/MCIndexSegmentSetMutualIntersector.cpp
```diff
@@ -124,14 +100,22 @@ MCIndexSegmentSetMutualIntersector::setBaseSegments(SegmentString::ConstVect* se
 void
 MCIndexSegmentSetMutualIntersector::process(SegmentString::ConstVect* segStrings)
 {
-    processCounter = indexCounter + 1;
-    nOverlaps = 0;
+    if (!indexBuilt) {
+        for (auto& mc: indexChains) {
+            mc.setId(indexCounter++);
```
dbaston, line 105: One thing I was trying to figure out is why we need to set these id variables. According to the comments it's so we can avoid comparing the same pair of chains twice, or comparing a chain itself. If that's all it's for, we should just be able to use the chain addresses to avoid the same thing, but doing this causes a test failure (I didn't look into the specifics but was hoping it was just a component ordering type thing)

include/geos/noding/MCIndexSegmentSetMutualIntersector.h
```diff
@@ -96,15 +105,15 @@ class MCIndexSegmentSetMutualIntersector : public SegmentSetMutualIntersector {
 
 private:
 
-    typedef std::vector<std::unique_ptr<index::chain::MonotoneChain>> MonoChains;
+    typedef std::vector<index::chain::MonotoneChain> MonoChains;
```
dbaston, line 108: I had turned `MonoChains` into a vector of vectors (which also preserves stable addresses), but this solution is cleaner and performs better

src/noding/MCIndexSegmentSetMutualIntersector.cpp
dbaston: The motivation for getting rid of `id` was that `MonotoneChain` takes up 72 bytes and if that can be brought down to 64 if seems like you'd get faster access to the vector of chains

dr-jts: The `id` is [used](https://github.com/libgeos/geos/blob/master/src/noding/MCIndexNoder.cpp#L85) in `MCIndexNoder`. It ensures that each two chains are compared once only.

dr-jts: Not sure why it's being referenced in this class, though.  It isn't in JTS.  Perhaps vestigial remains of something that was subsequently removed in JTS?

dbaston: Yes, but you can ensure the same thing in C++ with an address comparison

dr-jts: Gotcha.  Seems tacky somehow (but no worse than Shewchuk's Triangle implementation, where he used the low-order address bits as an index into a 3-element array!)

dbaston: I think the word you were looking for is "elegant" :)

src/noding/MCIndexSegmentSetMutualIntersector.cpp
```diff
@@ -58,15 +61,15 @@ MCIndexSegmentSetMutualIntersector::intersectChains()
     MCIndexSegmentSetMutualIntersector::SegmentOverlapAction overlapAction(*segInt);
 
     std::vector<void*> overlapChains;
-    for(const auto& queryChain : monoChains) {
+    for(auto& queryChain : monoChains) {
```
strk, line 64: Do  you need queryChain to be non-const ?

pramsey: Maybe? Not sure how deep into the guts I would have to follow the const chain.

pramsey: It starts running deep into NodedSegmentString, and NSS definitely is not const all the time. Maybe the whole thing should be non-const. https://github.com/pramsey/geos/pull/1

strk: I only asked becuase it _was_ const before, apparently, thus I was wondering why you had to turn into non-const here.

dr-jts:
Some other ideas to reduce size of MCs (which may or may not affect performance for better or worse):

- Instead of the `end` index keep an offset from the  `start` index.  This could be a single unsigned byte if chains are limited to 256 in length (which is fine)
- Don't store the envelope, but read the min/max X/Y directly from the coordinate sequence.  Flags could be kept to indicate which is which.

PR #377, opened 2021-01-02
Add GEOS_CC_LIBS for linking with c++

### LINKED ISSUE #376: geos-config reports wrong libs for mingw/msys2 on Windows
After compiling and installing geos on Windows under mingw and msys2, `-geos-config --libs` prints this:

```
-L/home/eyal0/.local/lib -lgeos-3.8.1
```

However, those flags do not work.  I tried every combination of one of these:

* `-L/home/eyal0/.local/lib`
* `-L/home/eyal0/.local/bin`

with one of these:

* `-lgeos`
* `-llibgeos`
* `-lgeos-3.8.1`
* `-llibgeos-3.8.1`

The only two combinations that work on Windows are:

* `-L/home/eyal0/.local/lib -lgeos`
* `-L/home/eyal0/.local/lib -llibgeos`

On Unix, these two work:

* `-L/home/eyal0/.local/lib -lgeos`
* `-L/home/eyal0/.local/lib -lgeos-3.8.1`

That first one is in common on both so I suggest that geos-config return the former in all cases.

eyal0:
This fixes #376

### DIFF
--- a/macros/geos.m4
+++ b/macros/geos.m4
@@ -1,7 +1,7 @@
 dnl 
 dnl GEOS_INIT (MINIMUM_VERSION)
 dnl
-dnl Test for GEOS: define HAVE_GEOS, GEOS_LIBS, GEOS_C_LIBS, GEOS_CFLAGS, GEOS_VERSION
+dnl Test for GEOS: define HAVE_GEOS, GEOS_LIBS, GEOS_C_LIBS, GEOS_CC_LIBS, GEOS_CFLAGS, GEOS_VERSION
 dnl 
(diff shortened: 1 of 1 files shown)


eyal0:
Hold off on merging this, I will write a test for the CI, too.

eyal0:
Okay, the test is pretty hard to write.  I guess that I could write a `configure.ac` and `Makefile.am` and software and then do a whole build and compile.  That would add a lot of files to the repo just to test a small thing.  I think that it's not a great solution.  So maybe it's not worthwhile to write a unit test for this?

Let me know.

pramsey:
Thanks. Taking this into master for now.

PR #378, opened 2021-01-05
Change the precision setting to be positional in all cases. 

pramsey:


dbaston:
Superseded by integration of ryu?

PR #381, opened 2021-01-11
Remove StackWalker

dbaston:
It is not used in any GEOS code and not included in the build.


pramsey:
Seems like a fine idea to me.

PR #383, opened 2021-01-14
GeometryPrecisionReducer tweaks

dbaston:
Tiny optimizations. Not terribly effective (10%) but committing them will prevent a future self from trying them again.


pramsey:
Looks good to me 👍 
We need to get those windows failures back to green 😢

dbaston:
I don't see myself spending the time digging into the MSVC issues here.

PR #384, opened 2021-01-15
Simplify geosop dispatch

dbaston:
Instead of requiring every geometry function to have the same signature
(geometry, geometry, double), pass a vector of untyped arguments.

### DIFF
--- a/util/geosop/GeomFunction.cpp
+++ b/util/geosop/GeomFunction.cpp
@@ -46,117 +46,129 @@ using geos::operation::overlayng::OverlayNG;
 /* static private */
 std::map<std::string, GeomFunction*> GeomFunction::registry;
 
-static std::unique_ptr<const PreparedGeometry> prepGeomCache;
-static Geometry *cacheKey;
+class PreparedGeometryCache {
+public:
+    const PreparedGeometry* get(const Geometry* key) {
+        if (m_key != key) {
+            m_pg = PreparedGeometryFactory::prepare(key);
+            m_key = key;
+        }
+
+        return m_pg.get();
+    }
+
+private:
+    std::unique_ptr<const PreparedGeometry> m_pg;
+    const Geometry* m_key;
+};
+
+PreparedGeometryCache prepGeomCache;
 
 /* static */
 void
 GeomFunction::init()
 {
     add("area",
-        [](const std::unique_ptr<Geometry>& geom, const std::unique_ptr<Geometry>& geomB, double d)->Result* {
-            return new Result( geom->getArea() );
+        [](const std::vector<Argument> & args) {
+            return args[0].getGeometry()->getArea();
         });
     add("boundary",
-        [](const std::unique_ptr<Geometry>& geom, const std::unique_ptr<Geometry>& geomB, double d)->Result* {
-            return new Result( geom->getBoundary() );
+        [](const std::vector<Argument> & args) {
+            return args[0].getGeometry()->getBoundary();
         });
-    add("buffer", "cmputes the buffer of geometry A", 1, 1,
-        [](const std::unique_ptr<Geometry>& geom, const std::unique_ptr<Geometry>& geomB, double d)->Result* {
-            return new Result( geom->buffer( d ) );
+    add("buffer", "computes the buffer of geometry A", 1, 1,
+        [](const std::vector<Argument> & args) {
+            return args[0].getGeometry()->buffer(args[1].getDouble());
         });
     add("centroid",
-        [](const std::unique_ptr<Geometry>& geom, const std::unique_ptr<Geometry>& geomB, double d)->Result* {
-            return new Result( geom->getCentroid() );
+        [](const std::vector<Argument> & args) {
+            return std::unique_ptr<Geometry>(args[0].getGeometry()->getCentroid());
         });
     add("copy",
-        [](const std::unique_ptr<Geometry>& geom, const std::unique_ptr<Geometry>& geomB, double d)->Result* {
-            return new Result( geom->clone() );
+        [](const std::vector<Argument> & args) {
+            return args[0].getGeometry()->clone();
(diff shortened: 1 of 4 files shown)


pramsey:
👍

dr-jts:
I dunno... this seems a bit too abstract, and perhaps too low-level.

What happens if the args vector is empty?

Seems like the functions should do some checking to ensure that they are not going to blow a seg-fault.  But would be nice to factor out that checking, to avoid huge amounts of duplicated code.  I did think that perhaps functions should include a signature for their param types in the metadata, which would allow some checking.  Perhaps instead of a raw `vector` the function parameter should be an `Arguments` class, which could encapsulate type checking?

One thing I was thinking of allowing for is making the first Geometry arg optional.  For example, JtsOp does this to allow passing an optional bounding box to things like point or grid generators, to allow them to be "located" on the plane, or default to a standard bounding box.  This would require some slightly complex code to check for this with this proposed change.

dbaston:
It seemed like a more general solution than requiring every possible function to have the same signature, and then dealing with the compiler warnings produced by ignoring unused arguments.

dr-jts:
Yes, definitely more general.  Good point about the warnings - I wasn't seeing those.

And my point about complexity of checking is probably misplaced.  Really the checking should be done before the function is called.  I think the current `geosop` logic won't call a function with a null first geometry argument.  But it could call a binary function with a single Geometry arg - it should check for that ahead of time.

dr-jts:
I'm still concerned that mixing the geometry params and the attribute params is going to make it awkward to write functions with optional arguments.  

What about splitting the input into two vectors, one for the geometry args and one for the rest? That will make it easy to determine if any geometry or attribute args have been omitted.

dbaston:
It would be easy to have `Argument::getDouble()` throw if the argument was not in fact a double.

If there are going to be a lot of functions with optional arguments then maybe a map is a better choice than a vector.

I'm not clear on the benefit of putting Geometry type arguments in a separate container from the other argument types.

dr-jts:
> It would be easy to have `Argument::getDouble()` throw if the argument was not in fact a double.

Yes, and that's probably a good idea. But really the type-checking should take place before the function is even called, so that the command can fail early and provide solid feedback to the user. 

> If there are going to be a lot of functions with optional arguments then maybe a map is a better choice than a vector.

Argument order is essential. Not sure how this would provide that?

> I'm not clear on the benefit of putting Geometry type arguments in a separate container from the other argument types.

Benefits are:
* the syntax for functions is [geometry [geometry]] scalar*.  Having two vectors represents that more accurately
* allows first vector to be restricted to geometry contents
* simplifies checking the actual number of geometry args provided

dbaston:
> Yes, and that's probably a good idea. But really the type-checking should take place before the function is even called, so that the command can fail early and provide solid feedback to the user.

I was assuming the exception would be caught and used to provide an error message like "buffer called with double, expected a geometry"

> Argument order is essential. Not sure how this would provide that?

I don't see why it's essential or even desirable. I would find calling `buffer ( "a" : geomA, "distance" : 20, "quadSegs" : 20, endCap : "miter" )` more expressive than `buffer(geomA, 20, 20, nullptr, "miter")`

Anyway, probably doesn't make sense to design the whole thing in a pull request. I'm fine with closing this and leaving it alone.

dr-jts:
> > Yes, and that's probably a good idea. But really the type-checking should take place before the function is even called, so that the command can fail early and provide solid feedback to the user.
> 
> I was assuming the exception would be caught and used to provide an error message like "buffer called with double, expected a geometry"

Yes, could do that, but validity checking by exception handling is not the best strategy, IMO.

> > Argument order is essential. Not sure how this would provide that?
> 
> I don't see why it's essential or even desirable. I would find calling `buffer ( "a" : geomA, "distance" : 20, "quadSegs" : 20, endCap : "miter" )` more expressive than `buffer(geomA, 20, 20, nullptr, "miter")`

Ok, I understand what you were thinking now.  But where would the parameter names come from?  The buffer operations parameters are by far the most complicated in the API, so not sure they should be the driving motivation for design.

util/geosop/GeomFunction.cpp
```diff
@@ -46,144 +46,160 @@ using geos::operation::overlayng::OverlayNG;
 /* static private */
 std::map<std::string, GeomFunction*> GeomFunction::registry;
 
-static std::unique_ptr<const PreparedGeometry> prepGeomCache;
-static Geometry *cacheKey;
+class PreparedGeometryCache {
```
dr-jts, line 49: This is a nice addition.  Shall I add it independently, or do you want to make a separate PR?

util/geosop/GeomFunction.h
```diff
@@ -55,8 +54,43 @@ class Result {
...
+        return m_value.int_val;
+    }
+
+    const Geometry* getGeometry() const {
+        return m_value.geom_val;
+    }
+
+private:
+
+    union {
```
dr-jts, line 86: I should switch Result to use a union.

util/geosop/GeomFunction.h
```diff
@@ -55,8 +54,43 @@ class Result {
     } typeCode;
 };
 
-typedef std::function<Result *( const std::unique_ptr<Geometry>&, const std::unique_ptr<Geometry>&, double d )>
-    geomFunSig;
+class Argument {
```
dr-jts, line 57: I wonder if `Argument` and `Result` can be merged into a single `Value` type?  But Results own their data, whereas Arguments don't - so maybe not?

util/geosop/GeomFunction.cpp
dbaston: Probably best for you to proceed independently.

dr-jts: Done- thanks.

PR #385, opened 2021-01-15
Add XML overlay test Shapely#1061

jorisvandenbossche:
Adding a test case for https://github.com/Toblerity/Shapely/issues/1061 (case that fails with current shapely using GEOS 3.8, fixed in 3.9.0 with OverlayNG)


tests/xmltester/tests/robust/overlay/TestOverlay-shapely-1061.xml
```diff
@@ -0,0 +1,18 @@
+<run>
+<precisionModel type="FLOATING" />
+
+<case>
+<desc>
+  Failure in intersection of two geometries in GEOS < 3.9.0, from Shapely report.
```
dbaston, line 6: Annoying, but this errors out for me because `<` is a special character.

dr-jts: That is the one annoyance of XML.  Use amp-lt-semicolon, or reword to "in GEOS 3.8"

jorisvandenbossche: Ah, tests passed locally, but easy enough to update!

PR #388, opened 2021-01-18
Add clustering functions

dbaston:
Adds the DBSCAN, simple distance, and intersection clustering that's available in PostGIS. Also adds a clustering based on envelope intersection.


nyalldawson:
@dbaston I'm wondering how these are intended to be used by clients. If I'm reading the code correctly, they return clusters as lists of geometries only. This means that clients who want to attribute data for the clusters would need to then do a second pass where the returned geometries are compared to the input geometries in order to match cluster members back to the original attributes.

To me this would be much more useful if it was possible to attach some user identifier to the input geometries and get this back with the returned clusters. Do you have any thoughts here?

dbaston:
Yes, the returned clusters are geometries only. The returned geometries are the same as the input geometries (they are not copied), so a caller could also set up a map to associate the returned geometry back to the original, or the caller could set the `userData` on the input geometries and then read that back from the result. The only alternative API I can think of would be to return a list of lists of numeric component IDs. But that gets kind of ugly in C, since we also have to return the length of those lists or make them null-terminated.

Another possible issue I see with this interface is that the caller is effectively destroying any geometries that are do not end up in a cluster (that's a possibility for DBSCAN) and transferring ownership of the rest to the cluster. That may be undesirable, but the alternative (copy all the inputs) isn't great either. It would be helpful if GEOS had a GeometryCollection that did not own its components.

I'm very open to ideas, though. Is this something you see an application for?

include/geos/operation/cluster/UnionFind.h
```diff
@@ -0,0 +1,137 @@
...
+    size_t find(size_t i) {
+        size_t root = i;
+
+        while(clusters[root] != root) {
+            root = clusters[root];
+        }
+
+        while (i != root) {
+            size_t next = clusters[i];
+            clusters[i] = root;
```
dr-jts, line 72: am surprised to find that `find` mutates the cluster indices!

include/geos/operation/cluster/UnionFind.h
```diff
@@ -0,0 +1,137 @@
...
+            i = next;
+        }
+
+        return i;
```
dr-jts, line 76: slightly clearer to say `return root`?

PR #389, opened 2021-01-28
Remove Travis configuration

dbaston:
It does not appear to have been running for a couple of months. Even if it was, it would be redundant with GitHub Actions.


strk:
Please also update https://trac.osgeo.org/geos/#BuildandInstall once this is merged

PR #391, opened 2021-01-28
Add GEOSDensify to CAPI

brendan-ward:
This adds `GEOSDensify` to the C API.

It was necessary to guard for empty and `NULL` geometries, or they result in segfaults from the `Densifier`.  Is that expected, or should those be checked there instead?

I wasn't sure if `NULL` inputs should be guarded, but avoiding segfaults for those seemed like a good idea.

The densifier behavior was not exactly as I expected when testing: I expected any segments _greater_ than `tolerance` to be subdivided; the behavior is that any segments greater than or equal to `tolerance` are subdivided.  I tried to clarify that in the function docstring, and I added specific tests for this to the test suite.


pramsey:
I tend towards a send-an-empty-get-an-empty, send-a-null-get-a-null policy, I don't think things that segfault, but frankly I don't know what our policy is, whether it's "avoid the footguns" or "no footguns". @dbaston ?

dbaston:
I'd lean towards returning EMPTY on EMPTY (at the Densifier level, not specifically to the C API) and segfaulting on NULL. Dereferencing a null pointer is a programming error and I don't think runtime guards against programming errors are generally a good idea.

brendan-ward:
Possible compatibility issue: PostGIS interprets tolerance as `greater than` rather than `greater than or equal to` as is done here (I prefer the PostGIS interpretation).  Not sure about JTS.

A tolerance of 10 in PostGIS matches the inputs for our first test case, whereas GEOS requires >10 (all segments in input geometry are <= 10).

It might be good to resolve this before making this publicly available in the CAPI.  Should I modify `Densifier` to handle this (perhaps should be a separate PR)?

tests/unit/capi/GEOSDensifyTest.cpp
```diff
@@ -0,0 +1,234 @@
...
+{
+    geom1_ = GEOSGeomFromWKT("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))");
+    ensure(geom1_ != nullptr);
+    GEOSSetSRID(geom1_, 3857);
+
+    geom2_ = GEOSDensify(geom1_, 10.00000000001);
+
+    ensure("result not null", geom2_ != nullptr);
+    ensure_equals("result == expected", GEOSEqualsExact(geom2_, geom1_, 0), 1);
+    ensure_equals("result SRID == expected SRID", GEOSGetSRID(geom2_), GEOSGetSRID(geom1_));
```
pramsey, line 85: Destroy geom1_ and geom2_ I think, so the test doesn't leak?

pramsey:
This behaviour in GEOS would be coming from JTS. @dr-jts might have an opinion/explanation.

capi/geos_ts_c.cpp
```diff
@@ -1110,6 +1111,24 @@ extern "C" {
         });
     }
 
+    Geometry*
+    GEOSDensify_r(GEOSContextHandle_t extHandle, const Geometry* g, double tolerance)
+    {
+        using geos::geom::util::Densifier;
+
+        return execute(extHandle, [&]() {
+            if (g->isEmpty()) {
```
dbaston, line 1120: Can you please move this logic into `Densifier` ?

tests/unit/capi/GEOSDensifyTest.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include <cstdlib>
+#include <cstring>
+
+namespace tut {
+//
+// Test Group
+//
+
+// Common data used in test cases.
+struct test_capigeosdensify_data {
```
dbaston, line 19: Can reduce boilerplate by including "capi_test_utils.h" and inheriting from `capitest::utility`. See `GEOSBoundaryTest` for an example.

brendan-ward:
It looks like the exact tolerance value is not specifically tested in the [JTS densifier tests](https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/org/locationtech/jts/densify/DensifierTest.java)

The description there is a bit ambiguous on this front as well:
>  All segments in the created densified geometry will be no longer than the given distance tolerance.

dbaston:
Looks great to me. Thanks for this, @brendan-ward !

pramsey:
@brendan-ward As long as we're consistent with JTS behaviour I don't have any particular horse in the race.

dr-jts:
> This behaviour in GEOS would be coming from JTS. @dr-jts might have an opinion/explanation.

This looks like a bug in the JTS code.  I agree that only segments **longer** than the distance tolerance should be densified.  This is the semantics implied by a strict interpretation of:

> All segments in the created densified geometry will be no longer than the given distance tolerance.

I will change the JTS behaviour to match this.  It's good to have it match PostGIS.

pramsey:
There you go @brendan-ward, match the postgis behaviour and let me know, I'll merge.

dr-jts:
> There you go @brendan-ward, match the postgis behaviour and let me know, I'll merge.

I am about to pull the PR trigger in JTS, so might want to look at that code when it lands.  (Not that it is definitely the World's Best Code, but nice to match it unless there's a good reason).

dr-jts:
JTS [PR 676](https://github.com/locationtech/jts/pull/676) has this fix.

brendan-ward:
Thanks @dr-jts !

I'm not sure the fix was quite complete though; the original approach would create more segments than necessary to satisfy the requirement of <= `tolerance`.  It also looks like [part of the fix](https://github.com/locationtech/jts/pull/676/files#diff-642fe2a0dd205fac310b1399efe234927911a796d9d302abc67c2b50ddfac300R91-R92) would add a duplicate last coordinate.

I adapted the intent here and used `ceil` to calculate the number of new segments required for each to be less than or equal to `tolerance`.  Some segments were originally cut more than they needed to be; e.g., a `tolerance` of 5 on a segment that was 10 long would produce 3 segments rather than the required 2.

This required some minor updates to the existing XMLtests here to match the new segment lengths.

tests/xmltester/tests/general/TestDensify.xml
```diff
@@ -16,39 +16,47 @@
...
+  <a>    MULTIPOLYGON (((0 0, 0 70, 70 70, 70 0, 0 0),
+  (10 10, 10 60, 60 60, 10 10)),
   ((80 110, 80 70, 120 70, 120 110, 80 110)))  </a>
-<test><op name="densify" arg1='A'  arg2='10.0'>    
-MULTIPOLYGON (((80 110, 88 110, 96 110, 104 110, 112 110, 120 110, 120 102, 120 94, 120 86, 120 78, 120 70, 112 70, 104 70, 96 70, 88 70, 80 70, 80 78, 80 86, 80 94, 80 102, 80 110)), 
-  ((0 0, 0 8.75, 0 17.5, 0 26.25, 0 35, 0 43.75, 0 52.5, 0 61.25, 0 70, 8.75 70, 17.5 70, 26.25 70, 35 70, 43.75 70, 52.5 70, 61.25 70, 70 70, 70 61.25, 70 52.5, 70 43.75, 70 35, 70 26.25, 70 17.5, 70 8.75, 70 0, 61.25 0, 52.5 0, 43.75 0, 35 0, 26.25 0, 17.5 0, 8.75 0, 0 0), 
-    (10 10, 16.25 16.25, 22.5 22.5, 28.75 28.75, 35 35, 41.25 41.25, 47.5 47.5, 53.75 53.75, 60 60, 51.66666666666667 60, 43.333333333333336 60, 35 60, 26.666666666666668 60, 18.333333333333336 60, 10 60, 10 51.66666666666667, 10 43.333333333333336, 10 35, 10 26.666666666666668, 10 18.333333333333336, 10 10)))      	</op></test>
+<test><op name="densify" arg1='A'  arg2='10.0'>
+  MULTIPOLYGON (((80 70, 80 80, 80 90, 80 100, 80 110, 90 110, 100 110, 110 110, 120 110, 120 100, 120 90, 120 80, 120 70, 110 70, 100 70, 90 70, 80 70)),
```
brendan-ward, line 56: Note: first ring normalized (rotated) a bit differently than previous version.  The primary change on this ring is that segments are now `10` long, as expected from the tolerance.

brendan-ward: If you are open to it, I'd like to simplify the coordinates of these densify tests to make it easier to spot and verify the expected output.

dr-jts: Excellent idea.

dr-jts:
> I'm not sure the fix was quite complete though

Good point.  I will follow your lead on the correct code.

dr-jts:
JTS [PR 677](https://github.com/locationtech/jts/pull/677) has the improved splitting algorithm from this PR.  

It also adds tests to the XML test file to capture the various additional tests in the unit test code (since it is more portable to test them in the XML files).  This could be copied to this PR.

Thanks for working on this @brendan-ward - this is a great improvement over the old code.

tests/unit/capi/GEOSDensifyTest.cpp
```diff
@@ -0,0 +1,214 @@
...
+    GEOSSetSRID(geom1_, 3857);
+
+    geom2_ = GEOSDensify(geom1_, 5.0);
+
+    ensure("result not null", geom2_ != nullptr);
+    ensure_geometry_equals(
+        geom2_,
+        "POLYGON ((0 0, 5 0, 10 0, 10 5, 10 10, 5 10, 0 10, 0 5, 0 0), (1 1, 1 4, 1 7, 4 7, 7 7, 7 4, 7 1, 4 1, 1 1))");
+
+    ensure_equals("result SRID == expected SRID", GEOSGetSRID(geom2_), 3857);
```
dr-jts, line 87: Sure would be nice to reduce the amount of code for tests like this (not just here, for all GEOS unit tests).

Any ideas, @dbaston ?  Perhaps some more helper functions?

dbaston: If I were writing it I would drop lines 77 and 82, but overall it seems pretty reasonable to me.

dr-jts: I don't have a problem with those lines per se (although if they are redundant would be good to remove them).  The real problem is how many lines of code are repeated in each test.  The only information content is the input WKT, the densify tolerance, and the result WKT. In JTS I factor out the boilerplate into a separate function which is called by each test.  I'm just not sure if the tut templated approach supports that?

dr-jts: It turns out that factoring out the test common code into a support function is very doable.  See [`GEOSMinimumClearanceTest`](https://git.osgeo.org/gitea/geos/geos/src/branch/master/tests/unit/capi/GEOSMinimumClearanceTest.cpp).cpp for an example.

@brendan-ward how do you feel about making this improvement?

dbaston: Yes, some of the tests do something like `doDensifierTest`. I actually find those harder to work with, mostly because our test framework has no way to run an individual test. So if you want to debug a test and the test code itself is in a common function, you have to comment out all of the remaining tests.

dbaston: Another negative is when trying to read a test, you have to follow a trail of function calls to figure out what `doDensifierTest` even means. I think the style of good test code is a bit different from implementation code, and good test code is about easy readability, not compactness.

dbaston: TIL the software engineering consultants call this "DAMP vs DRY"...there's always an acronym (or two)!

dr-jts: >  So if you want to debug a test and the test code itself is in a common function, you have to comment out all of the remaining tests.

I don't see how this is affected by factoring out the common code?

dr-jts: > Another negative is when trying to read a test, you have to follow a trail of function calls to figure out what `doDensifierTest` even means. 

Usually there is only a single function that needs to be understood.  And you could make the opposite argument - by using a common function you only need to understand that function to know what every test that uses it does, whereas if the code is duplicated you can't assume that tests always do the same thing, so need to read them individually (and keep them all in mind looking for minor differences).  These seems way more error-prone - there is a serious risk of some part of the test being omitted in a single test, which is unlikely to be noticed.

pramsey: Yeah, it's a balance. I've found having support functions tucked away in another file makes things more painful, but having them in the actual Test file is a-OK. Also, we can run individual *tests* (not assertions) with `./bin/test_geos_unit geos::io::WKBReader 6` so having more numbered chunks is often preferable to a long list of assertions, or a loop over a data array (my nemesis).

dbaston: > Also, we can run individual tests (not assertions) with ./bin/test_geos_unit geos::io::WKBReader 6

How am I only learning this now? 🤦

pramsey: I only found it out a couple months ago! I was like... "what if I add the number..."

pramsey: (Also, the numbers don't have to be sequential)

dbaston: > And you could make the opposite argument

Of course. My experience makes me less inclined to write something today the way I wrote the minimum clearance tests 5 years ago. Your experience seems to pull you in the other direction. That's fine.

dbaston:
I've merged this in, but please don't hesitate to submit a new PR if you want to make additional changes. Thanks!

PR #392, opened 2021-01-29
Read unsigned integers in WKBReader

dbaston:
OGC WKB spec specifies that the integers are unsigned. Use of signed
integers is probably a carryover from Java, which does not have unsigned
integers. Clear type conversion warnings in WKBReader and use C++11
int64_t type.

### DIFF
--- a/benchmarks/ClassSizes.cpp
+++ b/benchmarks/ClassSizes.cpp
@@ -76,6 +76,6 @@ main()
     check(triangulate::quadedge::QuadEdge);
     check(triangulate::quadedge::QuadEdgeQuartet);
     check(triangulate::quadedge::Vertex);
-    check(int64);
+    check(int64_t);
 }
 
--- a/include/geos/constants.h
+++ b/include/geos/constants.h
@@ -31,9 +31,6 @@ typedef __int64 int64;
 #include <limits>
 #include <cinttypes>
 
-
-typedef int64_t int64;
-
 namespace geos {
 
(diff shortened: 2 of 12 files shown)


pramsey:
I guess the fix for OOM is still to come? It looks like CoordinateSequence reading

https://github.com/dbaston/libgeos/blob/wkbreader-unsigned/src/io/WKBReader.cpp#L431

starts by allocating a new CoordinateArraySequence with a (size, dims) constructor

https://github.com/dbaston/libgeos/blob/wkbreader-unsigned/src/geom/CoordinateArraySequence.cpp#L36-L41

which initializes the underlying vector using the (count) constructor

https://en.cppreference.com/w/cpp/container/vector/vector

Which does

> "4) Constructs the container with count default-inserted instances of T. No copies are made."

So if you feed a degenerate WKB linestring with a 2B count, you can get GEOS to allocate the whole 2B before you actually try and read them and find the end on the input stream.

dr-jts:
> I guess the fix for OOM is still to come? 

JTS has fixed this in [PR 675](https://github.com/locationtech/jts/pull/675), at least for the case when the size of the input byte sequence is known.  It computes an upper bound on the value of all `num*` fields in the WKB, based on the size of the input. The `num*` fields are then checked against that limit.

dbaston:
> I guess the fix for OOM is still to come?

Right, this is unrelated to OOM.

PR #393, opened 2021-01-30
Avoid OOM on over-large WKB counts

pramsey:
A malformed WKB can include a point/ring/geometry count that is as large as the maximum integer size. The parser cannot know if there's data to support that count without reading forward to see. The current parser pre-allocates vectors to hold future entries. This change does away with pre-allocation and just grows the vectors as necessary. If the count is over-large, the parser will eventually hit the end of the buffer and complain.


dbaston:
Curious why you took this approach instead of sanity-checking reported ngeoms/npoints against remaining size in the buffer. I'd expect some performance and memory penalty from this?

pramsey:
I wasn't sure we could find the size of an arbitrary input stream, and this seemed expedient. The JTS approach did as you suggest but was left asserting that "anything less that MAXINT" was OK in the case of an inputstream-only reader. That could still, IMO, lead to OOM.

dbaston:
I'm not sure about the JTS approach, but `WKBReader::dis.size()` should provide the remaining size. We're only reading from fixed buffers, not streams.

With the incremental approach I'm hesitant about performance (copying the same coordinates over and over as the vector is incrementally resized) and memory usage (unneeded space in the vector is never reclaimed)

src/io/WKBReader.cpp
```diff
@@ -427,35 +430,40 @@ WKBReader::readGeometryCollection()
...
 WKBReader::readCoordinateSequence(uint32_t size)
 {
-    unsigned int targetDim = 2 + (hasZ ? 1 : 0);
-    auto seq = factory.getCoordinateSequenceFactory()->create(size, targetDim);
+    uint32_t targetDim = 2 + (hasZ ? 1 : 0);
+    // auto seq = factory.getCoordinateSequenceFactory()->create(size, targetDim);
     if(targetDim > inputDimension) {
         targetDim = inputDimension;
     }
+    std::vector<geom::Coordinate> curCoords;
```
dbaston, line 438: Building this up in a vector instead of a CoordinateSequence seems like a win no matter what

PR #394, opened 2021-01-30
Exclude vendored libs from GEOS compiler warnings

dbaston:
This PR is an attempt to stop the propagation of GEOS diagnostic flags to vendored libraries (tinyxml, ryu). The vendored code is moved into `OBJECT` libraries that are treated as separate entities in CMake but end up rolled into the consuming targets (`ryu` sources end up in `libgeos`; there is no `libryu` emitted.)


dbaston:
With this PR, I've removed all warnings from the GEOS builds, which will let us turn on `-Werror` (except for GeosOp; see https://trac.osgeo.org/geos/ticket/1098). I'm curious if anyone more experienced with CMake than I am has a better solution for the vendored libraries, though.

PR #395, opened 2021-01-30
CMakeLists.txt: add BUILD_BENCHMARKS

ffontaine:
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,7 @@ option(BUILD_SHARED_LIBS "Build GEOS with shared libraries" ON)
 set(CMAKE_CXX_STANDARD 11 CACHE STRING "C++ standard version to use (default is 11)")
 
 ## GEOS custom variables
+option(BUILD_BENCHMARKS "Build GEOS benchmarks" ON)
 cmake_dependent_option(GEOS_BUILD_DEVELOPER
   "Build with compilation flags useful for development" ON
   "GEOS_BUILD_FROM_GIT" OFF)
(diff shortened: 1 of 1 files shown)


dbaston:
Out of curiosity, is there some problem the benchmarks are causing?

ffontaine:
Yes, they raise some build failures on buildroot autobuilders:

```
[100%] Linking C executable ../../bin/perf_memleak_mp_prep
/home/giuliobenetti/autobuild/run/instance-2/output-1/host/opt/ext-toolchain/bin/../lib/gcc/riscv64-buildroot-linux-musl/9.2.0/../../../../riscv64-buildroot-linux-musl/bin/ld: warning: libgeos.so.3.9.0, needed by ../../lib/libgeos_c.so.1.16.2, not found (try using -rpath or -rpath-link)
```

Full build logs can be found here: http://autobuild.buildroot.org/?reason=libgeos%

I'm not able to reproduce them on my machine with cmake 3.18.4 so I think that these build failures are only raised with "old" cmake (e.g. 3.15.5).

Because buildroot targets embedded linux systems, it makes sense to just disable benchmarks for the moment.

dbaston:
I think the issue is that `perf_memleak_mp_prep` links to `geos_c` only:

https://github.com/libgeos/geos/blob/546f7846f548cbcafe65605f95641aa40f7db443/benchmarks/capi/CMakeLists.txt#L18

which would be OK if `geos_c` had `PUBLIC` linkage to `geos`, but instead it is `PRIVATE`:

https://github.com/libgeos/geos/blob/546f7846f548cbcafe65605f95641aa40f7db443/CMakeLists.txt#L260

I think the right fix here is to use `PUBLIC` linkage, since anything that links to `geos_c` must also link to `geos`.

ffontaine:
Indeed, this could fix the build failure. Still, do you think it would also make sense to merge this PR?

dbaston:
I'm fine with it. It sounds like you need it in the stable 3.9 branch?

ffontaine:
Yes, it would be perfect if you could add it in the stable 3.9 branch.

PR #397, opened 2021-02-01
Make polygon comparison compare shell and holes

brendan-ward:
This PR makes sure that `Polygon::compareToSameClass`` compares holes as well as outer shell and adds a 
 corresponding test.

Previously, the comparison was only based on comparing the outer shell.

Resolves [#1099](https://trac.osgeo.org/geos/ticket/1099)

This was adapted from [JTS](https://github.com/locationtech/jts/blob/jts-1.18.0/modules/core/src/main/java/org/locationtech/jts/geom/Polygon.java#L375-L396).  Note that the first variant of JTS `compareToSameClass` has the same issue as was present here.

I adapted another test in that unit test file to create the tests here; it was not obvious how to reduce the boilerplate in this case (i.e., simpler ways to instantiate from WKT here).

### DIFF
--- a/src/geom/Polygon.cpp
+++ b/src/geom/Polygon.cpp
@@ -334,8 +334,8 @@ Polygon::compareToSameClass(const Geometry* g) const
         return shellComp;
     }
 
-    int nHole1 = getNumInteriorRing();
-    int nHole2 = p->getNumInteriorRing();
+    size_t nHole1 = getNumInteriorRing();
+    size_t nHole2 = p->getNumInteriorRing();
     if (nHole1 < nHole2) {
         return -1;
     }
@@ -343,12 +343,10 @@ Polygon::compareToSameClass(const Geometry* g) const
         return 1;
     }
 
-    const LinearRing *lr1, *lr2;
     int holeComp = 0;
-    for (int i=0; i < nHole1; i++) {
-        lr1 = dynamic_cast<const LinearRing*>(getInteriorRingN(i));
-        lr2 = dynamic_cast<const LinearRing*>(p->getInteriorRingN(i));
-        holeComp = lr1->compareToSameClass(lr2);
+    for (size_t i=0; i < nHole1; i++) {
+        const LinearRing *lr = p->getInteriorRingN(i);
+        holeComp = getInteriorRingN(i)->compareToSameClass(lr);
         if (holeComp != 0) {
             return holeComp;
         }
--- a/src/geom/Polygon.cpp
+++ b/src/geom/Polygon.cpp
@@ -329,7 +329,32 @@ int
 Polygon::compareToSameClass(const Geometry* g) const
 {
     const Polygon* p = dynamic_cast<const Polygon*>(g);
-    return shell->compareToSameClass(p->shell.get());
+    int shellComp = shell->compareToSameClass(p->shell.get());
+    if (shellComp != 0) {
+        return shellComp;
+    }
+
+    int nHole1 = getNumInteriorRing();
+    int nHole2 = p->getNumInteriorRing();
+    if (nHole1 < nHole2) {
+        return -1;
+    }
+    if (nHole1 > nHole2) {
+        return 1;
+    }
+
+    const LinearRing *lr1, *lr2;
+    int holeComp = 0;
+    for (int i=0; i < nHole1; i++) {
+        lr1 = dynamic_cast<const LinearRing*>(getInteriorRingN(i));
+        lr2 = dynamic_cast<const LinearRing*>(p->getInteriorRingN(i));
+        holeComp = lr1->compareToSameClass(lr2);
+        if (holeComp != 0) {
+            return holeComp;
+        }
+    }
+
+    return 0;
 }
 
 /*
--- a/tests/unit/geom/PolygonTest.cpp
+++ b/tests/unit/geom/PolygonTest.cpp
@@ -611,4 +611,39 @@ void object::test<42>
     ensure(!poly_->isDimensionStrict(geos::geom::Dimension::L));
 }
 
+
+// test compareToSameClass for polygons including holes
+template<>
+template<>
+void object::test<43>
+()
+{
+    auto geo = reader_.read("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))");
+    ensure(geo != nullptr);
+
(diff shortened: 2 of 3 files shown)


src/geom/Polygon.cpp
```diff
@@ -329,7 +329,32 @@ int
...
+        return -1;
+    }
+    if (nHole1 > nHole2) {
+        return 1;
+    }
+
+    const LinearRing *lr1, *lr2;
+    int holeComp = 0;
+    for (int i=0; i < nHole1; i++) {
+        lr1 = dynamic_cast<const LinearRing*>(getInteriorRingN(i));
```
dbaston, line 349: `getInteriorRingN` returns a `LinearRing*` so there should be no need to cast.

src/geom/Polygon.cpp
```diff
@@ -329,7 +329,32 @@ int
 Polygon::compareToSameClass(const Geometry* g) const
 {
     const Polygon* p = dynamic_cast<const Polygon*>(g);
-    return shell->compareToSameClass(p->shell.get());
+    int shellComp = shell->compareToSameClass(p->shell.get());
+    if (shellComp != 0) {
+        return shellComp;
+    }
+
+    int nHole1 = getNumInteriorRing();
```
dbaston, line 337: Watch the signed vs signed types...`getNumInteriorRing` returns a `size_t`.

tests/unit/geom/PolygonTest.cpp
```diff
@@ -611,4 +611,39 @@ void object::test<42>
...
+// test compareToSameClass for polygons including holes
+template<>
+template<>
+void object::test<43>
+()
+{
+    auto geo = reader_.read("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))");
+    ensure(geo != nullptr);
+
+    PolygonPtr poly = dynamic_cast<PolygonPtr>(geo.get());
```
dbaston, line 624: You're testing the `Polygon` implementation of `Geometry::compareToSameClass`, so if you want to reduce boilerplate, you can avoid lines like 630-631 (which really test the WKTReader). I would also take out lines like 622 (again, the goal is not to test the `WKTReader`) but it really comes down to preference.

src/geom/Polygon.cpp
```diff
@@ -329,7 +329,32 @@ int
...
+    int nHole2 = p->getNumInteriorRing();
+    if (nHole1 < nHole2) {
+        return -1;
+    }
+    if (nHole1 > nHole2) {
+        return 1;
+    }
+
+    const LinearRing *lr1, *lr2;
```
dbaston, line 346: No need to declare these outside the loop; declare them where you define them

tests/unit/geom/PolygonTest.cpp
```diff
@@ -611,4 +611,39 @@ void object::test<42>
...
+    ensure(poly2 != nullptr);
+
+    auto geo3 = reader_.read("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))");
+    ensure(geo3 != nullptr);
+
+    PolygonPtr poly3 = dynamic_cast<PolygonPtr>(geo3.get());
+    ensure(poly3 != nullptr);
+
+    // polygon should equal self
+    ensure(poly->compareToSameClass(poly) == 0);
```
dbaston, line 640: Consider moving your comments into the assert `ensure("polygon should equal self", poly->compareToSameClass(poly) == 0` which will give a better error message if the assert fails

tests/unit/geom/PolygonTest.cpp
brendan-ward: Sorry to be dense, but I'm not seeing how to avoid these two lines here?

```
auto geo = reader_.read("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))");
PolygonPtr poly = dynamic_cast<PolygonPtr>(geo.get());
```

(total noob to C++ managed pointers)

dbaston: Well, you can't avoid the first :)

But you can call `geo->compareToSameClass(geo2.get())` directly; you don't need to explicitly cast it to a `Polygon`.

brendan-ward: Looks like I need to cast the left side, otherwise `compareToSameClass` is inaccessible.  But no need to cast the function parameter, so that simplifies a little.

brendan-ward:
@dbaston thanks for the fast review and guidance here!

tests/unit/geom/PolygonTest.cpp
dbaston: That's interesting -- `compareToSameClass` is `protected` in `Geometry` but `public` in `Polygon`. I can't think of a good reason for that off hand -- I wonder if it's an oversight.

dbaston: Yeah, `Point::compareToSameClass` is also `protected`, so the `Polygon` variant being public is probably a 15-year old oversight. The alternative is to just call `geo->compareTo(geo2.get())` which is really the public interface. Not asking you to change anything, just trying to explain.

brendan-ward: Thanks @dbaston for helping me understand!  The [comment on it](https://github.com/libgeos/geos/blob/3.9/include/geos/geom/Polygon.h#L137) implies that it was changed intentionally for `Polygon` class?

I think since we're specifically trying to test `compareToSameClass` here, the cast is OK to leave as is.

dr-jts: > That's interesting -- `compareToSameClass` is `protected` in `Geometry` but `public` in `Polygon`. I can't think of a good reason for that off hand -- I wonder if it's an oversight.

Sure seems like an oversight/misunderstanding.  Can this be fixed?

dbaston: > Sure seems like an oversight/misunderstanding. Can this be fixed?

Do you mean make it `protected` and then test `compareTo` instead of `compareToSameClass` ?

dr-jts: > Do you mean make it `protected` and then test `compareTo` instead of `compareToSameClass` ?

Yes.

dbaston:
Squashed and committed as 587049c216625a. Thanks!

PR #398, opened 2021-02-02
Avoid out-of-memory for malformed WKB

pramsey:
By simply creating a WKB input with a vertex count/ring count/geometry count that is very very large, a user can get GEOS to allocate a very large vector in advance of actually parsing the WKB to fill in that vector.

Instead, we do a quick bounds check, comparing the number of objects requested and the minimum size those objects could be with the amount of space left in the input stream, and raising an exception if the minimum possible request size exceeds the amount of input data.

### DIFF
--- a/src/io/WKBReader.cpp
+++ b/src/io/WKBReader.cpp
@@ -171,36 +171,35 @@ WKBReader::readHEX(std::istream& is)
 }
 
 void
-WKBReader::minMemSize(int geomType, uint32_t size)
+WKBReader::minMemSize(int geomType, uint64_t size)
 {
-    uint64_t size64 = (std::size_t)size;
     uint64_t minMemSize = 0;
-    const uint64_t minCoordSize = 2 * sizeof(double);
-    const uint64_t minPtSize = (1+4) + minCoordSize;
-    const uint64_t minLineSize = (1+4+4); // empty line
-    const uint64_t minRingSize = 4; // empty ring
-    const uint64_t minPolySize = (1+4+4); // empty polygon
-    const uint64_t minGeomSize = minLineSize;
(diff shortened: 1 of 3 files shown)


tests/unit/io/WKBReaderTest.cpp
```diff
@@ -52,6 +52,23 @@ struct test_wkbreader_data {
...
     {}
 
+    void
+    testParseError(const std::string& hexwkb, const std::string& errstr)
+    {
+        std::string exstr;
+        std::stringstream hexin(hexwkb);
+        try {
+            GeomPtr g(wkbreader.readHEX(hexin));
+            ensure("Parse did not fail", !g);
```
dbaston, line 62: I think there is a `fail()` or `failure()` or something you can call for this purpose.

src/io/WKBReader.cpp
```diff
@@ -170,6 +170,46 @@ WKBReader::readHEX(std::istream& is)
     return this->read(os);
 }
 
+void
+WKBReader::minMemSize(int geomType, uint32_t size)
+{
+    uint64_t size64 = (std::size_t)size;
+    uint64_t minMemSize = 0;
+    const uint64_t minCoordSize = 2 * sizeof(double);
```
dbaston, line 178: constexpr?

src/io/WKBReader.cpp
```diff
@@ -170,6 +170,46 @@ WKBReader::readHEX(std::istream& is)
     return this->read(os);
 }
 
+void
+WKBReader::minMemSize(int geomType, uint32_t size)
+{
+    uint64_t size64 = (std::size_t)size;
```
dbaston, line 176: Make the `size` argument a `size_t` ?

PR #399, opened 2021-02-02
TST: Expand XML densify tests, reduce boilerplate in C API densify tests

brendan-ward:
This updates the Densify tests based on what I understood to be the direction in #391.

This uses a function in the C API tests to reduce the boilerplate, and since it is a function in the same file, my understanding was that it wouldn't contribute to a decrease in clarity / usability of the tests.  (Note: the last test checks nulls and needed to be a special case).

This also expands the XML tests used for `Densifier`.  I did a bit of cleanup there to make them easier to add and read:
1) used XML formatter; this makes them a bit longer, but removes all the mental overhead of formatting / alignment.
2) pulled all geometry WKT to left margin so that you can more easily see and compare inputs and outputs; this is arguably one of the most important things to see in these test files.

### DIFF
--- a/tests/unit/capi/GEOSDensifyTest.cpp
+++ b/tests/unit/capi/GEOSDensifyTest.cpp
@@ -1,5 +1,5 @@
 //
-// Test Suite for C-API GEOSBuffer and GEOSBufferWithStyle
+// Test Suite for C-API GEOSDensify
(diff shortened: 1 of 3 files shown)


dr-jts:
Very nice!

PR #404, opened 2021-02-04
Remove autotools configuration instructions

eyal0:
autotools configuration was already removed in 29802a665505aa5fa00f942754d6d16ddfed5823

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -7,24 +7,7 @@ compiler.
 
 ### Unix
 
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -10,22 +10,6 @@ compiler.
 GEOS can be built on Unix systems using either the autotools or CMake build
```
dr-jts, line 10: Reference to autotools can be removed here too.

eyal0: Good catch!

PR #407, opened 2021-02-07
Represent empty envelope with NaN

dbaston:
This allows explicit isNull() checks to be avoided in some of the
envelope operations, since NaN semantics make the affected comparisons
false for null envelopes.

Improves performance of watershed union operation by 1%.

### DIFF
--- a/src/operation/overlayng/OverlayNGRobust.cpp
+++ b/src/operation/overlayng/OverlayNGRobust.cpp
@@ -265,7 +265,7 @@ OverlayNGRobust::snapTolerance(const Geometry* geom)
 double
 OverlayNGRobust::ordinateMagnitude(const Geometry* geom)
 {
-    if (geom == nullptr) return 0;
+    if (geom == nullptr || geom->isEmpty()) return 0;
     const Envelope* env = geom->getEnvelopeInternal();
     double magMax = std::max(
         std::abs(env->getMaxX()),
--- a/src/operation/overlayng/RingClipper.cpp
+++ b/src/operation/overlayng/RingClipper.cpp
@@ -83,17 +83,17 @@ RingClipper::intersection(const Coordinate& a, const Coordinate& b, int edgeInde
 {
     switch (edgeIndex) {
     case BOX_BOTTOM:
-        rsltPt = Coordinate(intersectionLineY(a, b, clipEnvMinY), clipEnvMinY);
+        rsltPt = Coordinate(intersectionLineY(a, b, clipEnv.getMinY()), clipEnv.getMinY());
         break;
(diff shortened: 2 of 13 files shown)


src/operation/overlayng/OverlayNGRobust.cpp
```diff
@@ -265,7 +265,7 @@ OverlayNGRobust::snapTolerance(const Geometry* geom)
 double
 OverlayNGRobust::ordinateMagnitude(const Geometry* geom)
 {
-    if (geom == nullptr) return 0;
```
dbaston, line 268: @dr-jts I fixed a couple issues in this commit where the meaningless placeholder values for a null Envelope are used in comparisons. The same issues may be present in JTS?

src/operation/overlayng/RingClipper.cpp
```diff
@@ -120,20 +120,24 @@ RingClipper::intersectionLineX(const Coordinate& a, const Coordinate& b, double
 bool
 RingClipper::isInsideEdge(const Coordinate& p, int edgeIndex) const
 {
+    if (clipEnv.isNull()) {
```
dbaston, line 123: @dr-jts Another one here.

src/operation/overlayng/OverlayNGRobust.cpp
dbaston: Another one (addressed in https://github.com/libgeos/geos/pull/407/commits/f16218f1b13b1268be72788928ca1be037244a74) comes from creating an `QuadEdgeSubdivision` with a null envelope.

dr-jts:
Good idea to change to using NaN as the indicator for a null envelope.

src/operation/overlayng/RingClipper.cpp
dr-jts: Technically yes, but OverlayNG should never call `RingClipper` with an empty clip envelope.

src/operation/overlayng/OverlayNGRobust.cpp
dr-jts: This should never be called with an empty geometry, since that will not create a robustness failure.  But it's good to harden the code.

src/operation/overlayng/RingClipper.cpp
dbaston: Well, the OverlayNG cases are both hit by the test suite -- that's why I added these guards.

dr-jts: Ah right, the unit tests cover more cases than should be seen in practice.  Good to have the guards, anyway.

PR #412, opened 2021-02-18
Remove linear CascadedUnion

dbaston:
This was added as a robustness improvement, but it appears no longer
necessary with the use of OverlayNG. Removing this eliminates a
significant amount of duplicated code.


dr-jts:
Good housekeeping.

dbaston:
Thanks, that sounds more noble than "I don't want to update this code to work with a different tree, so I'm just going to delete it and see what happens" :)

PR #417, opened 2021-03-02
CI: use different versions of CMake in GitHub Actions matrix

mwtoews:
The minimum version of CMake is set at 3.8, but testing often uses much newer versions of CMake that are not installed on (e.g.) Ubuntu 18.04 with cmake 3.10.2.

This PR uses a few different versions of CMake in the CI test matrix on GitHub actions. These were picked more-or-less at random, but [this is a good reference](https://cliutils.gitlab.io/modern-cmake/chapters/intro/newcmake.html) for anyone that thinks other versions should be specified.

PyPi is used to quickly install the specified version; see https://pypi.org/project/cmake/#history

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,8 @@ jobs:
           build_type: Coverage,
           cxxstd: 11,
           arch: 64,
-          packages: 'g++ cmake doxygen',
+          packages: 'g++',
+          cmake: 3.15.*,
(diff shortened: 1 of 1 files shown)


pramsey:
Thanks! Committed!

mwtoews:
Hmm, I'll need a follow-up to fix the PATH, as I see "cmake version 3.19.5" at the top of build ...

PR #418, opened 2021-03-02
CI: add .local/bin to GITHUB_PATH

mwtoews:
Follow-up from #417, to ensure things installed via `pip install --user` are available on the PATH.

Also change CMake version `3.12.*` to `3.13.*`, because there was only one release in the 3.12 series.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           cxxstd: 11,
           arch: 32,
           packages: 'g++-4.8-multilib gcc-4.8-multilib g++-multilib gcc-multilib',
-          cmake: 3.12.*,
+          cmake: 3.13.*,
           os: ubuntu-18.04
         }
         - {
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -92,7 +93,7 @@ jobs:
         cd build.cmake
         cmake --version
         cmake -DCMAKE_CXX_COMPILER=${{ matrix.ci.compiler }} -DCMAKE_CXX_STANDARD=${{ matrix.ci.cxxstd }} -DBUILD_DOCUMENTATION=YES -DCMAKE_BUILD_TYPE=${{ matrix.ci.build_type }} ..
-        cmake --build . -j 2
+        make -j 2
```
dbaston, line 96: What's this change for?

mwtoews: with cmake version 3.10.2
```
$ cmake --build . -j 2
Unknown argument -j
Unknown argument 2
Usage: cmake --build <dir> [options] [-- [native-options]]
Options:
  <dir>          = Project binary directory to be built.
  --target <tgt> = Build <tgt> instead of default targets.
                   May only be specified once.
  --config <cfg> = For multi-configuration tools, choose <cfg>.
  --clean-first  = Build target 'clean' first, then build.
                   (To clean only, use --target 'clean'.)
  --use-stderr   = Ignored.  Behavior is default in CMake >= 3.0.
  --             = Pass remaining options to the native tool.
```
we could use `-- -j 2`, if that's preferred?

dbaston: Thanks, no need to change. I didn't realize that flag was such a recent addition.

mwtoews: It's a [3.12 feature](https://cmake.org/cmake/help/latest/release/3.12.html), to be specific.

PR #419, opened 2021-03-03
Add GeoJSON Reader and Writer

jericks:
This pull request adds a GeoJSON Reader and Writer modeled after the WKT classes.  It uses the nlohmann/json header only library (https://github.com/nlohmann/json) and contains a suite of unit tests.


include/geos/io/GeoJSON.h
```diff
@@ -0,0 +1,109 @@
...
+namespace geos {
+namespace io {
+
+enum class GeoJSONValueType {
+    STRING, NUMBER, BOOLEAN, NULLTYPE, OBJECT, ARRAY
+};
+
+struct GeoJSONValue {
+
+    GeoJSONValueType type;
```
pramsey, line 49: Maybe a union?

src/io/GeoJSON.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+GeoJSONValue GeoJSONValue::createArrayValue() {
+    GeoJSONValue value{};
+    value.type = GeoJSONValueType::ARRAY;
+    return value;
+}
+
+std::string GeoJSONValue::toString() const {
+    std::stringstream os;
```
pramsey, line 65: I would tend to reverse this, have the << operator do all the work and have toString() just call the operator. Allows the << operator to stream if necessary/possible.

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+        return readGeometryCollection(j);
+    }
+    return std::unique_ptr<geom::Geometry>(geometryFactory.createEmptyGeometry());
+}
+
+std::unique_ptr<geom::Point> GeoJSONReader::readPoint(nlohmann::json& j) {
+    std::vector<double> coords = j["coordinates"].get<std::vector<double>>();
+    if (coords.size() == 0) {
+        return std::unique_ptr<geom::Point>(geometryFactory.createPoint(2));
+    } else {
```
pramsey, line 167: What if coords.size() == 1?

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+        std::vector<geom::Coordinate> coordinates;
+        for (std::vector<int>::size_type j = 0; j < polygonCoords[i].size(); j++) {
+            coordinates.push_back(geom::Coordinate{polygonCoords[i][j].first, polygonCoords[i][j].second});
+        }
+        geom::CoordinateArraySequence coordinateSequence { std::move(coordinates) };
+        rings.push_back(geometryFactory.createLinearRing(std::move(coordinateSequence)));
+    }
+    if (rings.size() == 0) {
+        return std::unique_ptr<geom::Polygon>(geometryFactory.createPolygon(2));
+    } else if (rings.size() == 1) {
```
pramsey, line 196: What about one ring with zero coordinates in it? Should also return POLYGON EMPTY.

tests/unit/io/GeoJSONReaderTest.cpp
```diff
@@ -0,0 +1,264 @@
...
+    GeomPtr geom(geojsonreader.read(geojson));
+    ensure_equals("POINT EMPTY", geom->toText());
+}
+
+template<>
+template<>
+void object::test<12>
+()
+{
+    std::string geojson { "{\"type\":\"LineString\",\"coordinates\":[]}" };
```
pramsey, line 164: Need more tests of incorrectly structured coordinates. Just one coordinate is classic.
```[[1,]]```
Or, two coordinates to start, then one.
```[[1,2],[3,]]```

File: tests/unit/io/GeoJSONReaderTest.cpp [hunk 6]
```diff
@@ -0,0 +1,264 @@
...
+    GeomPtr geom(geojsonreader.read(geojson));
+    ensure_equals("LINESTRING EMPTY", geom->toText());
+}
+
+template<>
+template<>
+void object::test<13>
+()
+{
+    std::string geojson { "{\"type\":\"Polygon\",\"coordinates\":[]}" };
```
pramsey, line 174: Non-empty list of empty rings is a fun case.

tests/unit/io/GeoJSONReaderTest.cpp
```diff
@@ -0,0 +1,264 @@
...
+    GeomPtr geom(geojsonreader.read(geojson));
+    ensure_equals("MULTIPOINT EMPTY", geom->toText());
+}
+
+template<>
+template<>
+void object::test<15>
+()
+{
+    std::string geojson { "{\"type\":\"MultiLineString\",\"coordinates\":[]}" };
```
pramsey, line 194: Same as polygon, non-empty list of empty line strings? What happens with overly-wide dimensionality? 3-, 4-, 5- coordinates?

src/io/GeoJSONReader.cpp
jericks: What should we do if there is only one coordinate in the coordinates array?  Return an empty point?

include/geos/io/GeoJSON.h
jericks: I converted this to a union and it turned out to be tricky because of std::string, std::map, and std::vector.  I

include/geos/io/GeoJSON.h
```diff
@@ -0,0 +1,135 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2011 Sandro Santilli <strk@kbt.io>
```
dbaston, line 6: Update copyrights?

src/io/GeoJSON.cpp
```diff
@@ -0,0 +1,210 @@
...
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************/
+
+#include <geos/io/GeoJSON.h>
+
+#undef DEBUG_GEOJSON
```
dbaston, line 17: Remove debug code

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+        }
+        geom::CoordinateArraySequence coordinateSequence { std::move(coordinates) };
+        lines->push_back(geometryFactory.createLineString(std::move(coordinateSequence)));
+    }
+    return std::unique_ptr<geom::MultiLineString>(geometryFactory.createMultiLineString(lines));
+}
+
+std::unique_ptr<geom::MultiPolygon> GeoJSONReader::readMultiPolygon(nlohmann::json& json) {
+    std::vector<std::vector<std::vector<std::pair<double,double>>>> multiPolygonCoords = json["coordinates"].get<std::vector<std::vector<std::vector<std::pair<double,double>>>>>();
+    std::vector<geom::Geometry *>* polygons = new std::vector<geom::Geometry *>();
```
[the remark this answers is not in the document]
jericks: Done! Except for Points.

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+            coordinates.push_back(geom::Coordinate{coords[i][j].first, coords[i][j].second});
+        }
+        geom::CoordinateArraySequence coordinateSequence { std::move(coordinates) };
+        lines->push_back(geometryFactory.createLineString(std::move(coordinateSequence)));
+    }
+    return std::unique_ptr<geom::MultiLineString>(geometryFactory.createMultiLineString(lines));
+}
+
+std::unique_ptr<geom::MultiPolygon> GeoJSONReader::readMultiPolygon(nlohmann::json& json) {
+    std::vector<std::vector<std::vector<std::pair<double,double>>>> multiPolygonCoords = json["coordinates"].get<std::vector<std::vector<std::vector<std::pair<double,double>>>>>();
```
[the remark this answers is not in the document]
jericks: Done!

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+    std::vector<geom::Geometry *>* polygons = new std::vector<geom::Geometry *>();
+    for(std::vector<int>::size_type i = 0; i < multiPolygonCoords.size(); i++) {
+        std::vector<std::vector<std::pair<double,double>>> polygonCoords = multiPolygonCoords[i];
+        std::vector<geom::LinearRing *> rings;
+        for(std::vector<int>::size_type j = 0; j < polygonCoords.size(); j++) {
+            std::vector<geom::Coordinate> coordinates;
+            for (std::vector<int>::size_type k = 0; k < polygonCoords[j].size(); k++) {
+                coordinates.push_back(geom::Coordinate{polygonCoords[j][k].first, polygonCoords[j][k].second});
+            }
+            geom::CoordinateArraySequence coordinateSequence { std::move(coordinates) };
```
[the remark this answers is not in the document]
jericks: I ended up doing something like this:  

`auto coordinateSequence = geometryFactory.getCoordinateSequenceFactory()->create(std::move(coordinates));`

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+std::unique_ptr<geom::MultiPolygon> GeoJSONReader::readMultiPolygon(nlohmann::json& json) {
+    std::vector<std::vector<std::vector<std::pair<double,double>>>> multiPolygonCoords = json["coordinates"].get<std::vector<std::vector<std::vector<std::pair<double,double>>>>>();
+    std::vector<geom::Geometry *>* polygons = new std::vector<geom::Geometry *>();
+    for(std::vector<int>::size_type i = 0; i < multiPolygonCoords.size(); i++) {
+        std::vector<std::vector<std::pair<double,double>>> polygonCoords = multiPolygonCoords[i];
+        std::vector<geom::LinearRing *> rings;
+        for(std::vector<int>::size_type j = 0; j < polygonCoords.size(); j++) {
+            std::vector<geom::Coordinate> coordinates;
+            for (std::vector<int>::size_type k = 0; k < polygonCoords[j].size(); k++) {
+                coordinates.push_back(geom::Coordinate{polygonCoords[j][k].first, polygonCoords[j][k].second});
```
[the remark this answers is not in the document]
jericks: I converted all of the for loops to range for loops.

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+            polygons->push_back(geometryFactory.createPolygon(outerRing, innerRings));        
+        }
+    }
+    return std::unique_ptr<geom::MultiPolygon>(geometryFactory.createMultiPolygon(polygons));
+}
+
+std::unique_ptr<geom::GeometryCollection> GeoJSONReader::readGeometryCollection(nlohmann::json& j) {
+    std::vector<geom::Geometry *>* geometries = new std::vector<geom::Geometry *>();
+    auto jsonGeometries = j["geometries"];
+    for (auto jsonGeometry : jsonGeometries) {
```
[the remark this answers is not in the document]
jericks: Yikes, thanks for catching this.  I added const & to all for loops.

src/io/GeoJSONReader.cpp
```diff
@@ -0,0 +1,269 @@
...
+    } else if (type == "MultiPoint") {
+        return readMultiPoint(j);
+    } else if (type == "MultiLineString") {
+        return readMultiLineString(j);
+    } else if (type == "MultiPolygon") {
+        return readMultiPolygon(j);
+    } else if (type == "GeometryCollection") {
+        return readGeometryCollection(j);
+    }
+    return std::unique_ptr<geom::Geometry>(geometryFactory.createEmptyGeometry());
```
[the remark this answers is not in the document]
jericks: I am throwing more ParseExceptions and added several tests.

include/geos/io/GeoJSON.h
jericks: What should this be?

src/io/GeoJSON.cpp
jericks: Done!

include/geos/io/GeoJSON.h
dbaston: 2021 Jared Erickson ?

jericks: Done.

src/io/GeoJSONReader.cpp
jericks: I am throwing a ParseException whenever I find a coordinate array with only one value.

src/io/GeoJSON.cpp
jericks: When I refactored GeoJSONValue to use a union, I got rid of this.  I can add it back it you want.

src/io/GeoJSONReader.cpp
jericks: I added unit test 21.

tests/unit/io/GeoJSONReaderTest.cpp
jericks: I am adding unit test 28 for the first case and 29 for second.

tests/unit/io/GeoJSONReaderTest.cpp
jericks: I am adding unit test 27

tests/unit/io/GeoJSONReaderTest.cpp
jericks: I am adding unit test 26

rouault:
>  It uses the nlohmann/json header only library (https://github.com/nlohmann/json) and contains a suite of unit tests.

Could you make sure that the use of nlohmann/json by GEOS is in a dedicated namespace that isn't the default one, so as to avoid potential clashes with other users that could be linked in the same process. Such other user can be recent PROJ versions that also use it, potentially in a different version. In PROJ we do the following
```
// to avoid any clash if PROJ users have another version of nlohmann/json.hpp
#define nlohmann proj_nlohmann
#include "nlohmann/json.hpp"
```

include/geos/io/GeoJSONWriter.h
```diff
@@ -0,0 +1,120 @@
...
+
+    void encodeMultiPolygon(const geom::MultiPolygon* m, nlohmann::ordered_json& j);
+
+    void encodeGeometryCollection(const geom::GeometryCollection* g, nlohmann::ordered_json& j);
+
+    void encodeFeature(const geom::Geometry* g, nlohmann::ordered_json& j);
+
+    void encodeFeatureCollection(const geom::Geometry* g, nlohmann::ordered_json& j);
+
+    void encodeFeature(const GeoJSONFeature feature, nlohmann::ordered_json& j);
```
rouault, line 107:
```suggestion
    void encodeFeature(const GeoJSONFeature& feature, nlohmann::ordered_json& j);
```

include/geos/io/GeoJSONWriter.h
```diff
@@ -0,0 +1,120 @@
...
+
+    void encodeGeoJSONValue(std::string key, GeoJSONValue value, nlohmann::ordered_json& j);
```
rouault, line 109:
```suggestion
    void encodeGeoJSONValue(const std::string& key, const GeoJSONValue& value, nlohmann::ordered_json& j);
```

include/geos/io/GeoJSONWriter.h
```diff
@@ -0,0 +1,120 @@
...
+class GEOS_DLL GeoJSONWriter {
+public:
+    
+    ~GeoJSONWriter() = default;
+
+    std::string write(const geom::Geometry* geometry, GeoJSONType type = GeoJSONType::GEOMETRY);
+
+    std::string writeFormatted(const geom::Geometry* geometry, GeoJSONType type = GeoJSONType::GEOMETRY, int indent = 4);
+
+    std::string write(const GeoJSONFeature feature);
```
rouault, line 75:
```suggestion
    std::string write(const GeoJSONFeature& feature);
```

include/geos/io/GeoJSONWriter.h
```diff
@@ -0,0 +1,120 @@
...
+
+    std::string write(const GeoJSONFeatureCollection features);
```
rouault, line 77:
```suggestion
    std::string write(const GeoJSONFeatureCollection& features);
```

rouault (commented):
A number of use of const& references should improve performance

include/geos/io/GeoJSON.h
```diff
@@ -0,0 +1,135 @@
...
+        GeoJSONValue(bool);
+        GeoJSONValue(const std::map<std::string, GeoJSONValue>&);
+        GeoJSONValue(const std::vector<GeoJSONValue>&);
+
+        ~GeoJSONValue();
+        GeoJSONValue(const GeoJSONValue&);
+        GeoJSONValue& operator=(const GeoJSONValue&);
+
+        double getNumber() const;
+        std::string getString() const;
```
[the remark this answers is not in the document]
jericks: Even, thank you for the code review and suggestions.  I really appreciate your time and help making this pull request better.  I added all of the constant references and attempted to put the json.hpp include in a dedicated namespace.  I am not sure I did that part correctly though.

dbaston:
I think the nlohmann headers should be put somewhere like `include/geos/vend/nlohmann`. I'm pretty sure we don't want to be installing headers outside of the `geos` root. @pramsey ?

pramsey:
ryu ended up in `./src/deps`, that would be OK?

include/geos/io/GeoJSONReader.h
```diff
@@ -22,7 +22,12 @@
 #include <geos/geom/CoordinateSequence.h>
 #include <geos/geom/Geometry.h>
 #include <string>
-#include <json.hpp>
+
+#ifdef GEOS_COMPILATION
```
rouault, line 26: hum, if this header is usable outside of GEOS internals, we can't use a different namespace between GEOS compilation and use of GEOS.
The only viable alternative is:
- don't use a vendorized nlohmann/json, but a system one
- or if using a vendorized nlohmann/json, modify the vendorized header to use a dedicated namespace. Using the #define trick would be dangerous for people using another nlohmann/json and GEOS headers at the same time

hobu: Vendorize and namespace is the way. You can install into ``#include <geos/vend/json.hpp`` or some such. Many projects use nlohmann and are likely to clash if GEOS installed it in anywhere but it's own private spot and privately used it and did not expose it in its public headers.

jericks: Really appreciate the feedback and discussion.  I am a little confused about what I should do.  So, I should move json.hpp to include/geos/vend/json.hpp and then create another file called include_nlohmann_json.hpp with something like:

```
#ifndef GEOS_VEND_JSON
#define GEOS_VEND_JSON
namespace geos {
    namespace vend {
        #include "json.hpp"
    }
}
#endif
```

And then use `#include "geos/vend/include_nlohmann_json.hpp"` in all other header files?

jericks: Or is this enough?

```
#define nlohmann geos_nlohmann
#include "geos/vend/json.hpp"
```

rouault: I don't think what you suggest above would work since json.hpp includes various headers, and that wouldn't be safe to do so in a namespace.

I think include_nlohmann_json.hpp should #undef nlohmann after #include "nlohmann/json.hpp"  , and then you you use explictly the geos_nlohmann namespace.
And for extra safety in case other folks before us would have played with #define nlohmann in a non-transient way:
```
#ifdef nlohmann
  #error "GEOS modifies the nlohmann define "
#endif
#define nlohmann geos_nlohmann
#include "nlohmann/json.hpp" 
#undef nlohmann
```

dbaston: I'm probably missing something obvious, but doesn't this work? (I didn't try it)

```
// normal includes
#include <string>

namespace geos {
#include "nlohmann/json.hpp"

namespace io {

class GeoJSONReader {
 // stuff
};

}
}
```

dbaston: Definitely does not work.

jericks: Changes made.  Thanks again Even and Dan.  I learned something new today!

jorisvandenbossche:
It's nice to see GeoJSON functionality in GEOS! I was just wondering a bit about the goals of this feature. Is it mainly meant for internal use, or also for downstream packages? If the latter, are there plans to add a C API? 
(although that might not be that straightforward, given the several C++ classes that are currently used for reading and writing, GeoJSONFeatureCollection, GeoJSONFeature, GeoJSONValue, etc)

caspervdw:
Agreed with @jorisvandenbossche, it would be extremely useful to have a C API function that transforms a string buffer (GeoJSON) in a vector of GEOSGeometry objects.

PR #423, opened 2021-03-08
Fix crash with GEOSPreparedIntersects/Covers on a POINT EMPTY

rouault:
Fixes use case spotted by https://github.com/OSGeo/gdal/issues/3542#issuecomment-792659437


dbaston:
Backported to 3.9 as 57ca44f2edf16c08b86111d38b252c43b0851718; 3.8 as 63c7f6c42ceee33a5553ca32c3b79fd252adcf73. Issue not present in 3.7.

PR #424, opened 2021-03-08
Add GEOSCoordSeq_copyFromBuffer

dbaston:
Add a method to copy coordinates into a GEOS structure from a contiguous chunk of memory. While this is not as good as using that chunk of memory directly, it greatly simplifies and improves the performance of bringing coordinates into GEOS.

Results of included benchmark show the performance on coordinate sequences of size 10, 1000, and 10 000 relative to existing methods (createByOrdinate: setX/setY and createByCoordinate: setXY)

```
-----------------------------------------------------------------------------------
Benchmark                                         Time             CPU   Iterations
-----------------------------------------------------------------------------------
BM_CoordSeq_CreateByOrdinate<10, 2>             167 ns          167 ns      4181695
BM_CoordSeq_CreateByCoordinate<10, 2>          78.7 ns         78.7 ns      8633210
BM_CoordSeq_CopyFromBuffer<10, 2>              41.3 ns         41.3 ns     17085606
BM_CoordSeq_CreateByOrdinate<10, 3>             224 ns          224 ns      3109464
BM_CoordSeq_CreateByCoordinate<10, 3>          84.6 ns         84.6 ns      8025830
BM_CoordSeq_CopyFromBuffer<10, 3>              39.0 ns         39.0 ns     17938309
BM_CoordSeq_CreateByOrdinate<1000, 2>         12733 ns        12731 ns        54801
BM_CoordSeq_CreateByCoordinate<1000, 2>        4640 ns         4639 ns       151689
BM_CoordSeq_CopyFromBuffer<1000, 2>            1319 ns         1319 ns       517824
BM_CoordSeq_CreateByOrdinate<1000, 3>         19150 ns        19146 ns        36583
BM_CoordSeq_CreateByCoordinate<1000, 3>        5288 ns         5287 ns       127731
BM_CoordSeq_CopyFromBuffer<1000, 3>             926 ns          926 ns       740152
BM_CoordSeq_CreateByOrdinate<10000, 2>       121681 ns       121665 ns         5632
BM_CoordSeq_CreateByCoordinate<10000, 2>      50718 ns        50708 ns        12875
BM_CoordSeq_CopyFromBuffer<10000, 2>          13983 ns        13981 ns        48776
BM_CoordSeq_CreateByOrdinate<10000, 3>       178439 ns       178433 ns         3901
BM_CoordSeq_CreateByCoordinate<10000, 3>      53240 ns        53238 ns        12754
BM_CoordSeq_CopyFromBuffer<10000, 3>          11894 ns        11893 ns        57954
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,8 @@ Changes in 3.10.0
   - geosop CLI for GEOS (Martin Davis)
   - Full doxygen of the C-API (Paul Ramsey)
   - CAPI: GEOSDensify (Brendan Ward)
+  - CAPI: GEOSCoordSeq_copyFromArrays, GEOSCoordSeq_copyFromBuffer,
+          GEOSCoordSeq_copyToArrays, GEOSCoordSeq_copyToBuffer (Daniel Baston)
 
 - Fixes/Improvements:
   - Preserve ordering of lines in overlay results (Martin Davis)
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -8,6 +8,9 @@
 # by the Free Software Foundation.
 # See the COPYING file for more information.
 ################################################################################
+
+find_package(benchmark QUIET)
+
 add_executable(perf_class_sizes ClassSizes.cpp)
 target_link_libraries(perf_class_sizes PRIVATE geos)
 
--- a/benchmarks/algorithm/CMakeLists.txt
+++ b/benchmarks/algorithm/CMakeLists.txt
@@ -19,7 +19,6 @@ target_link_libraries(perf_voronoi geos)
 add_executable(perf_unaryunion_segments UnaryUnionSegmentsPerfTest.cpp)
 target_link_libraries(perf_unaryunion_segments geos)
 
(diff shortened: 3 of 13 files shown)


nyalldawson:
@dbaston lovely stuff -- this is an exciting improvement for downstream clients!!

I'm wondering... would a similar approach be possible for separate buffers of x/y/z arrays? Something like:


    GEOSCoordSeq_copyFromBuffer(const double* x, const double* y, const double* z, unsigned int size )

(where the z buffer is optional and if null then a 2d sequence is implied).

This is how QGIS natively stores rings, so having the ability to ingest in this structure would really minimise the cost of converting QGIS geometry representations to GEOS objects...

dbaston:
Funny, I was _just_ looking to see how QGIS did it. Yes, makes sense to support the alternate representation. And then maybe support the reverse conversion too.

nyalldawson:
> And then maybe support the reverse conversion too.

That would be great! Again, the fastest ring construction method in qgis is from arrays of doubles for each separate dimension, so we could really benefit from that approach!

PR #432, opened 2021-04-15
Fix GEOSisValidReason documentation

### LINKED ISSUE #431: Diff between doc and code for GEOSisValidReason_r
In the documentation, it is said that an empty string will be return for a valid geometry, however, we get `"Valid Geometry"`.

It seems pretty easy to fix, let me know to which version you want to stick (I prefer the empty string) and I'll do it :)

BuonOmo:
Fixes #431


pramsey:
Thanks!

PR #436, opened 2021-04-21
Oss-fuzz integration

catenacyber:
As well as updating tinyxml2
and fixing a first bug cf https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33475

### DIFF
--- a/tests/fuzz/fuzz_geo2.c
+++ b/tests/fuzz/fuzz_geo2.c
@@ -47,7 +47,9 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
     GEOSGeometry *g1 = GEOSGeomFromWKT(Data);
 
     if (g1 != NULL) {
-        GEOSGeometry *g2 = GEOSGeomFromWKB_buf(Data+sep, Size-sep);
+        char * r = malloc(Size-sep);
+        memcpy(r, Data+sep, Size-sep);
+        GEOSGeometry *g2 = GEOSGeomFromWKT(r);
         if (g2 != NULL) {
             size_t usize;
             GEOSGeometry *g3 = GEOSIntersection(g1, g2);
@@ -60,7 +62,6 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
             free(uptr);
             GEOSGeom_destroy(g2);
         }
-        char * r = GEOSGeomToWKT(g1);
         free(r);
         GEOSGeom_destroy(g1);
     }
--- a/tests/fuzz/fuzz_geo2.c
+++ b/tests/fuzz/fuzz_geo2.c
@@ -1,0 +1,69 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdarg.h>
+#include <string.h>
+
+#include "geos_c.h"
+
+static int initialized = 0;
+FILE * flogOut;
+
+void
+notice(const char *fmt, ...) {
+    va_list ap;
+    fprintf( flogOut, "NOTICE: ");
+    va_start (ap, fmt);
+    vfprintf( flogOut, fmt, ap);
+    va_end(ap);
+    fprintf( flogOut, "\n" );
+}
+
+void
+log_and_exit(const char *fmt, ...) {
+    va_list ap;
+    fprintf( flogOut, "ERROR: ");
+    va_start (ap, fmt);
+    vfprintf( flogOut, fmt, ap);
+    va_end(ap);
+    fprintf( flogOut, "\n" );
+}
+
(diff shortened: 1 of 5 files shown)


tests/fuzz/fuzz_geo2.c
```diff
@@ -0,0 +1,69 @@
...
+log_and_exit(const char *fmt, ...) {
+    va_list ap;
+    fprintf( flogOut, "ERROR: ");
+    va_start (ap, fmt);
+    vfprintf( flogOut, fmt, ap);
+    va_end(ap);
+    fprintf( flogOut, "\n" );
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
```
pramsey, line 32: I'm not sure I understand this test.
* Get a random hunk of binary data
* Try to parse it as WKT?
* If success... try to parse it as WKB?
* If success (how?) ... try to intersect/difference/union it with itself (bearing in mind those functions require valid inputs)
I guess I can see throwing garbage geometry at intersection/difference/union on the theory that they shouldn't crash ever, even if they might return nonsense. What I don't see is how this attempt to use the input data as both WKB and WKT works.

catenacyber: > Try to parse it as WKT?
If success... try to parse it as WKB?

This is not the same data.

cf `GEOSGeomFromWKT(Data);` vs  `GEOSGeomFromWKB_buf(Data+sep, Size-sep);` where sep is the index of the first null character

Is this what you were missing ?

pramsey: Yes, kind of. But also, what is generating the inputs? Is there, like, a "fuzzy geometry generator"? If they are independent random inputs the odds of them ever being both valid at once seem very low. I guess I'm missing something important here.

catenacyber: > Yes, kind of. But also, what is generating the inputs?

They are called the fuzzing engines.
They are libfuzzer, afl, honggfuzz.

> Is there, like, a "fuzzy geometry generator"? If they are independent random inputs the odds of them ever being both valid at once seem very low. I guess I'm missing something important here.

They are not independent random inputs.
The big secret is the coverage as a feedback.
You can search for tutorials about it

Komzpa: For the sake of sanity when trying to debug the oss-fuzz result, please keep both inputs as WKT. 
At the end of the day, oss-fuzz will find something and a human will have to look at what's going on there.

It's ok to have more fuzzing targets for different kinds of inputs, just like PostGIS does.

catenacyber: > please keep both inputs as WKT.

done

robe2:
This has been here for a while, are we planning to commit this or does it need more changes? @pramsey  @dbaston

pramsey:
I am personally unexcited about it, as every time I try to address a ossfuzz ticket I burn a solid day trying to figure out how to extract the input data from the reporting format. So, I dunno, I'd say whomever attaches us to this system should be willing to file tickets with actual test cases attached so the cognitive overhead of dealing with ossfuzz is taken out of the equation.

catenacyber:
Ok, I can close this.

Out of curiosity, what is the difficulty for reproducing ?
Does https://google.github.io/oss-fuzz/advanced-topics/reproducing/ help ?
Do you have trouble logging to get the reproducer case ? building the fuzz targets ? Or converting the reproducer test case in your preferred format for input ?

pramsey:
For me it was never super straightforward to go from "ticket in OSSFuzz" to "test case in the debugger". If someone else wants to do that, power to them!

catenacyber:
Ok, thanks for the feedback.
Did you try the oss-fuzz python wrapper scripts around Docker ?
(The point of using a docker container is I think, to have everything set, and not need additional stuff, which can indeed be complex when you try to reproduce on a local machine)

Komzpa:
I think part of the reason why might be as @pramsey is not using the gmail account for the oss-fuzz notifications so it may be not really showing the full UI. It was mentioned as a weird side note somewhere I think. It was pretty straightforward to me - get the reproducer file, feed it into the test suite file (I think another problem in PostGIS is that it's not built when running `make` so you have to run another `make` somewhere else instead of plain `cat crasher_file | logbt ./fuzzer/wkb_import_fuzzer`).

I think we can invest some of Kontur's time into getting this piece better, maybe we need to just polish the infra a bit and write a five-step-doc quickstart guide for @pramsey? Let me know if that's a good idea.

pramsey:
Could be, but at @catenacyber notes the actual environment is fiddly to set up, and providing "easy simple" instructions over a host of potential platforms might be harder than one likes. 

I didn't try the docker solution, but I wonder to what extent one could attach a debugger to it? That's kind of what it comes down to, being able to run the test program, with the bad inputs, with a debugger, against a local build of the library.

PR #439, opened 2021-04-26
Fix DPSimplifier collapsed ring removal

dr-jts:
Fixes issue in [GEOS 1115](https://trac.osgeo.org/geos/ticket/1115).


dr-jts:
Fixed by [9f3dfbdc](https://github.com/libgeos/geos/commit/f3dfbdc949dbe954c818e63abceb4e493ab396ff)

PR #442, opened 2021-04-28
remove memory leak in BufferOp

averello:
When instance variable resultGeometry was computed by a call to
getResultGeometry() it was never freed.

This change adds a destructor to BufferOp and destroys resultGeometry if
it was computed.

### DIFF
--- a/include/geos/operation/buffer/BufferOp.h
+++ b/include/geos/operation/buffer/BufferOp.h
@@ -192,6 +192,8 @@ public:
     {
     }
 
+    ~BufferOp();
+
     /** \brief
      * Specifies the end cap style of the generated buffer.
      *
--- a/src/operation/buffer/BufferOp.cpp
+++ b/src/operation/buffer/BufferOp.cpp
@@ -78,6 +78,13 @@ OLDprecisionScaleFactor(const Geometry* g,
 }
 #endif
 
+BufferOp::~BufferOp(void)
+{
+    if (resultGeometry != nullptr) {
+        argGeom->getFactory()->destroyGeometry(resultGeometry);
+    }
+}
+
 /*private*/
 double
 BufferOp::precisionScaleFactor(const Geometry* g,


pramsey:
Have you run 'make check' on this? It's a pile of segfaults for me.

averello:
Yes it's a mess. I acted too fast to request this PR.
No I did not run make check. 

I see a possible leak in BufferOp that is probably hidden by the way it is used by the rest of the codebase.

averello:
At least with this minimal example I have a leak in my code base:

```c++
std::unique_ptr<geos::geom::Geometry> envelope = ...;
geos::geom::BufferParameters bufferParams;
geos::geom::BufferOp bufferOperator(envelope.get(), bufferParams);
auto const &buffer = bufferOperator.getResultGeometry(Distance);
// use buffer
// get out of scope
```

`buffer` is a direct pointer to the `bufferOperator`'s instance `resultGeometry`.

As I see it, it is the responsibility of `BufferOp` to free this instance.

It is not documented in the `getResultGeometry` that the ownership passes to the caller.

pramsey:
For all released versions of GEOS (<= 3.9), a documentation patch is the most we can do, we cannot change the life-cycle contract on BufferOp in a patch release.

For main, the "correct" thing to do would be to modernize the whole thing a bit so it's returning a std::unique_ptr<Geometry> that way it's all quite explicit who owns what. The whole "don't double compute" thing that having the resultGeometry lying around is maybe for? That would have to be checked to see that it in fact is not actually something that happens, at least within our code base (I cannot imagine it is).

dr-jts:
> It is not documented in the `geos::geom::BufferOp::getResultGeometry()` that the ownership passes to the caller.

It is the intent that the caller owns the return value of `geos::geom::BufferOp::getResultGeometry()`. So this should be documented.  And as Paul says, modernized to use `std::unique_ptr`.

averello:
So, first I could change this PR to document the intended ownership transfer.
On a later PR modernize the API to use std::unique_ptr

Sounds good?

PR #444, opened 2021-05-01
Avoid undefined behavior in casts in Densify(), DiscreteHausdorffDistance and DiscreteFrechetDistance

rouault:
Those pieces of code do things like (integer_type)potential_big_double_value,
which is undefined behavior if the potential_big_double_value is beyond
the validity range of integer_type. Catch that before the cast.


pramsey:
Thanks!

PR #445, opened 2021-05-03
Install geosop using option BUILD_GEOSOP

mwtoews:
I was expecting to find this new tool in `bin`, but this PR should fix that.

### DIFF
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -9,4 +9,9 @@
 # See the COPYING file for more information.
 ################################################################################
 
-add_subdirectory(geosop)
+option(BUILD_GEOSOP "Build geosop command-line interface" ON)
+
+message(STATUS "GEOS: Build geosop ${BUILD_GEOSOP}")
+if(BUILD_GEOSOP)
+  add_subdirectory(geosop)
+endif()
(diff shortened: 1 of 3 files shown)


mwtoews:
A second commit here adds an option to not build/install geosop, which might be wanted by folks that only need the library. E.g.:

    cmake -DBUILD_GEOSOP=OFF ..

will disable the build/install of geosop to /path/to/bin.

Also note that I've verified that `make uninstall` works as expected, with or without this option.

util/CMakeLists.txt
```diff
@@ -9,4 +9,8 @@
 # See the COPYING file for more information.
 ################################################################################
 
-add_subdirectory(geosop)
+option(BUILD_GEOSOP "Build geosop command-line interface" ON)
```
dr-jts, line 12: How about calling this `GEOS_BUILD_GEOSOP`, so that GEOS options are grouped together?

(And we should really change a couple of the others, like `DISABLE_OVERLAYNG` ==> `GEOS_DISABLE_OVERLAYNG`)

dr-jts: I guess there is also an existing pattern for `BUILD_...`.  So could continue with that.

mwtoews: yeah, I know what you mean and did briefly consider the two conventions.

mwtoews: If it helps, [PROJ's CMake options](https://proj.org/install.html#cmake-configure-options) use a <tt>BUILD_*NAME*</tt> convention for each tool, which is what swayed my decision here. The main reason why `GEOS_`  is prefixed to many CMake variables is to avoid clashes with other variables when creating large CMake projects. No other tool is named geosop, so no clash.

mwtoews:
I've merged the previous commits and made a few subtle changes, including to show a message for this option, and to remove unnecessary "RUNTIME" from the `install` command, which should make [`CMP0006 NEW`](https://cmake.org/cmake/help/latest/policy/CMP0006.html) policy behaviour happier. (An alternative would be to specify both "RUNTIME DESTINATION" and "BUNDLE DESTINATION").

It seems there are a few unrelated CI failures.

pramsey:
Thanks! Merged.

PR #446, opened 2021-05-04
Fix memory leaks

rouault:
The first commit is a fix in the library itself. Might be backported. Alongside the second commit that is in the test suite.
The third commit is just for master.


pramsey:
Thanks!

PR #450, opened 2021-05-27
Add GeoJSON reader to C API

caspervdw:
Following up on the work done in #419 , I added basic read functionality to the C API. It follows the API of the WKB and WKT readers and reads GeoJSON into a single geometry. Features and properties are ignored.

@jorisvandenbossche

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -39,6 +39,7 @@
 #define GEOSWKTWriter geos::io::WKTWriter
 #define GEOSWKBReader geos::io::WKBReader
 #define GEOSWKBWriter geos::io::WKBWriter
+#define GEOSGeoJSONReader geos::io::GeoJSONReader
 typedef struct GEOSBufParams_t GEOSBufferParams;
 typedef struct GEOSMakeValidParams_t GEOSMakeValidParams;
 
@@ -70,6 +71,7 @@ using geos::io::WKTReader;
 using geos::io::WKTWriter;
 using geos::io::WKBReader;
 using geos::io::WKBWriter;
+using geos::io::GeoJSONReader;
 
 
 
@@ -1286,6 +1288,25 @@ extern "C" {
     GEOSWKBWriter_setIncludeSRID(GEOSWKBWriter* writer, const char newIncludeSRID)
     {
         GEOSWKBWriter_setIncludeSRID_r(handle, writer, newIncludeSRID);
+    }
+
+    /* GeoJSON Reader */
+    GeoJSONReader*
+    GEOSGeoJSONReader_create()
+    {
+        return GEOSGeoJSONReader_create_r(handle);
+    }
+
+    void
+    GEOSGeoJSONReader_destroy(GeoJSONReader* reader)
+    {
+        GEOSGeoJSONReader_destroy_r(handle, reader);
+    }
+
+    Geometry*
+    GEOSGeoJSONReader_read(GeoJSONReader* reader, const char* geojson)
+    {
+        return GEOSGeoJSONReader_read_r(handle, reader, geojson);
     }
 
 
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -21,6 +21,7 @@
 #include <geos/io/WKBReader.h>
 #include <geos/io/WKTWriter.h>
 #include <geos/io/WKBWriter.h>
+#include <geos/io/GeoJSONReader.h>
 #include <geos/util/Interrupt.h>
 
 #include <stdexcept>
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -1585,6 +1585,13 @@ typedef struct GEOSWKBReader_t GEOSWKBReader;
 */
 typedef struct GEOSWKBWriter_t GEOSWKBWriter;
 
+/**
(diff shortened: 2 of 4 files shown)


caspervdw:
@pramsey I am not sure if you want to include this in the C API. If so, I can resume working on the writer functions.

The concept of `Feature` and `FeatureCollection` in the C API requires a bit more thinking. Would these fit in the scope of the GEOS C API?

pramsey:
You can go as far as you wish, but having it in the CAPI will certainly enlarge the population of potential users. 

I am a little leery of Feature, since all of a sudden we have a whole type model, attribute stuff and so on, most of which is orthogonal to geometry processing, they are just coming along for the ride. I'd lean towards saying "no". I recognize the difficulty here with JSON, since any GeoJSON processor would probably want to parse things at the Feature or FeatureCollection level, and if we don't do that they won't delegate the Geometry level to us, they'll just parse it themselves and construct GEOS geometry by hand.

caspervdw:
The feature properties indeed seem out of scope of GEOS. There is however a reserved property name “id” which according to the spec _should_ be used as a unique identifier. GEOS could provide that like “_readWithId”. It can be both int an string.

Not sure about this. Maybe it is good to do feature properties well or not do it at all.

I’ll leave it out of scope in this PR and focus on pure geometry IO.

caspervdw:
I added the writer functions. Please let me know what you think about the API (one function, with `type` and `indent` arguments).

There is one issue that seems impossible to solve without touching the C++. When reading a FeatureCollection with ``GeoJSONReader->read``, you obtain a ``GEOSGeometry`` with type GeometryCollection. However, when writing the same GeometryCollection using ``GeoJSONWriter->write(g, FEATURE_COLLECTION)``, you get a FeatureCollection with only one geometry, a geometrycollection. See the unittest over here:

https://github.com/libgeos/geos/pull/450/files#diff-01a4c7854f11540c77b4b77026a4b5168ad0746efb98c03ac4151a69e57360ebR75

I expected that ``GeoJSONWriter->write(g, FEATURE_COLLECTION)`` would only accept collections, and turn each element in a collection into a feature. Currently it seems impossible to write feature collections without introducing the types `Feature` and `FeatureCollection`, which is something I'd like to avoid.

What to do? Change the ``encodeFeatureCollection`` ([source](https://github.com/libgeos/geos/blob/43587c30c4d67bd450d75e1d40bb61a23f11533f/src/io/GeoJSONWriter.cpp#L146)) to unpack collections and to error when simple geometries are provided? Or introduce a public ``GeoJSONWriter->writeCollection`` or something like that?

tests/unit/capi/GEOSGeomGeoJSONWriteTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+template<>
+template<>
+void object::test<3>
+()
+{
+    std::string wkt("GEOMETRYCOLLECTION (POINT (-117.000 33.000), POINT (-122.000 45.000))");
+    std::string expected("{\"type\":\"FeatureCollection\",\"features\":[{\"type\":\"Feature\",\"geometry\":{\"type\":\"Point\",\"coordinates\":[-117.0,33.0]}},{\"type\":\"Feature\",\"geometry\":{\"type\":\"Point\",\"coordinates\":[-122.0,45.0]}}]}");
+    test_geojson(wkt, expected, GEOSGEOJSON_FEATURE_COLLECTION, -1);
```
caspervdw, line 77: Getting
```
---> group: capi::GEOSGeomGeoJSONWrite, test: test<3>
     problem: assertion failed
     failed assertion: `Values are not equal: expected `{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Point","coordinates":[-117.0,33.0]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-122.0,45.0]}}]}` actual `{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[-117.0,33.0]},{"type":"Point","coordinates":[-122.0,45.0]}]}}]}``
```

caspervdw:
@pramsey  I want to have another take at this, but I prefer to have some green light on the target API.

My current plan is doing:

```
- `GEOSGeoJSONReader_readGeometry`: reads any geojson (geometry, feature, featurecollection) into a single Geometry. If a collection is returned, there is no way to tell if it came from a single geojson geometry/feature or from a featurecollection.
- `GEOSGeoJSONWriter_writeGeometry`: writes any geometry into a geojson geometry object
- `GEOSGeoJSONWriter_writeFeatureCollection`: writes only collections into a geojson featurecollection object (errors on simple geoms)
```

Contrary what I said earlier, I believe this can be done without touching the C++

pramsey:
Still unclear on what the FeatureCollection work is for? If we agree that FC is out of scope, why an option to write them? I'm OK with the CAPI function names as described.

caspervdw:
Agreed, writing featurecollections without properties doesn't add anything. I dropped the proposed `GEOSGeoJSONWriter_writeFeatureCollection` and finished the PR.

caspervdw:
@pramsey This PR is finished; please tell me if you need anything from me before merging. I’d like to start implementing this in pygeos and see what are the improvements in geojson I/O. I have high expectations especially for reading largish files.

PR #451, opened 2021-06-02
Correctly round-trip collections of EMPTY

pramsey:
As described in https://trac.osgeo.org/geos/ticket/1040


mwtoews (approved):
Changes look good, and results are now as expected from shapely too.

PR #455, opened 2021-06-16
SegmentNodeList::getSplitCoordinates(): fix quadratic performance pattern

rouault:
getSplitCoordinates() calls addEdgeCoordinates() repeatdly, which itself
calls std::unique() on the whole coordList, thus iterating from the
start each time, which is a quadratic performance pattern. As
addEdgeCoordinates() is a private method only called by
getSplitCoordinates(), move the call to std::unique() to the end of
getSplitCoordinates() itself.

Issue was detected when observing long runtimes on safe-TestBufferJagged.xml

Before this PR:
$ /home/even/geos/build_cmake/bin/test_xmltester -v /home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test1: buffer(a, 0.35): ok. (11 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test2: buffer(a, 0.75): ok. (155 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test3: buffer(a, 1.01): ok. (209 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test4: buffer(a, 1.1): ok. (210 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test5: buffer(a, 1.5): ok. (310 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test6: buffer(a, 2): ok. (429 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test7: buffer(a, 5): ok. (1326 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test8: buffer(a, 10): ok. (2497 ms)

After:
$ /home/even/geos/build_cmake/bin/test_xmltester -v /home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test1: buffer(a, 0.35): ok. (11 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test2: buffer(a, 0.75): ok. (65 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test3: buffer(a, 1.01): ok. (79 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test4: buffer(a, 1.1): ok. (78 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test5: buffer(a, 1.5): ok. (105 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test6: buffer(a, 2): ok. (133 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test7: buffer(a, 5): ok. (302 ms)
/home/even/geos/tests/xmltester/tests/misc/safe-TestBufferJagged.xml: case1: test8: buffer(a, 10): ok. (464 ms)
Files: 1
Tests: 8
Failed: 0
Succeeded: 8

### DIFF
--- a/src/noding/SegmentNodeList.cpp
+++ b/src/noding/SegmentNodeList.cpp
@@ -301,6 +301,8 @@ SegmentNodeList::getSplitCoordinates()
         addEdgeCoordinates(eiPrev, ei, coordList);
         eiPrev = ei;
     }
+    // Remove duplicate Coordinates from coordList
+    coordList.erase(std::unique(coordList.begin(), coordList.end()), coordList.end());
     return coordList;
 }
 
@@ -311,8 +313,6 @@ SegmentNodeList::addEdgeCoordinates(const SegmentNode* ei0, const SegmentNode* e
     auto pts = createSplitEdgePts(ei0, ei1);
     // Append pts to coordList
     pts->toVector(coordList);
-    // Remove duplicate Coordinates from coordList
-    coordList.erase(std::unique(coordList.begin(), coordList.end()), coordList.end());
 }
 
 


dr-jts:
Good find.

I note that even with this change the GEOS code is [different](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/noding/SegmentNodeList.java#L281-L289) to JTS.  JTS uses the `CoordinateList` class, which prevents duplicate coordinates from being inserted (as opposed to removing them afterwards).   Would this approach be any faster than using `coordList.erase(std::unique(...))`?  If so, perhaps it's worth using the GEOS `CoordinateList` here.

rouault:
> Would this approach be any faster than using coordList.erase(std::unique(...))? 

I anticipate it to be marginally / not measurably faster. I can give it a try though.

>  If so, perhaps it's worth using the GEOS `CoordinateList` here.

Side note: I think the port of CoordinateList to C++ isn't ideal by using a std::list wherea JTS use an ArrayList. In C++, a std::list<> is really a double-linked list, wheras ArrayList is mostly a vector. I read an article recently whose point was: you should hardly ever use std::list<>

rouault:
> I anticipate it to be marginally / not measurably faster. I can give it a try though.

My anticipation was right. I've added a commit that checks if points are duplicates before adding. The timings are exactly the same as with the first commit

pramsey:
If it's not faster, but it's way more invasive... ?

rouault:
> If it's not faster, but it's way more invasive... ?

I'm happy to only keep the first commit

Komzpa:
I confirm that in the slow buffer profiles I have on hand I see a lot of geos::noding::SegmentNodeList::addEdgeCoordinates. It also feels like it leaks like hell (32 processes of Postgres manage to get 700GB of RAM consumed even though input is guaranteed <1GB), but I have no good proofs.

dr-jts:
> > If it's not faster, but it's way more invasive... ?
> 
> I'm happy to only keep the first commit

+1 from me.  Thanks for checking.

dr-jts:
> Side note: I think the port of CoordinateList to C++ isn't ideal by using a std::list wherea JTS use an ArrayList. In C++, a std::list<> is really a double-linked list, wheras ArrayList is mostly a vector. I read an article recently whose point was: you should hardly ever use std::list<>

Good point!  That should definitely be changed.  Not sure how much GEOS `CoordinateList` is used - in JTS it's used quite a bit.

rouault:
> > I'm happy to only keep the first commit
> 
> +1 from me. Thanks for checking.

Second commit dropped

pramsey:
Also back-ported into 3.9

PR #458, opened 2021-06-17
Fix GEOSSTRtree_query_r to be thread-safe on first call (lock TemplateSTRtreeImpl on build)

jorisvandenbossche:
This attempts to fix the problem that while querying the STRtree itself is thread-safe, building the tree is not, and building is only done on the first query, making the query also unsafe.

I ran into this while testing a parallel+partitioned spatial join (using the Dask, a python framework for parallel/distributed computing), where I end up querying the same tree from multiple threads at the same time. But I noticed occasional segfaults with this approach (https://github.com/pygeos/pygeos/issues/361). Doing a dummy query in advance to ensure the tree is already built before it's queried from multiple threads fixes it on the PyGEOS side, but adding a lock guard on the `build()` method also independently fixes the root problem in GEOS itself.

Note:

- I needed to add a custom copy constructor and copy assignment (because the `std::mutex lock_` is not copyable), but no idea if this is implemented properly.
- Not directly sure how to test this here

### DIFF
--- a/include/geos/index/strtree/TemplateSTRtree.h
+++ b/include/geos/index/strtree/TemplateSTRtree.h
@@ -27,6 +27,7 @@
 
 #include <vector>
 #include <queue>
+#include <mutex>
 
 namespace geos {
 namespace index {
@@ -143,6 +144,25 @@ public:
         numItems(0) {
         auto finalSize = treeSize(itemCapacity);
         nodes.reserve(finalSize);
+    }
+
+    /**
+     * Copy constructor, needed because mutex is not copyable
+     */
+    TemplateSTRtreeImpl(const TemplateSTRtreeImpl& other) :
+        root(other.root),
+        nodeCapacity(other.nodeCapacity),
+        numItems(other.numItems) {
+        nodes = other.nodes;
+    }
+
+    TemplateSTRtreeImpl& operator=(TemplateSTRtreeImpl other)
+    {
+        root = other.root;
+        nodeCapacity = other.nodeCapacity;
+        numItems = other.numItems;
+        nodes = other.nodes;
+        return *this;
     }
 
     /// @}
@@ -320,6 +340,8 @@ public:
 
     /** Build the tree if it has not already been built. */
     void build() {
+        std::lock_guard<std::mutex> lock(lock_);
+
         if (built()) {
             return;
         }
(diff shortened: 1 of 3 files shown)


include/geos/index/strtree/TemplateSTRtree.h
```diff
@@ -145,6 +146,25 @@ class TemplateSTRtreeImpl {
         nodes.reserve(finalSize);
     }
 
+    /**
+     * Copy constructor, needed because mutex is not copyable
+     */
+    TemplateSTRtreeImpl(const TemplateSTRtreeImpl& other) :
```
jorisvandenbossche, line 152: As mentioned in the top post, I needed to add a copy constructor because the default one doesn't work because the std::mutex lock_ is not copyable, but no idea if this is a sensible implementation.

dbaston: I wouldn't expect a tree to be copied, except possibly as a member variable somewhere else that is initialized later. Another alternative might be to remove the use of the copy constructor.

jorisvandenbossche: The compilation error pointed to 

https://github.com/libgeos/geos/blob/fbaaba88c18cacd4de4263680c9d071a225590ce/src/algorithm/locate/IndexedPointInAreaLocator.cpp#L57

jorisvandenbossche: I suppose this copy assignment can be avoided with a different way to construct the index here? (also `TemplateSTRtreeTest.cpp` would need to be updated)

jorisvandenbossche:
@dbaston thanks a lot for adding the test! (and for the review)

pramsey (approved):

tests/unit/capi/GEOSSTRtreeTest.cpp
```diff
@@ -339,6 +340,47 @@ void object::test<10>()
...
+            (void) g;
+            (void) userdata;
+        }, nullptr);
+    };
+
+    GEOSGeometry* q1 = GEOSBuffer(geoms[10], 10, 48);
+    GEOSGeometry* q2 = GEOSBuffer(geoms[40], 10, 48);
+
+    std::thread t1(query, q1);
+    std::thread t2(query, q1);
```
pramsey, line 369: `std::thread t2(query, q2);` ?

dbaston: Sure.

pramsey:
Merged at 917aeecad

PR #459, opened 2021-06-24
Fix Accidentally Enabled HeuristicOveraly Debugging

jbronn:
Fixes [#1116](https://trac.osgeo.org/geos/ticket/1116); `#ifdef` statements were used on a macro that was already defined; correct to use `#if` like the rest of the file.


pramsey:
Thanks!

Issue #460, opened 2021-06-28
GEOS compilation fails with Intel compiler


adamjstewart:
I'm trying to compile GEOS 3.9.1 on CentOS 7 with Intel 19.1.1.217 and I'm getting the following compilation error:
```
ASLocalizer.cpp(411): error #913: invalid multibyte character sequence
  	addPair("Formatted  %s\n", L"Форматиран  %s\n");		// should align with unchanged
  	                             ^
```
Here is the full [build log](https://github.com/libgeos/geos/files/6723049/spack-build-out.txt).

Any suggestions for how to resolve this issue?

pramsey:
You could strip out those lines, or even just remove astyle from the cmakelists.txt file entirely. It's only used for code formatting so it's not necessary at all. I have no idea why icc chokes on that character string.

dbaston:
I wonder if it may be an issue with the build environment or compiler flags, e.g. https://github.com/spack/spack/pull/14450 or https://community.intel.com/t5/Intel-C-Compiler/improper-UTF-8-characters-handling-by-ICC-on-Windows/td-p/1099463

Adding a `BUILD_ASTYLE` option to the CMake configuration probably makes sense.

mwtoews:
> Adding a `BUILD_ASTYLE` option to the CMake configuration probably makes sense.

Agree, see [gh-461](https://github.com/libgeos/geos/pull/461)

adamjstewart:
I switched from the Autotools build to the CMake build and the problem magically went away!

adamjstewart:
Nevermind, I was using GCC that time. Trying again with Intel...

mwtoews:
For what it's worth, I'm unable to reproduce the error using the free Intel oneAPI:
```
mkdir build && cd build
cmake -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc ..
make
ctest
```
this compiles fine (no errors, no warnings), but fails some of the tests:
```
The following tests FAILED:
	 12 - unit-algorithm-LocatePointInRing (Failed)
	 15 - unit-algorithm-OrientationIndexFailure (Failed)
	 18 - unit-algorithm-RobustLineIntersector (Failed)
	129 - unit-geom-Triangle (Failed)
	152 - unit-math-DD (Failed)
	166 - unit-operation-distance-DistanceOp (Failed)
	211 - unit-triangulate-Voronoi (Failed)
	244 - general-TestOverlayPLPrec (Failed)
[...]

PR #461, opened 2021-06-28
CMake: add option BUILD_ASTYLE, with default ON

mwtoews:
Closes [gh-460](#460) (should there be a Trac ticket?)

There was a bit of discussion in 7e8575f4c8672fb51d28601f82ec724477220084 regarding the CMake name format for these options.


mwtoews:
An alternative idea is to change the default to `OFF`, as it does not seem to be a required tool.

PR #465, opened 2021-07-05
Remove support/use of Ubuntu 16.04 (xenial)

mwtoews:
It seems there were a few remaining reliance on Ubuntu 16.04 (xenial), which is long out of support.

This PR fixes the azp ubuntu-18.04 vmImage to use bionic instead of xenial, which is one of the recent CI failures.

It also drops a job from `.github/workflows/ci.yml` that uses ubuntu-16.04 with clang.

Further, it simplifies `tools/cppcheck.sh` by removing a workaround for cppcheck-1.72 distributed with ubuntu-16.04.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -107,7 +107,7 @@ stages:
         sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
         if test -n "${LLVM_REPO}" ; then
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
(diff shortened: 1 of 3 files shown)


mwtoews:
It seems that some older Clang versions are not provided for ubuntu-18.04. Should some of these tests be dropped? Or is there a better package source?

pramsey:
You seem to be retracing steps I already covered in my isvalid branch... could you hold up for a day or so? I already made a number of CI related changes in there (I know, should have done a separate branch, sorry)

mwtoews:
Cool, I'll leave you to it! From what I've seen ([eg](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=bionic)) several older compilers should now be retired, as they are simply not found in the wild anymore.

PR #466, opened 2021-07-11
Modify GeoJSON module

whuaegeanse:
Modify codes of GeoJSON module.
- Make use of right value reference for temp variables;
- Reserve memory of std::vector.

### DIFF
--- a/src/io/GeoJSONReader.cpp
+++ b/src/io/GeoJSONReader.cpp
@@ -45,94 +45,112 @@ GeoJSONReader::GeoJSONReader(): GeoJSONReader(*(GeometryFactory::getDefaultInsta
 
 GeoJSONReader::GeoJSONReader(const geom::GeometryFactory& gf) : geometryFactory(gf) {}
 
-std::unique_ptr<geom::Geometry> GeoJSONReader::read(const std::string& geoJsonText) const {
+std::unique_ptr<geom::Geometry> GeoJSONReader::read(const std::string& geoJsonText) const
+{
     try {
         const json& j = json::parse(geoJsonText);
         const std::string& type = j["type"];
         if (type == "Feature") {
             return readFeatureForGeometry(j);
-        } else if (type == "FeatureCollection") {
+        }
+        else if (type == "FeatureCollection") {
             return readFeatureCollectionForGeometry(j);
-        } else {
+        }
+        else {
             return readGeometry(j);
         }
-    } catch (json::exception& ex) {
+    }
+    catch (json::exception& ex) {
         throw ParseException("Error parsing JSON", ex.what());
     }
 }
 
-GeoJSONFeatureCollection GeoJSONReader::readFeatures(const std::string& geoJsonText) const {
+GeoJSONFeatureCollection GeoJSONReader::readFeatures(const std::string& geoJsonText) const
+{
     try {
         const json& j = json::parse(geoJsonText);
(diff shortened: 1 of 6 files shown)


pramsey:
Can you do this w/o all the formatting changes? Your actual changes are hard to locate in the PR.

whuaegeanse:
All of changes in this PR have been formatted with AStyle of GEOS.
Some files in this PR does not change after using AStyle of GEOS.

dbaston:
@pramsey see https://github.com/libgeos/geos/pull/466/files?w=1

whuaegeanse:
> Can you do this w/o all the formatting changes? Your actual changes are hard to locate in the PR.

After searching on google, I learned how to use AStyle to format files on the windows system. Maybe we explain how to use AStyle formatting files under windows system, just like providing instructions for use under linux system.

pramsey:
Well, this situation is complicated by the fact that the original author did not astyle their commit, so you astyling on top of his code doesn't result in a nice clean "changes only" result, it's still very noisy. I will just go ahead and try and review through the static.

src/io/GeoJSONReader.cpp
```diff
@@ -45,210 +45,268 @@ GeoJSONReader::GeoJSONReader(): GeoJSONReader(*(GeometryFactory::getDefaultInsta
...
 }
 
-std::unique_ptr<geom::Polygon> GeoJSONReader::readPolygon(const std::vector<std::vector<std::vector<double>>>& polygonCoords) {
+std::unique_ptr<geom::Polygon> GeoJSONReader::readPolygon(
+    const std::vector<std::vector<std::vector<double>>>& polygonCoords) const
+{
     std::unique_ptr<geom::LinearRing> shell;
     std::vector<std::unique_ptr<geom::LinearRing>> rings;
-    for(const auto& ring : polygonCoords) {
+    rings.reserve(polygonCoords.size());
```
pramsey, line 259: Do we have tests for the various cases where the ordinate arrays can be zero-length? Is reserving 0 acceptable behaviour?

src/io/GeoJSONWriter.cpp
```diff
@@ -39,111 +39,132 @@ using json = geos_nlohmann::ordered_json;
...
     return geojson;
 }
 
-std::string GeoJSONWriter::writeFormatted(const geom::Geometry* geometry, GeoJSONType type, int indent) {
+std::string GeoJSONWriter::writeFormatted(const geom::Geometry* geometry, GeoJSONType type, int indent)
+{
     json j;
     encode(geometry, type, j);
-    std::string geojson = j.dump(indent);
+    const std::string& geojson = j.dump(indent);
```
pramsey, line 54: Just for my curiousity (I'm no cxx expert) is this really saving a copy (assigning to a const reference before returning) or is it just making us feel good about apparently saving a copy?

src/io/GeoJSONReader.cpp
whuaegeanse: > Do we have tests for the various cases where the ordinate arrays can be zero-length? Is reserving 0 acceptable behaviour?

Only the new capacity is greater than the existing capacity will trigger memory allocation.

src/io/GeoJSONWriter.cpp
whuaegeanse: I have made a new  PR #467  to fix bug and enhance functions.

PR #468, opened 2021-07-16
OpenMP Parallelization for CascadedUnion [WIP]

pramsey:
Experiment in using multi-threading to speed up computationally expensive processes. Works "as advertised" but best way to support build on multiple platforms and to expose hooks so that controlling systems (PgSQL) can throttle the number of threads to a set maximum is still TBD.


src/operation/union/CascadedPolygonUnion.cpp
```diff
@@ -94,13 +97,58 @@ CascadedPolygonUnion::Union()
...
+#endif
 
+/* private */
 std::unique_ptr<geom::Geometry>
-CascadedPolygonUnion::binaryUnion(const std::vector<const geom::Geometry*> & geoms,
-                                  std::size_t start, std::size_t end)
+CascadedPolygonUnion::pairUnion(
+    std::vector<const geom::Geometry*>& inply) const
+{
+    int sz = (int)(inply.size());
```
rouault, line 116: if you used size_t here and in other places, I guess you could avoid the explicit casts below

src/operation/union/CascadedPolygonUnion.cpp
```diff
@@ -94,13 +97,58 @@ CascadedPolygonUnion::Union()
...
+    int nsz = (sz / 2) + (sz % 2);
+    std::vector<std::unique_ptr<geom::Geometry>> outply((std::size_t)nsz);
+
+#pragma omp parallel for
+    for (int i = 0; i < sz; i += 2) {
+        /* Work backwards to preserve an old regression result */
```
rouault, line 122: Hum I don't see this backward effect in the binaryUnion() implementation. At least not in an obvious way

src/operation/union/CascadedPolygonUnion.cpp
pramsey: Annoyingly, OpenMP on Windows insists that the loop iterators be `int`, which then cascades into all these wonderful casts. 🤷

rouault: ok, a comment in the code would help other readers then

nyalldawson:
Nice work @pramsey ! Have you got any benchtests available of the performance changes yet?

pramsey:
Only on my 2-core laptop, which takes a union from 3.5s to 2.7s. Since the last (biggest) union will always be single-threaded, we'll never get up to 1/N speed-ups with this.

pramsey:
This was fun, but I don't think it's The Way we are going to do parallelism. OpenMP is Too Cute By Half, and probably we're just going to have to bite the bullet and do pthreads implementations for the widest/simplest cross platform acceptance.

PR #483, opened 2021-10-02
TST: Adds several CAPI tests for GEOSDistanceWithin

brendan-ward:
This adds several simple tests for the GEOSDistanceWithin CAPI function, structured similarly to the tests in GEOSPreparedDistanceWithin and the more streamlined test approach in #399, so hopefully these are easy to review.


strk:
Thanks! Tests are always great to have.
I wonder if we should really CHANGE the result for tests involving EMPTY, as I've to say I don't really feel comfortable with an EMPTY geometry being within *any* distance from anything else

strk:
I've merged this with 2696f86ec20ce4100c742b730c87ab401492f1b7

PR #485, opened 2021-10-05
TST: Failing test case for GEOSDistance floating point overflows

brendan-ward:
This adds a test case demonstrating that floating point overflow exceptions are raised by `GEOSDistance` when calculating the distance between boundable geometries but not points with the same distance.

This was detected in this [pygeos issue](https://github.com/pygeos/pygeos/issues/404).  Over there, the floating point status is checked by numpy and bubbles up as a warning in Python.

### DIFF
--- a/tests/unit/capi/GEOSDistanceTest.cpp
+++ b/tests/unit/capi/GEOSDistanceTest.cpp
@@ -9,6 +9,7 @@
 #include <cstdarg>
 #include <cstdio>
 #include <cstdlib>
+#include <fenv.h>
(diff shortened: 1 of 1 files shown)


pramsey:
Thanks, fixed in main, hopefully along with all other examples of this quirk.

pramsey:
Also back ported into 3.9 at c5437e903

brendan-ward:
Thanks @pramsey !  Confirmed the fix is working for this specific case in pygeos.

PR #486, opened 2021-10-10
capi: Fix geometry destructor function name

ghost:
There is no GEOSGeometry_destroy function.
Also fixes generated documentation.


mwtoews:
similarly there is no `GEOSGeometry_setPrecision` function (it should be corrected to `GEOSGeom_setPrecision`)

Issue #501, opened 2021-11-14
Segmentation fault with CascadedPolygonUnion via C API


mwtoews:
A bug reported to Shapely (https://github.com/Toblerity/Shapely/issues/1230) is new with GEOS 3.10. The bug can be reproduced with macOS or Linux and Shapely via ctypes (not main, but perhaps maint-1.8) using the following :
```
$ cat issue-1230.py 
from shapely import geos, wkt
from shapely.ops import unary_union

print(geos.geos_version_string)
empty = wkt.loads("POLYGON EMPTY")
print(empty)
geom = unary_union(empty)
print(geom)
$ export LD_LIBRARY_PATH=/data/mtoews/src/GEOS/builds/geos-3.10.1/lib
$ gdb python3.6-dbg
...
(gdb) run issue-1230.py
Starting program: /usr/bin/python3.6-dbg issue-1230.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[Detaching after fork from child process 15112]
3.10.1-CAPI-1.16.0
POLYGON EMPTY

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff4b1c5b3 in geos::operation::geounion::CascadedPolygonUnion::binaryUnion(std::vector<geos::geom::Geometry const*, std::allocator<geos::geom::Geometry const*> > const&, unsigned long, unsigned long) () from /data/mtoews/src/GEOS/builds/geos-3.10.1/lib/libgeos.so.3.10.1
```
The same issue happens with GEOS 3.10.0. It does not happen with GEOS 3.9.2 built using CMake:
```
$ export LD_LIBRARY_PATH=/data/mtoews/src/GEOS/builds/geos-3.9.2-cmake/lib
$ gdb python3.6-dbg
...
(gdb) run issue-1230.py
Starting program: /usr/bin/python3.6-dbg issue-1230.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[Detaching after fork from child process 15245]
3.9.2-CAPI-1.14.3
POLYGON EMPTY
GEOMETRYCOLLECTION EMPTY
[Inferior 1 (process 15241) exited normally]
```
the issue also does not happen if the input geometry is GEOMETRYCOLLECTION EMPTY for any version. The bad mix appears to be with GEOS 3.10 and POLYGON EMPTY.

[...]

PR #502, opened 2021-11-16
Fix unaryUnion to avoid segfault with empty polygon

mwtoews:
Closes GH-501


mwtoews:
Backported to `3.10` branch (for upcoming 3.10.2) with a3f0368323a491a3639f504c214fefaba0289735

PR #504, opened 2021-11-17
SnapRoundingNoder fix

mngr777:
Fix for "GEOSGeom_setPrecision returns error for valid geometry" https://trac.osgeo.org/geos/ticket/1127

`nearnessTol` calculation moved from SnapRoundingIntersectionAdder to SnapRoundingNoder; same tolerance value is passed to MCIndexNoder.  This allows to detect all cases of vertices positioned close enough to segments to be added as intersections by SnapRoundingIntersectionAdder::processNearVertex.  Currently closely positioned segments can be filtered out in `MonotoneChain::computeOverlaps()` by `MonotoneChain::overlaps()` called with zero `overlapTolerance`.

### DIFF
--- a/src/noding/snapround/SnapRoundingNoder.cpp
+++ b/src/noding/snapround/SnapRoundingNoder.cpp
@@ -79,8 +79,7 @@ SnapRoundingNoder::addIntersectionPixels(std::vector<SegmentString*>& segStrings
 {
     double tolerance = 1.0 / pm->getScale() / INTERSECTION_NEARNESS_FACTOR;
     SnapRoundingIntersectionAdder intAdder(tolerance);
-    MCIndexNoder noder(nullptr, tolerance);
-    noder.setSegmentIntersector(&intAdder);
+    MCIndexNoder noder(&intAdder, tolerance);
     noder.computeNodes(&segStrings);
     std::unique_ptr<std::vector<Coordinate>> intPts = intAdder.getIntersections();
     pixelIndex.addNodes(*intPts);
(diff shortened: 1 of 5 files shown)


Komzpa:
@mngr777 can you please add the minimized test case to the suite so it won't break in the future?
Thanks.

Komzpa (approved):

pramsey:
@dr-jts is going to check this fix on JTS too, just to make sure. Seems pretty magical.

dr-jts:
NIce work on finding this bug.  I fixed it in `SnappingNoder`, but of course `SnapRoundingNoder` needs the same thing.  Glad the fix is so simple (and not some fundamental design flaw in the Snap-Rounding approach!).

src/noding/snapround/SnapRoundingNoder.cpp
```diff
@@ -77,8 +77,9 @@ SnapRoundingNoder::snapRound(std::vector<SegmentString*>& inputSegStrings, std::
 void
 SnapRoundingNoder::addIntersectionPixels(std::vector<SegmentString*>& segStrings)
 {
-    SnapRoundingIntersectionAdder intAdder(pm);
-    MCIndexNoder noder;
+    double tolerance = 1.0 / pm->getScale() / INTERSECTION_NEARNESS_FACTOR;
+    SnapRoundingIntersectionAdder intAdder(tolerance);
+    MCIndexNoder noder(nullptr, tolerance);
     noder.setSegmentIntersector(&intAdder);
```
dr-jts, line 83: You can merge these two lines, like is done [here](https://github.com/libgeos/geos/blob/main/src/noding/snap/SnappingNoder.cpp#L129).

mngr777: Done!  Can't believe I just ignored that nullptr like that.

dr-jts:
Unfortunately it looks to me like this fix (although correct) does not solve 2 of the 3 test cases for in the original GEOS issue.  @mngr777 can you confirm if you are still seeing this issue in `GeometryPrecisionReducer`?  If so, we should reopen an issue for this, and capture the failing test cases.

(Note: the fix I suggested of reducing the precision of the scale factor still works, although it is awkward to do using the current C API).

mngr777:
@dr-jts Yes, 2 cases still produce exceptions.  Please reopen the issue.

Issue #490, opened 2021-10-24
Iterator invalidation is ignored


mmomtchev:
Windows debug builds of 3.10.0 generate a runtime exception in `MakeValid` because there is an iterator invalidation condition that is ignored

The exception is raised on the following line 
https://github.com/libgeos/geos/blob/2653b5bcc2b1cc88fcd24af593a37825229f5e0e/include/geos/index/strtree/TemplateSTRtree.h#L440

The iterator invalidation happens here (I think):
https://github.com/libgeos/geos/blob/2653b5bcc2b1cc88fcd24af593a37825229f5e0e/include/geos/index/strtree/TemplateSTRtree.h#L478

Normally, and depending on the library/compiler, this is something that is supposed to work (especially since the element is added at the back), but it is contrary to the C++ specification and MSVC will detect it in debug mode (it inserts extra runtime checking) and it will trigger an exception

pramsey:
Trying to figure a fix for this:
* have a `childNodes` vector that is built up in the `createParentNodes()` process and then copied into `nodes` vector at each iteration of the `build()` loop? More copying (boo) but separates the vector being iterated on from the additions... 
* start passing around a start iterator and a size instead? (it seems like the end iterator would be the one getting invalidated)

dbaston:
The solution may be to not use iterators. This has been floating around in my subconscious for a bit, but it's hard to be motivated to fully rewrite it because there is no practical problem.

dbaston:
I would interpret [this](https://stackoverflow.com/a/6442829) to mean that all iterators (start and end) are invalidated whenever we do an insertion.

pramsey:
Yeah, it's not an actual problem only a theoretical one (since the vector only grows, never shrinks) and only shows up in debug mode on a single platform. I'm going to Won't Fix this, but note here that PRs are accepted.

mmomtchev:
@pramsey @dbaston, in theory, the C++ specification allows a runtime to implement an `std::vector` iterator as a pointer, not an index - in this case even adding to the vector can invalidate the pointer, as the whole array can be reallocated, but in practice, no one does that
You can probably add [_ITERATOR_DEBUG_LEVEL=0](https://docs.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170) to your Windows build (or directly as a `#define` in the source file)

mmomtchev:
[...]

PR #508, opened 2021-11-18
do not use end() iterators

mmomtchev:
Tentative fix for #490

It avoids using the `end()` iterator which, under a very strict interpretation of the C++ specifications, should be invalid after an `emplace_back`


pramsey:
Thanks!

PR #518, opened 2021-12-03
CMake: improvements so GEOS works as a subproject

rcoup:
Currently GEOS doesn't work as a subproject in CMake via `add_subdirectory()` or FetchContent. 

[Example hello-world super-project](https://gist.github.com/rcoup/2bb5a8b6f0f9d7c26d83f285a7dd314a) & current output.

This is a series of fixes that move GEOS from erroring during generation to building & linking as a CMake subproject:

1. stops using `CMAKE_SOURCE_DIR`/`CMAKE_BINARY_DIR` since they're relative to the top-level project which may not be GEOS.
2. expose `GEOS::geos_c` and `GEOS::geos` alias targets, which is what `find_package(GEOS CONFIG)` has already
3. don't setup testing, packaging, developer-mode, etc when GEOS is a subproject. Moves the `project()` call as early as possible to use [`PROJECT_IS_TOP_LEVEL`](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) for this.
4. fix `geos_c` headers needing `geos` headers when using GEOS from the build tree

I'm not a CMake guru, but 1-3 are recommended practises from [Professional CMake](https://crascit.com/professional-cmake/). I haven't looked at any subproject install issues yet.

### DIFF
--- a/capi/CMakeLists.txt
+++ b/capi/CMakeLists.txt
@@ -18,6 +18,12 @@ target_include_directories(geos_c
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
     $<INSTALL_INTERFACE:include/geos>)
 
+target_include_directories(geos_c
+  INTERFACE
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include/geos>)
+
 # Copy these over so they match the @VARIABLES@ used by autoconf
 # in geos_c.h.in
 set(VERSION ${GEOS_VERSION})
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,3 +253,31 @@ jobs:
        cd build
        ctest --output-on-failure -C Release
 
+  cmake-subproject:
+    name: 'CMake Subproject'
+    runs-on: ubuntu-18.04
+    steps:
+    - name: 'Install'
+      run: |
+        set -e
+        uname -a
+        sudo -E apt-get update
+        sudo -E apt-get -yq --no-install-suggests --no-install-recommends install make python3-pip g++
+        python3 -m pip install --disable-pip-version-check --user cmake==3.13.*
+        echo "$(python3 -m site --user-base)/bin" >> $GITHUB_PATH
+
+    - name: 'Check Out'
+      uses: actions/checkout@v2
+      with:
+        path: geos
+
+    - name: 'CMake Superbuild'
+      run: |
+        set -e
+        cp geos/tests/superbuild.CMakeLists.txt ./CMakeLists.txt
+        cp geos/examples/capi_read.c .
+        cmake --version
+        cmake -S . -B build
+        cmake --build build -j 2
+        build/capi_read
+        test ! -f build/geos/bin/test_geos_unit || { echo "Error: GEOS tests were built" 1>&2 ; exit 1; }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,11 +22,11 @@ if(${CMAKE_GENERATOR} MATCHES "Visual Studio")
   cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
 endif()
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 # TODO: Follow CMake detection of git and version tagging
 #       https://gitlab.kitware.com/cmake/cmake/blob/master/Source/CMakeVersionSource.cmake
-if(EXISTS ${CMAKE_SOURCE_DIR}/.git)
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
   set(GEOS_BUILD_FROM_GIT ON)
 endif()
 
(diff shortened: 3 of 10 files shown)


mwtoews:
This looks like a useful improvement! A few remarks:

- Everything works perfectly, and it didn't before!
- It's missing coverage in CI. This can wait until later, however. With PROJ this is [done within a script](https://github.com/OSGeo/PROJ/blob/2072894385205dc9e24aa86ad3da820391f5f002/travis/install.sh#L125-L144) that runs on several CI jobs. Possibly a different strategy would be needed for geos.
- The `PROJECT_IS_TOP_LEVEL` variable looks useful. But, I wonder if the scope of the broad `if(PROJECT_IS_TOP_LEVEL)` blocks should be restricted a bit more, e.g. `if(PROJECT_IS_TOP_LEVEL AND BUILD_BENCHMARKS)`?
- The `install` target for a subproject will install 471 files from geos. Is this expected? Should a subset of files be installed, e.g. just `lib`? But not `bin/geos-config`, `include`, `lib/cmake` or `lib/pkgconfig`?

rcoup:
> The PROJECT_IS_TOP_LEVEL variable looks useful. But, I wonder if the scope of the broad if(PROJECT_IS_TOP_LEVEL) blocks should be restricted a bit more, e.g. if(PROJECT_IS_TOP_LEVEL AND BUILD_BENCHMARKS)?

So you're thinking just repeat the conditional for every relevant section? eg:

```cmake
#-----------------------------------------------------------------------------
# Tests
#-----------------------------------------------------------------------------
if(PROJECT_IS_TOP_LEVEL)
  include(CTest)
  if(BUILD_TESTING)
    add_subdirectory(tests)
  endif()
endif()

#-----------------------------------------------------------------------------
# Benchmarks
#-----------------------------------------------------------------------------
if(PROJECT_IS_TOP_LEVEL AND BUILD_BENCHMARKS)
  add_subdirectory(benchmarks)
endif()

#-----------------------------------------------------------------------------
# Utils
#-----------------------------------------------------------------------------
if(PROJECT_IS_TOP_LEVEL)
  add_subdirectory(util)
endif()
```

Quite happy to do that if you think it reads more clearly.

> The install target for a subproject will install 471 files from geos. Is this expected? Should a subset of files be installed, e.g. just lib? But not bin/geos-config, include, lib/cmake or lib/pkgconfig?

Yeah, I haven't looked at any subproject install issues yet. My CMake super-project conversion hasn't got that far yet :-) 

In principle I'd suggest just the libraries. Super-projects can always reach in and grab other things or point to them. I was planning a followup-PR once I'd got it all working to improve the install aspect.

rcoup:
> It's missing coverage in CI. This can wait until later, however. With PROJ this is done within a script that runs on several CI jobs. Possibly a different strategy would be needed for geos.

I figured downstream people will complain if it breaks - currently they can't use it at all 😄  But yeah, I think a simple super-project in CI probably isn't hard to do — checking that only relevant stuff is installed (ideally and that irrelevant stuff isn't built), making sure CTest doesn't run, etc.

mwtoews:
Re: repeating the PROJECT_IS_TOP_LEVEL conditional for specific sections, yes I think this might look a bit better. This way it can be more clear and granular which components the subproject logic applies. Perhaps that could do for a first PR.

Agree that further work would be to modify some of the install targets to include/exclude specific components that should be expected with a subproject.

As for the CI side, usually I'd like to see this tested somehow, as it's unfortunate when a new release breaks something downstream. I can probably take care of this at some stage, as I'm keen to eventually see these broader coverages (including post-install tests).

rcoup:
> Re: repeating the PROJECT_IS_TOP_LEVEL conditional for specific sections, yes I think this might look a bit better. This way it can be more clear and granular which components the subproject logic applies. Perhaps that could do for a first PR.

Done.

> As for the CI side, usually I'd like to see this tested somehow, as it's unfortunate when a new release breaks something downstream. I can probably take care of this at some stage, as I'm keen to eventually see these broader coverages (including post-install tests).

I added a simple GH Actions workflow to at least check it builds as a subproject, and that the tests aren't built. Is a starting point at least.

rcoup:
@mwtoews let me know if you want any further changes here

mwtoews (approved):
Thanks for the updates, this looks excellent! There are no conflicts to any other aspects of the build, and obviously fixes the library to be used as a sub-project.

I'll let this wait for a few days if anyone else wants to weigh-in, but will otherwise use a "merge commit" to preserve the individual commits.

mwtoews:
Thanks again @rcoup 

A "merge commit" was not enabled for this repo (to my supprise), so I edited the "squash and merge" commit with the primary component from each message.

capi/CMakeLists.txt
```diff
@@ -18,6 +18,12 @@ target_include_directories(geos_c
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
     $<INSTALL_INTERFACE:include/geos>)
 
+target_include_directories(geos_c
+  INTERFACE
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include/geos>)
+
```
SpaceIm, line 26: Sorry to drop a review 7 months after the PR, but I see these modifications with the new release. Actually this block is useless and misleading, `geos_c` doesn't have any interface dependency to `geos` headers, just private usage of `geos` headers handled by the link of `geos` target in top CMakeLists.

Actually, everything can be simplified with:

```cmake
target_include_directories(geos_c PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
```

and it was already in this CMakeLists

rcoup: It's 7 months later, so I'll need to go and dig it out again... but _something like this_ was clearly needed for it to build/work as a subproject, otherwise it wouldn't have been here 😃 

As a guess, I think because pre-install when building the super-project `source/include/` is needed, since `install/include/geos` isn't populated yet?

PR #519, opened 2021-12-05
Optimizations to RingClipper implementation

dbaston:
- Inline trivial methods
- Use std::vector instead of CoordinateSequence internally
- Avoid some Coordinate copies

### DIFF
--- a/include/geos/operation/overlayng/RingClipper.h
+++ b/include/geos/operation/overlayng/RingClipper.h
@@ -81,17 +81,64 @@ private:
     /**
     * Clips line to the axis-parallel line defined by a single box edge.
     */
-    std::unique_ptr<CoordinateArraySequence> clipToBoxEdge(const CoordinateSequence* pts, int edgeIndex, bool closeRing) const;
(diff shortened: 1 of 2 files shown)


pramsey (approved):
Looks good to me.

dbaston:
It seems to cause an infinite loop in MSVC for some reason.

pramsey:
Just that one version too, very impressive.

dbaston:
The "infinite loop" turned out to be the result of MSVC trying to pop up an "assertion failed" dialog box on an Azure server somewhere with nobody to click "OK." I fixed that error but can no longer verify a performance improvement so I'm holding this for now.

robe2:
Shouldn't this and #605 be overlaid.

dbaston:
I can revisit this if/after #721 is merged.

dbaston:
If we wanted to optimize this (and I'm not sure it's really warranted), I think the time is probably better directed bringing in a more efficient algorithm. Some interesting ones are implemented by QGIS in https://github.com/qgis/QGIS/blob/84c3613cf7f4fae1c6b872956a7784a83a135653/src/core/qgsclipper.h

PR #521, opened 2021-12-06
docs: update flags used in tar example

dylrich:
The documentation suggests using the tar flag 'z', but it appears that downloads are bzipped instead of gzipped, so 'j' should be the recommended flag.

### DIFF
--- a/web/content/usage/download.md
+++ b/web/content/usage/download.md
@@ -28,7 +28,7 @@ Builds with CMake are done "outside the tree" either in a build directory in the
 
 ```bash
 # Unpack and setup build directory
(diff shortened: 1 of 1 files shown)


pramsey:
Actually... Seems like modern versions  of tar (bsd (bsdtar 3.3.2) and gnu (tar (GNU tar) 1.34) variants) don't care if you (a) use a compress flag at all (b) use x or j! For complete correctness I'll accept the PR anyways :)

dylrich:
Ah! I'm using 1.29 and should have checked it wasn't just old version behavior. Thanks!

PR #523, opened 2021-12-07
WIP: Add forEach and forEachSegment methods to CoordinateSequence

dbaston:
This PR adds `forEach` and `forEachSegment` methods to `CoordinateSequence`.

Iterating over a `CoordinateSequence` in GEOS is slow because `size`, `getAt` and `operator[]` are all virtual methods that cannot be inlined. To improve performance, a `CoordinateFilter` can be used but this is tedious to write and arguably comes with a readability penalty.

This implementation would allow you to write code like

```
double minX = std::numeric_limits<double>::positive_infinity();
seq.forEach([&minX](const Coordinate* c) {
  minX = std::min(minX, c.x);
});
```

or 

```
seq.forEachSegment[&index](const Coordinate& c1, const Coordinate& c2) {
  index.add(c1, c2);
})
```

Performance is roughly double that of a for loop. Whether that is important depends on the application, but sometimes it is. I suspect that still better performance would be obtained by ditching `CoordinateSequence` in favor of `std::vector<Coordinate>`.

```
2021-12-06T20:00:59-05:00
Running bin/perf_coordseq
Run on (8 X 4200 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 6144 KiB (x1)
Load Average: 1.27, 0.55, 0.38
---------------------------------------------------------------------------------------
Benchmark                                             Time             CPU   Iterations
---------------------------------------------------------------------------------------
BM_CoordinateSequenceForLoop                      41666 ns        41383 ns        16913
BM_CoordinateSequenceForLoopCacheSize             22494 ns        22265 ns        35950
BM_CoordinateSequenceFilter                       17066 ns        17065 ns        46334
BM_CoordinateSequenceForEach                      16230 ns        16228 ns        42949
BM_CoordinateSequenceSegmentForLoopCacheSize      49302 ns        49154 ns        14228
BM_CoordinateSequenceSegmentInspector             29130 ns        28899 ns        25200
BM_CoordinateSequenceSegmentForEach               28161 ns        27407 ns        23331
```

If this seems like a good approach I will add tests and apply the method where applicable through the code base.

### DIFF
--- a/benchmarks/geom/CMakeLists.txt
+++ b/benchmarks/geom/CMakeLists.txt
@@ -18,4 +18,16 @@ IF(benchmark_FOUND)
             $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>)
     target_link_libraries(perf_envelope PRIVATE
             benchmark::benchmark geos_cxx_flags)
+
+    add_executable(perf_coordseq
+            CoordinateSequencePerfTest.cpp
+            ${CMAKE_SOURCE_DIR}/src/geom/Coordinate.cpp
+            ${CMAKE_SOURCE_DIR}/src/geom/CoordinateArraySequence.cpp
+            ${CMAKE_SOURCE_DIR}/src/geom/CoordinateArraySequenceFactory.cpp
+            ${CMAKE_SOURCE_DIR}/src/geom/CoordinateSequence.cpp)
+    target_include_directories(perf_coordseq PUBLIC
+            $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include>
+            $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>)
+    target_link_libraries(perf_coordseq PRIVATE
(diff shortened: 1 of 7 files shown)


pramsey:
(Proforma apologies for ignorance)...
Why do we not have
```
for (auto& coord: coordSeq) {
   x = coord.x;
}
```
Would this still be suffering under the weight of virtual, even for the CoordinateArraySequence?

dbaston:
It would be nice. I can't figure out a way to plumb it to avoid a virtual call on each access. We'd need to define a `begin()` and `end()` on `CoordinateSequence`. Those methods would have to have a consistent return type (`CoordinateSequenceIterator` ?) for all implementations. So we'd either (a) have `CoordinateSequenceIterator` call our virtual `getAt()` method, or (b) we'd end up with a `CoordinateArraySequenceIterator` subclass, with a virtual dereference operator.

PR #525, opened 2021-12-07
inline: ensure inline.h is included in all .inl includers

### LINKED ISSUE #524: Some headers use `GEOS_INLINE` without necessary include
Every header file that has an associated `.inl` file should include `<geos/inline.h>` to ensure that `GEOS_INLINE` is defined and the associated `.inl` is included. WKTReader is an example of a header that does not currently do this. There are probably others.

dylrich:
This commit adds inline.h includes in all remaining files that were including an .inl file conditionally based on `GEOS_INLINE`.

Fixes #524


dbaston:
Thanks!

PR #533, opened 2021-12-29
Fix conditional inlining in WKTReader

macdrevx:
This looks like a typo — every other conditional include of an `inl` file is inside of an `#ifdef GEOS_INLINE`, but the one in WKTReader.cpp was within an `#ifndef GEOS_INLINE`.

### DIFF
--- a/src/io/WKTReader.cpp
+++ b/src/io/WKTReader.cpp
@@ -49,7 +49,7 @@
 #include <iostream>
 #endif
 
-#ifndef GEOS_INLINE
+#ifdef GEOS_INLINE
 #include <geos/io/WKTReader.inl>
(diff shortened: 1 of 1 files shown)


macdrevx:
Nevermind — looks like it's about 50/50 `#ifdef` vs `#ifndef`. Not sure what to make of it yet.

pramsey:
The ifdefs are all in the .h files and the ifnders are in the cpp files. "If inlined, put the implementation in the header marked as 'inline'. If not inlined, put the implementation into the module without the 'inline' marker." I'd love to rip the whole thing out and just have "inlines be inlines" but I think there are still some platforms that can't take it.

Issue #539, opened 2022-01-12
MonotoneChainBuilder::findChainEnd: Assertion `start < npts' failed.


strk:
Continuation of bug originally reported in  https://trac.osgeo.org/geos/ticket/1003 -- still occurring in 3.9.0 according to https://gitlab.com/nibioopensource/pgtopo_update_sql/-/issues/193

The MonotoneChainBuilder::findChainEnd method was removed in 3.10 as part of a refactoring in commit c655bb7f1d990e2a8f2560274816dfdb86058417 (PR #414)

strk:
The assert would fail when MonotoneChainBuilder::getChains gets called with a CoordinateSequence containing NO points (empty). An easy fix would be to just have ::getChain return nothing in that case, but it's important to get a testcase going. Should be something about noding as that's where MonotoneChainBuilder::getChain is used from

strk:
The `Closes GH-539` commit log message did NOT result in closing this issue. I now forgot what's the actual syntax to get GH ticket closed. Please GH fans help :) -- I'm closing this manually

PR #540, opened 2022-01-12
Fix MonotoneChainBuilder::getChain with empty CoordinateString

strk:
See https://github.com/libgeos/geos/issues/539

### DIFF
--- a/src/index/chain/MonotoneChainBuilder.cpp
+++ b/src/index/chain/MonotoneChainBuilder.cpp
@@ -70,6 +70,7 @@ public:
 
 private:
     void finishChain() {
+        if ( m_i == 0 ) return;
         std::size_t chainEnd = m_i - 1;
         m_list.emplace_back(*m_seq, m_start, chainEnd, m_context);
         m_start = chainEnd;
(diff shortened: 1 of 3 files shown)


strk:
I'm also moving a test which was previously found in the MonotoneChainBuilderTest.cpp file into a better named Geometry/differenceTest.cpp file (\cc @pramsey)

strk:
The new test is found to segfault in the 3.9 branch (didn't test in main branch yet, CI will)

strk:
Nicely: the testcase segfaults in main branch too

dbaston:
This is testing a null `CoordinateSequence`, not an empty `CoordinateSequence`. Am I missing something?

strk:
@dbaston yes, sorry, now updated (force-pushed) to test an empty one.

strk:
main branch is NOT crashing at ::getChains, just returning an unexpected vector of 1 chain rather than 0 chains (as I'd expect). It still crashes afterwards though, when calling getEnvelope on that single (empty) chain.

dbaston:
Let me know if you want me to fix it. I would probably just guard against `m_i == 0` in `finishChain`

strk:
> Let me know if you want me to fix it. I would probably just guard against `m_i == 0` in `finishChain`

Thanks, I'll take care of it. Will just ping you for review when I'm happy with my changes

PR #541, opened 2022-01-12
[3.9] Fix MonotoneChainBuilder::getChain with empty CoordinateString

strk:
See https://github.com/libgeos/geos/issues/539

Also enable previously disabled Geometry::difference test


strk:
Merged as c80b5bfc07c34ed66fb9edb47ec5c50f6514d90a

Issue #544, opened 2022-01-14
GEOSSTRtree_remove behaviour change in 3.10?


dark-panda:
I am currently upgrading my ffi-geos Ruby gem (https://github.com/dark-panda/ffi-geos) for 3.10 compatibility and ran across a failing test involving `GEOSSTRtree_remove`. The test is effectively the following (forgive the Ruby code, but this should be hopefully fairly straight forward):

```ruby

def test_remove
  @tree = Geos::STRtree.new(3)
  @item_1 = { item_1: :test }
  @item_2 = [:test]
  @item_3 = Object.new

  @geom_1 = read('LINESTRING(0 0, 10 10)')
  @geom_2 = read('LINESTRING(20 20, 30 30)')
  @geom_3 = read('LINESTRING(20 20, 30 30)')

  @tree.insert(@geom_1, @item_1)
  @tree.insert(@geom_2, @item_2)
  @tree.insert(@geom_3, @item_3)

  @tree.remove(read('POINT (5 5)'), @item_1)

  assert_equal([],
    @tree.query(read('LINESTRING(5 5, 6 6)')))

  assert_equal([],
    @tree.query(read('LINESTRING(20 0, 30 10)')))

  assert_equal([@item_2, @item_3],
    @tree.query(read('LINESTRING(25 25, 26 26)')))

  assert_equal([@item_2, @item_3],
    @tree.query(read('LINESTRING(0 0, 100 100)')))
end
```

In 3.9 and prior, the line for `@tree.remove(read('POINT (5 5)'), @item_1)` correctly removes the item from the tree and the tests succeed. In 3.10 and in main, the tests fail because the item is not removed.

I noticed there is no test coverage for `GEOSSTRtree_remove`, which is what I base most of my Ruby tests for new code on, so I'm unsure if I'm correctly implementing `Geos::STRtree#remove` in my Ruby gem, although it has worked up until this point. When I debug the GEOS code via llvm, I'm noticing that it's bailing out at this point:

https://github.com/libgeos/geos/blob/main/include/geos/index/strtree/TemplateSTRtree.h#L309

```c++
bool remove(const BoundsType& itemEnv, const ItemType& item) {
  if (root == nullptr) { // the root is a nullptr coming from my Ruby code
    return false;
  }
```

[...]

PR #545, opened 2022-01-15
Multiple fixes to STRtree remove method

dbaston:
Fixes https://github.com/libgeos/geos/issues/544

### DIFF
--- a/include/geos/index/strtree/TemplateSTRtree.h
+++ b/include/geos/index/strtree/TemplateSTRtree.h
@@ -548,9 +548,11 @@ protected:
 
         for (auto *child = node.beginChildren(); child < node.endChildren(); ++child) {
             if (child->boundsIntersect(queryEnv)) {
-                if (child->isLeaf() && !child->isDeleted()) {
-                    if (!visitLeaf(visitor, *child)) {
-                        return;
+                if (child->isLeaf()) {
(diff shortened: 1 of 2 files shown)


include/geos/index/strtree/TemplateSTRtree.h
```diff
@@ -306,6 +306,8 @@ class TemplateSTRtreeImpl {
     /// @{
 
     bool remove(const BoundsType& itemEnv, const ItemType& item) {
+        build();
```
jorisvandenbossche, line 309: Unrelated to the actual PR, but something that I never found fully clear in the C API docs, and was reminded about when seeing this change.

The comments in this file state:

> once the tree has been built (explicitly or on the first call to `query`), items may not be added or removed.

For which reason I always wondered a bit what the use case if for `GEOSSTRtree_remove` (we also don't expose it in Shapely). But if that comment is correct, and calling `remove` also builds the tree, does that mean you can only call `remove` once? That would be strange, but then the comment is no longer up to date?

dbaston: Yeah, I think the comment is wrong. More items cannot be added, but they can be "removed" -- which really means "retained but flagged to not be returned in queries."

Issue #553, opened 2022-01-27
Add Hilbert code to C API


brendan-ward:
In the Geopandas / shapely ecosystem, there is an interest in calculating Hilbert curve codes for geometries.  Currently this is based on the centroid of a geometry's extent, but someday may be some other representative point.

It looks like it should be straightforward to add parts of `HilbertCode` / `HilbertEncoder` C++ APIs to the C API, but I wanted to get some feedback first before opening a PR.

I think we may want to expose 2 API functions (`_r` and context handle omitted for clarity):

1) `int GEOSHilbert(GEOSGeometry* geom, unsigned int level)`: 
gets envelope, creates `HilbertEncoder` and calls `HilbertEncoder::encode()`.  Return -1 on error since otherwise Hilbert codes should be unsigned 32 bit integers.

This is convenient for our current usage in that we don't first have to do separate calls to GEOS obtain the bounding coordinates and derive the midpoint.

I'm assuming empty geometries should return -1.

2) `int GEOSHilbert_xy(unsigned int x, unsigned int y, unsigned int level)`: 
calls `HilbertCode::encode()`

This provides a more general case that could be used for any arbitrary x,y.

Any opposition to that plan?

Crossref: https://github.com/geopandas/geopandas/pull/2297

/cc @jorisvandenbossche 

(edited to fix that x,y parameters are unsigned ints not doubles; user must have already scaled their double coords to ints before calling)

pramsey:
> Return -1 on error since otherwise Hilbert codes should be unsigned 32 bit integers.

This is problematical, since you can't return -1 in an unsigned 32 bit integer. Maybe instead, 
```
int GEOSHilbert(GEOSGeometry* geom, unsigned int level, unsigned int hilbert)
```
Where the return value is a status. For example, here's the signature for GEOSArea
```
extern int GEOS_DLL GEOSArea(
    const GEOSGeometry* g,
    double *area);
```

brendan-ward:
> you can't return -1 in an unsigned 32 bit integer

correct, which is why I was thinking of returning a signed `int` instead of `unsigned int`.  Sorry, didn't make it explicitly clear that there would be a step between the internal calls to `encode()` (which return unsigned) and the final return (which would have returned signed).

However, I like your suggest of int status return better: 0 on error, 1 on success.

jorisvandenbossche:
> 1. `int GEOSHilbert(GEOSGeometry* geom, unsigned int level)`:
[...]

PR #554, opened 2022-01-30
GEOSExtent function to get overall bounding box of set of geometries

jorisvandenbossche:
Would there be interest in adding such a function? It calculates the aggregated envelope of a set of geometries. 

This goes a bit beyond the typical scope of the current C API functions, as it works on an array of geometries (although you could compare it to `GEOSGeom_createCollection` which also creates a single geometry for an array input, although here we also perform some calculation on the input). 

I should maybe have first opened an issue, but I coded this up anyway to check if it would be beneficial and whether I would propose it. But so if this seems interesting/acceptable, I can further update this PR (docs, tests, etc).

When exposed in python through pygeos/shapely, I compared this with our current "total_bounds" implementation (which gets the bounds of each geometry with `GEOSEnvelope`, get the bbox from that, and then calculates the overall min/max of those bboxes). And this dedicated GEOS function gives around 10x speedup for a simple benchmark case.

I took the "GEOSExtent" naming following PostGIS, which also has a [`ST_Extent`](https://postgis.net/docs/ST_Extent.html) as an aggregation version of [`ST_Envelope`](https://postgis.net/docs/ST_Envelope.html)

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -159,6 +159,7 @@ using geos::geom::PrecisionModel;
 using geos::geom::CoordinateSequence;
 using geos::geom::GeometryCollection;
 using geos::geom::GeometryFactory;
+using geos::geom::Envelope;
 
 using geos::io::WKTReader;
 using geos::io::WKTWriter;
@@ -1061,6 +1062,32 @@ extern "C" {
             Geometry* g3 = g1->getEnvelope().release();
             g3->setSRID(g1->getSRID());
             return g3;
+        });
+    }
+
+    Geometry*
+    GEOSExtent_r(GEOSContextHandle_t extHandle, Geometry** geoms, unsigned int ngeoms)
+    {
+        return execute(extHandle, [&]() {
+            Envelope extent;
+            const Geometry* geom;
+            for (std::size_t i = 0; i < ngeoms; i++) {
+                geom = geoms[i];
+                if (extent.isNull()) {
+                    extent = *(geom->getEnvelopeInternal());
+                } else {
+                    extent.expandToInclude(*(geom->getEnvelopeInternal()));
+                }
+            }
+            // if (extent.isNull()) {
+            //     return;
+            // }
+
+            GEOSContextHandleInternal_t* handle = reinterpret_cast<GEOSContextHandleInternal_t*>(extHandle);
+            const GeometryFactory* gf = handle->geomFactory;
+            std::unique_ptr<Geometry> g;
+            g = gf->toGeometry(&extent);
+            return g.release();
         });
     }
 
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -414,6 +414,12 @@ extern "C" {
     }
 
     Geometry*
+    GEOSExtent(Geometry** geoms, unsigned int ngeoms)
+    {
+        return GEOSExtent_r(handle, geoms, ngeoms);
+    }
+
+    Geometry*
     GEOSIntersection(const Geometry* g1, const Geometry* g2)
     {
         return GEOSIntersection_r(handle, g1, g2);
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -723,6 +723,12 @@ extern GEOSGeometry GEOS_DLL *GEOSEnvelope_r(
     GEOSContextHandle_t handle,
     const GEOSGeometry* g);
 
+/** \see GEOSExtent */
+extern GEOSGeometry GEOS_DLL *GEOSExtent_r(
+    GEOSContextHandle_t handle,
+    GEOSGeometry** geoms,
+    unsigned int ngeoms);
+
 /** \see GEOSIntersection */
 extern GEOSGeometry GEOS_DLL *GEOSIntersection_r(
     GEOSContextHandle_t handle,
@@ -2409,6 +2415,9 @@ extern void GEOS_DLL GEOSGeom_destroy(GEOSGeometry* g);
 * Caller is responsible for freeing with GEOSGeom_destroy().
 */
 extern GEOSGeometry GEOS_DLL *GEOSEnvelope(const GEOSGeometry* g);
+
+/** \see GEOSExtent */
+extern GEOSGeometry GEOS_DLL *GEOSExtent(GEOSGeometry** geoms,unsigned int ngeoms);
 
 /**
 * Returns the intersection of two geometries: the set of points


capi/geos_ts_c.cpp
```diff
@@ -1064,6 +1065,32 @@ extern "C" {
...
     }
 
+    Geometry*
+    GEOSExtent_r(GEOSContextHandle_t extHandle, Geometry** geoms, unsigned int ngeoms)
+    {
+        return execute(extHandle, [&]() {
+            Envelope extent;
+            const Geometry* geom;
+            for (std::size_t i = 0; i < ngeoms; i++) {
+                geom = geoms[i];
```
brendan-ward, line 1075: This probably needs 
```
if (geom == NULL) {
    continue
}
```
(otherwise `geom->` calls below would break)

May also want to combine that with a check for empty too in order to skip over empty geometries.

pramsey:
Why is iterating in shapely so much slower? Is it having to make fresh GEOSGeometry each time in order to call the GEOS function?

jorisvandenbossche:
I haven't profiled it, but doing it with the current C API, `GEOSEnvelope` indeed returns a new Geometry each time, and to get the bounds of that, we get the exterior ring, get the coord seq, and get the x/y coordinates (which means 6 C API calls, I don't know how much overhead that gives) (https://github.com/shapely/shapely/blob/f6185ee1a7ec83b0c7f721a3f08774bb605f2c35/src/ufuncs.c#L2551-L2604). Is there a better way to get the bounds of a Geometry?

dbaston:
> Is there a better way to get the bounds of a Geometry?

https://github.com/isciences/exactextract/blob/88fee6316fd16e9f9083953496208589e852dac8/src/geos_utils.cpp#L97

jorisvandenbossche:
That's the `GEOSGeom_getXMin` et al?

jorisvandenbossche:
Using `GEOSGeom_getXMin` et al instead of going through a CoordSeq seems to be slightly faster to get the bounds, but the specialized function in this PR is still a lot faster.

```python
In [1]: import pygeos 

In [2]: polygons = pygeos.polygons(np.random.randn(1_000_000, 3, 2))

# current way through `bounds` (get bbox of GEOSEnvelope)
In [3]: %timeit pygeos.total_bounds(polygons) 
208 ms ± 1.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# using GEOSExtent of this PR
In [4]: %timeit pygeos.bounds(pygeos.lib.extent(polygons)) 
28.4 ms ± 1.38 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

dr-jts:
It seems unfortunate that the C API doesn't provide a struct equivalent of the `Envelope` class.   That would allow a function:
```
GEOSEnvelopeInternal_r(GEOSContextHandle_t extHandle, const Geometry* g1, GEOSEnvelope* env)
//-- or perhaps GEOSEnvelope could be called GEOSExtent?
```
Which seems it should be a fair bit faster than 4 calls to `GEOSGeom_getXMin` et al?  Maybe fast enough to obviate the need for the proposed `GEOSExtent`?

If that doesn't pan out, or as well as - the current proposal seems reasonable to me.  Although functions operating on arrays of `Geometry*` are a new pattern, the most important concern IMO is that the GEOS API provides minimum impedance to performance.

Does the use case require a `Geometry` as the result of this call?  Or is it really just interested in the extent bounds?  In which case perhaps the idea of supporting a GEOS C `Envelope` equivalent should still be pursued? (Of course the cost of creating and interrogating the envelope geometry is amortized over the entire input array, but still it's nice to not require object creation where not needed.)

dbaston:
> Using GEOSGeom_getXMin et al instead of going through a CoordSeq seems to be slightly faster to get the bounds

What does that implementation look like? I would expect it to be much faster, because it avoids `2N` heap allocations. There is an overhead to function calls but it is tiny compared to heap allocations.

jorisvandenbossche:
Whoops, sorry, I did:

```
      if (!GEOSGeom_getXMin_r(ctx, envelope, x1)) {
          errstate = PGERR_GEOS_EXCEPTION;
          goto finish;
      }
      if (!GEOSGeom_getYMin_r(ctx, envelope, y1)) {
          errstate = PGERR_GEOS_EXCEPTION;
          goto finish;
      }
      if (!GEOSGeom_getXMax_r(ctx, envelope, x2)) {
          errstate = PGERR_GEOS_EXCEPTION;
          goto finish;
      }
      if (!GEOSGeom_getYMax_r(ctx, envelope, y2)) {
          errstate = PGERR_GEOS_EXCEPTION;
          goto finish;
      }
```

but after getting the Envelope, while I of course can directly call this on the input geometry (avoiding the overhead creating/destroying the envelope geometry).
The original is at https://github.com/shapely/shapely/blob/f6185ee1a7ec83b0c7f721a3f08774bb605f2c35/src/ufuncs.c#L2551-L2604, but so in my test I only replaced the part after `envelope = GEOSEnvelope_r(ctx, in1);` with the above (and a quick profile indeed confirms that it is mostly this `GEOSEnvelope` constructing a polygon and destroying the polygon again that takes most time). 

Doing it properly indeed shows a much bigger effect, and now the special case function in this PR is only 2x faster on this simple benchmarks (instead of almost 10x):

```python
In [1]: import pygeos   

In [2]: polygons = pygeos.polygons(np.random.randn(1_000_000, 3, 2))  

In [3]: %timeit pygeos.bounds(pygeos.lib.extent(polygons))   
28.2 ms ± 1.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit pygeos.total_bounds(polygons)               
56.9 ms ± 2.07 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

So yes, this is something we can definitely improve in pygeos / shapely itself (thanks for the hint!)

dbaston:
If that can satisfy the urgent need, I think it would be worth giving some thought to the more general pattern here. For example, would it make sense to have a non-owning `GeometryCollection` that you could then pass to many existing C API functions? OTOH, I'm not seeing a huge maintenance burden with `GEOSExtent`, so I have no objection to merging it in (after adding tests) if it really helps. Another option is a `GEOSEnvelope` variant that takes `xmin, xmax, ymin, ymax` pointers as arguments, to reduce the number of function calls by 4. I'm more skeptical of exposing an `Envelope` object at the C API level, because then we end up adding a heap allocation to pass four numbers around.

dr-jts:
> If that can satisfy the urgent need, I think it would be worth giving some thought to the more general pattern here. For example, would it make sense to have a non-owning `GeometryCollection` that you could then pass to many existing C API functions? 

Hmm, not sure.  There is an appeal to sticking with OGC structures for sure, but they come with baggage.  Surely this is more heavyweight than just passing an array?  And might be confusing in the API signature?  Although if it is standalone and not a Geometry subclass, then it's less confusing.

> Another option is a `GEOSEnvelope` variant that takes `xmin, xmax, ymin, ymax` pointers as arguments, to reduce the number of function calls by 4. 

Seems like a good option to me (I'm actually surprised that wasn't the original design direction).

> I'm more skeptical of exposing an `Envelope` object at the C API level, because then we end up adding a heap allocation to pass four numbers around.

If the function allowed pass-by-reference couldn't the caller just allocate the `Envelope` struct locally?

dbaston:
> Surely this is more heavyweight than just passing an array? 

A GeometryCollection is nothing more than an array of geometries with an envelope so no, I don't think it imposes a large overhead. There are many uses for a non-owning geometry collection such as unioning a set of geometries while preserving the originals, buffering a set of geometries while preserving the originals, etc. If you make a method to create such a collection you get all this functionality without writing a GEOSUnionArray, GEOSBufferArray, etc.

> If the function allowed pass-by-reference couldn't the caller just allocate the Envelope struct locally?

The GEOS C API uses opaque pointers, so the caller cannot allocate them.

dr-jts:
> There are many uses for a non-owning geometry collection such as unioning a set of geometries while preserving the originals, buffering a set of geometries while preserving the originals, etc. If you make a method to create such a collection you get all this functionality without writing a GEOSUnionArray, GEOSBufferArray, etc.

Agreed, it's handy to use `GeometryCollection` in this way, since you get lots of things for free, and it keeps the API simpler.

> > If the function allowed pass-by-reference couldn't the caller just allocate the Envelope struct locally?
> The GEOS C API uses opaque pointers, so the caller cannot allocate them.

So you can't pass a pointer to a stack-allocated struct?  But you can to `double`?

dr-jts:
> would it make sense to have a non-owning `GeometryCollection` that you could then pass to many existing C API functions? 

@dbaston how would a non-owning `GeometryCollection` be implemented?

jorisvandenbossche:
Before trying the approach in this PR, I actually initially tried to get a speed-up for this "global extent" for a set of geometries by doing exactly the "put them in a collection and ask for the bounds of this one geometry" (assuming that it would internally basically do the same as I do here for the array of geometries). 
However, that's much slower indeed because of the cost of cloning all geometries and creating the collection. 

And for an existing example: for the `unary_union` implementation in shapely/pygeos, we put the input geometries in a GeometryCollection (to only directly destroy it afterwards). So that would be a nice use case for such a non-owning collection.

jorisvandenbossche:
BTW, also in https://github.com/libgeos/geos/issues/553, the pattern of working on an array of geometries was brought up (but also there, there is probably a good enough way without it for now)

dbaston:
> So you can't pass a pointer to a stack-allocated struct? But you can to double?

You can pass a pointer to a stack-allocated struct, but the design of the GEOS C API has the definitions of the structs in the implementation file (`.cpp`). The caller, who is working only with the header, does not know the layout or contents of the struct, so they cannot create one on the stack.

> how would a non-owning GeometryCollection be implemented?

By redefining the `geometries` member of `GeometryCollection` to be something other than `std::vector<std::unique_ptr<Geometry>>`. Finding the best way do to that is would require some thought. We could define our own pointer type that keeps track of its ownership, or we could just go back to `std::vector<Geometry*>` and keep track of whether `~GeometryCollection` should delete `geometries` or not.

dbaston:
A less elegant but very easy way to accomplish the same thing would be to expose `GeometryCollection::releaseGeometries` to the C API. So you could take a set of geometries, stick them in a collection, do something (e.g. union) with the collection, then release ownership from the collection.

dr-jts:
> > So you can't pass a pointer to a stack-allocated struct? But you can to double?
> You can pass a pointer to a stack-allocated struct, but the design of the GEOS C API has the definitions of the structs in the implementation file (`.cpp`). The caller, who is working only with the header, does not know the layout or contents of the struct, so they cannot create one on the stack.

Well, yes, would have to expose the definition of the `Envelope` struct in the header file.  Is this a bad thing?

dr-jts:
> A less elegant but very easy way to accomplish the same thing would be to expose `GeometryCollection::releaseGeometries` to the C API. So you could take a set of geometries, stick them in a collection, do something (e.g. union) with the collection, then release ownership from the collection.

Easy is good.  And since `GeometryCollection::releaseGeometries` is the technique for this in the C++ code, seems reasonable to expose to the C API as well?

dbaston:
> Well, yes, would have to expose the definition of the Envelope struct in the header file. Is this a bad thing?

It's not how the rest of the GEOS C API works, though you could probably break the convention if you had a good reason to. Remember that the header and the shared library might come from different versions of GEOS and might be built using different compilers. Passing a struct (or pointer to a struct) between them assumes that both client and library define the struct the same way. Probably that is safe for a struct of four doubles. With the opaque pointer approach that GEOS uses, only the library knows the layout of the struct, so it's not a concern.

dr-jts:
> > Well, yes, would have to expose the definition of the Envelope struct in the header file. Is this a bad thing?
> 
> It's not how the rest of the GEOS C API works, though you could probably break the convention if you had a good reason to. Remember that the header and the shared library might come from different versions of GEOS and might be built using different compilers. Passing a struct (or pointer to a struct allocated by the client) between them assumes that both client and library define the struct the same way. Probably that is safe for a struct of four doubles. With the opaque pointer approach that GEOS uses, only the library knows the layout of the struct, so it's not a concern.

Makes sense, thanks for the design info.  As you say, for a simple struct should not be a big concern.  But maybe simpler just to provide a method like this?:
```
GEOSGeom_getExtent_r(GEOSContextHandle_t extHandle, const Geometry* g, 
      double* xmin, double* ymin, double* xmax, double* ymax)
```

strk:
The reason why C-API uses opaque structures is specificall to
avoid ABI breakage. the C-API ABI *never* broke, so it was
a successful approach. Let's try to think ahead and make it
virtually impossible for it to break.

brendan-ward:
> `GEOSGeom_getExtent_r(GEOSContextHandle_t extHandle, const Geometry* g, 
      double* xmin, double* ymin, double* xmax, double* ymax)`

and the (roughly) symmetric 

```
GEOSGeometry* GEOSGeom_createEnvelope_r(GEOSContextHandle_t extHandle, 
      double xmin, double ymin, double xmax, double ymax)
```

would be nice to cut down on some of the (calling side) overhead around getting / setting extents and then using them for later operations.

For instance, in #553 which is a bit related here, since we'd want to get the total bounds of all geometries then immediately use that as input to get Hilbert codes, having a lightweight ability to create and pass in an envelope seems more in line with the C existing API than passing in xmin, ymin, xmax, ymax to that function)

jorisvandenbossche:
I opened https://github.com/libgeos/geos/pull/555 for a `GEOSGeom_getExtent` that gives you the 4 numbers with one API call. On the simple polygon case I showed above, it's another ~25% faster compared to using the 4 `GEOSGeom_getX/Ymin/max` calls.

dbaston:
Performance aside, it seems reasonable to add from an API usability standpoint.

jorisvandenbossche:
Yes, certainly! (I was just curious what the difference would be)

jorisvandenbossche:
Indeed, this discussion resulted in adding `GEOSGeom_getExtent`

PR #556, opened 2022-02-01
ENH: Add GEOSHilbertCode to C API

### LINKED ISSUE #553: Add Hilbert code to C API

brendan-ward:
closes #553

Exposes `HilbertEncoder` to C API.

`HilbertEncoder` will return seemingly valid values for geometries that fall outside the extent used for scaling prior to calling 
`HilbertCode`.  Should this instead be handled as an error?  For now I noted that it is the caller's responsibility to ensure the geometry falls in the extent.

I used @jorisvandenbossche 's [GeoPandas PR #2297](https://github.com/geopandas/geopandas/pull/2297) to validate the Hilbert Codes used in the test here, since both use the same calculations.

### DIFF
--- a/src/shape/fractal/HilbertEncoder.cpp
+++ b/src/shape/fractal/HilbertEncoder.cpp
@@ -65,7 +65,7 @@ HilbertEncoder::sort(std::vector<geom::Geometry*>& geoms)
         bool
         operator()(const geom::Geometry* a, const geom::Geometry* b)
         {
-            return enc.encode(a->getEnvelopeInternal()) > enc.encode(b->getEnvelopeInternal());
+            return enc.encode(a->getEnvelopeInternal()) < enc.encode(b->getEnvelopeInternal());
         }
     };
 
--- a/src/shape/fractal/HilbertEncoder.cpp
+++ b/src/shape/fractal/HilbertEncoder.cpp
@@ -65,7 +65,7 @@ HilbertEncoder::sort(std::vector<geom::Geometry*>& geoms)
         bool
         operator()(const geom::Geometry* a, const geom::Geometry* b)
         {
-            return enc.encode(a->getEnvelopeInternal()) < enc.encode(b->getEnvelopeInternal());
+            return enc.encode(a->getEnvelopeInternal()) > enc.encode(b->getEnvelopeInternal());
         }
     };
 
--- a/tests/unit/shape/fractal/HilbertEncoderTest.cpp
+++ b/tests/unit/shape/fractal/HilbertEncoderTest.cpp
@@ -66,7 +66,7 @@ void object::test<2>() {
     ensure_equals(encoder.encode(point2->getEnvelopeInternal()), 0u);
 }
 
-// test that geometries are sorted
+// test that geometries are sorted in ascending Hilbert order
 template <>
 template <>
 void object::test<3>() {
@@ -85,7 +85,7 @@ void object::test<3>() {
     encoder.sort(geoms);
 
     // geometries will be sorted in descending Hilbert order
-    uint32_t expected[3] = {10, 2, 0};
+    uint32_t expected[3] = {0, 2, 10};
     for (size_t i = 0; i < 3; i++) {
         ensure_equals(encoder.encode(geoms[i]->getEnvelopeInternal()), expected[i]);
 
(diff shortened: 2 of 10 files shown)


brendan-ward:
Note: I think that we don't actually need `GEOSHilbert_xy` as proposed in #553; caller can construct and pass in points that represent whatever they want (midpoint, point on surface, something else entirely) so long as the fall within the extent.

tests/unit/shape/fractal/HilbertEncoderTest.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    encoder.sort(geoms);
+
+    // geometries will be sorted in descending Hilbert order
+    uint32_t expected[3] = {10, 2, 0};
+    for (size_t i = 0; i < 3; i++) {
+        ensure_equals(encoder.encode(geoms[i]->getEnvelopeInternal()), expected[i]);
+
+        // cleanup geom
+        factory_->destroyGeometry(geoms[i]);
```
brendan-ward, line 93: I found something like this in one of the other tests, but I'm not confident this is the right way to handle this; without it there is a memory leak.

dbaston: That's fine. `GEOSGeom_destroy` is OK too. It's awkward because the interface of `HilbertEncoder` is forcing you to work with raw pointers.

src/shape/fractal/HilbertEncoder.cpp
```diff
@@ -65,7 +65,7 @@ HilbertEncoder::sort(std::vector<geom::Geometry*>& geoms)
         bool
         operator()(const geom::Geometry* a, const geom::Geometry* b)
         {
-            return enc.encode(a->getEnvelopeInternal()) > enc.encode(b->getEnvelopeInternal());
+            return enc.encode(a->getEnvelopeInternal()) < enc.encode(b->getEnvelopeInternal());
```
brendan-ward, line 68: @pramsey I changed the sort direction to better match expected results; `std::sort` returns items in ascending order, and the few other places I checked in GEOS that were sorting objects with a comparator like this were also doing so in ascending order.  Presuming the descending Hilbert sort order was not intentional, of course.

I didn't see any other usage of this function in GEOS.  Does this merit a note in the `NEWS`?

brendan-ward: Was only changing out of interest of consistency / expected results; since this PR added tests based on expected behavior, seemed like a good time to do so.

Definitely don't want to piss anyone off, just thought it was odd that a sort that normally would produce ascending order did the opposite here.

Would you like me to revert this change?

pramsey: Yes, I think so. One of the privileges of being a widely used library is being stuck with your history, particularly in the case of behaviour.

dbaston: The new C API function could do an ascending sort, no?

pramsey: True

PR #558, opened 2022-02-03
ENH: Add GEOSGeom_createRectangle to C API

brendan-ward:
(edited to match latest signature)

This adds a convenience method to the C API that creates a bounding envelope polygon from bounding coordinates:

```
GEOSGeometry* GEOSGeom_createRectangle(double xmin, double ymin, double xmax, double ymax)
```

~~This simply exposes the existing `Rectangle::toPolygon()` method.~~
This wraps the `GeometryFactory::toGeometry(Envelope* envelope)` method

It helps users of the C API from having to make a bunch of calls to create a coordinate sequence, set coords, create a ring, then create a polygon.

This could use an entirely different name if more appropriate.  Some other options:
* `GEOSGeom_createRect`
* `GEOSGeom_createBox`

I only used `envelope` for potential consistently with `GEOSEnvelope` which returns something of the same structure.


capi/geos_ts_c.cpp
```diff
@@ -2685,6 +2685,23 @@ extern "C" {
...
+    GEOSGeom_createEnvelope_r(GEOSContextHandle_t extHandle,
+                            double xmin, double ymin,
+                            double xmax, double ymax)
+    {
+        using geos::geom::Polygon;
+        using geos::operation::intersection::Rectangle;
+
+        return execute(extHandle, [&]() {
+            GEOSContextHandleInternal_t* handle = reinterpret_cast<GEOSContextHandleInternal_t*>(extHandle);
+            const GeometryFactory* gf = handle->geomFactory;
```
dbaston, line 2698: `GeometryFactory::toGeometry(const Envelope*)` is probably the one to use here.

tests/unit/capi/GEOSGeom_createEnvelopeTest.cpp
```diff
@@ -0,0 +1,33 @@
...
+template <>
+void object::test<1>() {
+    GEOSGeometry* expected = GEOSGeomFromWKT("POLYGON ((0 0, 0 2, 1 2, 1 0, 0 0))");
+    GEOSGeometry* geom = GEOSGeom_createEnvelope(0, 0, 1, 2);
+
+    ensure_equals(GEOSEqualsExact(geom, expected, 0), 1);
+
+    GEOSGeom_destroy(expected);
+    GEOSGeom_destroy(geom);
+}
```
dbaston, line 31: Should probably also pin down the behavior for degenerate cases like `createEnvelope(0,0,0,0)`.

brendan-ward: Meaning, degenerate case should raise an exception & return `NULL`?

dbaston: I don't really have an opinion on what the behavior should be (though it would be easy to just go with what the `GeometryFactory` method produces). I just think it's important to put in tests to avoid unintentional future changes in that behavior, whatever it is.

dr-jts:
Is this to support things like queries by extent against spatial indexes?

If so, then huh...  sure would be nice if GEOS had a true `Envelope` structure in the C API, to make that a lot more efficient.

brendan-ward:
Yes, this is intended to help with inputs to query against STRtree as well as to prepare input for things like #556.

> `GEOSGeom_createRectangle`

Sounds good, will change.

dbaston:
I wouldn't recommend this for querying an STRtree (use a two-point LineString or an existing geometry instead). It's still a useful addition.

tests/unit/capi/GEOSGeom_createEnvelopeTest.cpp
brendan-ward: I opted to add validation to the C API side, and raise an error if the bounds do not produce a rectangle with both width and height.

The `GeometryFactory` method produces a point if there is no width and height, which didn't seem right for the return value here.

brendan-ward:
Thanks for the review and tips @dbaston !

tests/unit/capi/GEOSGeom_createEnvelopeTest.cpp
dr-jts: I would opt for not enforcing non-zero width/height.  That doesn't cause any structural problems, it just creates an invalid geometry.  Most (all) GEOS constructors don't enforce validity - they leave that up to the client if it's required.  And this lets the function create extents which can represent point queries.

I think I'd also check the order of the input ordinates and normalize them as required (i.e. swap to ensure min/max order), to produce a rectangle in a known orientation.   Just like the Envelope constructor does.

brendan-ward: Looks like the Envelope contructor takes care of normalizing the coordinates, so I didn't add anything extra there.

Added note to indicate that this can return a point if width and height are 0.

dr-jts: Yes, that's a good way to handle it.

Issue #565, opened 2022-02-11
Geometry.intersects is wrong for intersection very close to boundary point


pramsey:
As reported in [postgis-users](https://lists.osgeo.org/pipermail/postgis-users/2022-February/045264.html).

For this pair of geometries `intersects` incorrectly reports `false`.  `PreparedGeometry.intersects` correctly reports `true`, however.
```
01020000200912000007000000642F25DC75A24CC0E4DE5740FCB34840A7CEFE9B72A24CC09DA85B2CFBB34840B5519D0E64A24CC091FAA188FBB34840FA449E245DA24CC054C2137AFDB34840F4ACFFCE51A24CC09FEB562A03B448405328C1D144A24CC09A3DD00A0CB44840404C10C03CA24CC0EA07FE6910B44840
```
and
```
0102000020091200000E00000004BE47A23CA24CC098A1F14410B448409871AEBC3FA24CC078341F2114B448400858AB764DA24CC09D0546031DB448406BFD3E2D50A24CC0BEDDEDD522B4484004824AA654A24CC02DC9A60128B44840EE377FB850A24CC0FA18BD642DB44840CCAF8B474EA24CC02CCCE78134B44840D7158E7B4EA24CC01D7C17A53AB44840ACFA01B452A24CC02688BA0F40B44840DB508C8752A24CC006CDF80846B44840A1F31ABB44A24CC0C891730756B44840009AF7EE45A24CC06B7649415CB448408C2ECAC749A24CC0CE57248161B44840A74302A150A24CC07DD00E1368B44840
```

dr-jts:
This case also has `intersects`= FALSE and `intersectsPrep` = TRUE
```
LINESTRING (-29796.696826656284 138522.76848210802, -29804.3911369969 138519.3504205817)
```
```
LINESTRING (-29802.795222153436 138520.05937757515, -29802.23305474065 138518.7938969792)
```

PR #566, opened 2022-02-11
PrepGeom->intersects() should equal Geom->intersects()

pramsey:
But for some geometries that is not so! This pull request adds a test demonstrating the problem and stands ready to receive the solution. Addresses #565.


dr-jts:
Superseded by #570

PR #588, opened 2022-03-10
add windows/arm64 cross-compilation build to Azure pipeline

nsait-linaro:
Adding a build step to CI pipeline for cross-compilation of `geos` library for windows/arm64 platforms.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -197,10 +197,17 @@ stages:
         #   TOOLSET: msvc-14.3
         #   CXXSTD: 20
         #   VM_IMAGE: 'windows-2022'
-        VS 2019 C++20:
+        VS 2019 C++20 x64:
           TOOLSET: msvc-14.2
           CXXSTD: 20
           VM_IMAGE: 'windows-2019'
+          ARCH: x64
+        VS 2019 C++20 ARM64:
+          TOOLSET: msvc-14.2
+          CXXSTD: 20
+          VM_IMAGE: 'windows-2019'
+          ARCH: ARM64
+
     pool:
       vmImage: $(VM_IMAGE)
     steps:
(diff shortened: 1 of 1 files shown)


.azure-pipelines.yml
```diff
@@ -212,19 +219,20 @@ stages:
...
-        cd build.cxx%CXXSTD%
+        cd build.cxx%CXXSTD%.%ARCH%
         cmake --build . --config %BUILD_TYPE%
       displayName: 'Build'
     - script: |
-        cd build.cxx%CXXSTD%
+        cd build.cxx%CXXSTD%.%ARCH%
         ctest -V --output-on-failure -C %BUILD_TYPE%
       displayName: 'Test'
+      condition: and(succeeded(), eq(variables['ARCH'], 'x64'))
```
mwtoews, line 235: Why skip Test for ARM64? Were there failures?

mwtoews:
I see there are [many warnings for ARM64](https://dev.azure.com/libgeos/geos/_build/results?buildId=1867&view=logs&jobId=dded70eb-633c-5c42-e995-a7f8d1f99d91&j=dded70eb-633c-5c42-e995-a7f8d1f99d91&t=696f2e49-e3ef-5e19-60e9-ed2e10d75470) that are absent from the [x64 equivalent](https://dev.azure.com/libgeos/geos/_build/results?buildId=1867&view=logs&jobId=2bd7b19d-6351-5e7f-8417-63f327ab45bc&j=2bd7b19d-6351-5e7f-8417-63f327ab45bc&t=549c8367-5889-5507-1708-200535760ded). Not sure what the cause is.

pramsey:
Am I right that the implication here is we have access to actual ARM64 runners on Azure now? We should have some Linux ARM64 spun up too. Like, this isn't actually "cross compilation", is it? Or is that why the tests aren't being run?

nsait-linaro:
This is actually a "cross-compilation" so we cannot run the test cases. Azure still doesn't have ARM64(Windows) runners. Hopefully very soon! I

pramsey:
I'm going to bring this in while feeling very free to just take it out again if it gets noisy and nobody cares.

PR #596, opened 2022-04-20
Debug output fixes

mngr777:
* Compilation and runtime error fixes for GEOS_DEBUG=1;
* `#ifdef GEOS_DEBUG` replaced with `#if GEOS_DEBUG` (since GEOS_DEBUG is always defined in .cpp);
* few other minor fixes related to GEOS_DEBUG

### DIFF
--- a/include/geos/algorithm/LineIntersector.h
+++ b/include/geos/algorithm/LineIntersector.h
@@ -20,14 +20,11 @@
 #pragma once
 
 #include <geos/export.h>
-#include <geos/algorithm/LineIntersector.h>
 #include <geos/algorithm/Intersection.h>
 #include <geos/geom/Coordinate.h>
 #include <geos/geom/Envelope.h>
 
 #include <string>
-
-#include <geos/geom/Coordinate.h>
 
 // Forward declarations
 namespace geos {
@@ -86,6 +83,7 @@ public:
         :
         precisionModel(initialPrecisionModel),
         result(0),
+        inputLines(),
         isProperVar(false)
     {}
 
--- a/src/algorithm/HCoordinate.cpp
+++ b/src/algorithm/HCoordinate.cpp
@@ -45,7 +45,7 @@ HCoordinate::intersection(const Coordinate& p1, const Coordinate& p2,
 
 #if GEOS_DEBUG
     std::cerr << __FUNCTION__ << ":" << std::endl
-         << setprecision(20)
+         << std::setprecision(20)
          << " p1: " << p1 << std::endl
          << " p2: " << p2 << std::endl
          << " q1: " << q1 << std::endl
--- a/src/algorithm/InteriorPointLine.cpp
+++ b/src/algorithm/InteriorPointLine.cpp
@@ -29,7 +29,7 @@
 #define GEOS_DEBUG 0
 #endif
 
-#ifdef GEOS_DEBUG
+#if GEOS_DEBUG
 #include <iostream>
 #endif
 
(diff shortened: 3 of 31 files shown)


dr-jts:
The plethora of this boilerplate
```
#ifndef GEOS_DEBUG
#define GEOS_DEBUG 0
#endif
```
is brittle and somewhat opaque.  It would be nice to find a way to eliminate it.

Is it possible to treat `GEOS_DEBUG` not defined as equal to `GEOS_DEBUG = 0`?   I.e. what is the result of `#if GEOS_DEBUG` and `#if GEOS_DEBUG > 1` if `GEOS_DEBUG` is not defined?

Alternatively, perhaps most debugging blocks can use `#ifdef GEOS_DEBUG`, so it only needs to be defined externally?

@strk thoughts?

pramsey:
Yeah, moving in the direction of universally using #ifdef instead of #if would hopefully result in less boilerplate.

mngr777:
If GEOS_DEBUG is undefined, then it will be evaluated to 0, so `GEOS_DEBUG > 1` will be 0.

C++ standard 15.2.11, 2020/03/31:
> After all replacements due to macro expansion and evaluations of defined-macro-expressions, has-include-
> expressions, and has-attribute-expressions have been performed, all remaining identifiers and keywords, except
> for true and false, are replaced with the pp-number 0

gcc has -Wundef flag to make it complain when undefined macro is used.

I think, it's better to move this code to e.g. `include/geos/Debug.h` and just include it in files that need debug output.
It can even define a macro similar to `qDebug` in Qt, so one liners can be written like this:
```
#include <geos/Debug.h>
// ...
geosDebug() << "debug message" << std::endl;
```
instead of 
```
#if GEOS_DEBUG
   std::cerr << "debug message" << std::endl;
#endif
```
https://github.com/qt/qtbase/blob/b6251958932be5966258f362dc9bb7c7475d2d0a/src/corelib/global/qlogging.h

dr-jts:
+1 for adding a `Debug.h`.  

Sounds like `#define GEOS_DEBUG 0` isn't needed in any case, right?

strk:
Whatever works for you, I'm happy :)
Only keep in mind the need of debugging a *SINGLE* file, which I often use, toggling a pre-existing line on the top of the file (uncommenting a commented-out line would have the same simplicity)

dr-jts:
>  Only keep in mind the need of debugging a _SINGLE_ file, which I often use, toggling a pre-existing line on the top of the file (uncommenting a commented-out line would have the same simplicity)

Agreed, that is a definite requirement.  Not sure how this should work.  Can `#define X` be overridden?

mngr777:
Here's one idea how to allow debugging for single file or just a part of it even.

main.cpp:
```
int main() {
	// No output by default
#include "./debug.h"
	geosDebug() << "message 0" << std::endl;

	// Outputs "message 1"
#undef GEOS_DEBUG
#define GEOS_DEBUG 1
#include "./debug.h"
	geosDebug() << "message 1" << std::endl;

	// No output
#undef GEOS_DEBUG
#define GEOS_DEBUG 0
#include "./debug.h"
	geosDebug() << "message 2" << std::endl;
}
```
debug.h:
```
// No "#pragma once"

#ifndef GEOS_DEBUG
# define GEOS_DEBUG 0
#endif

#include <iostream>

#undef geosDebug

#if GEOS_DEBUG
# define geosDebug() \
	(std::cerr)
#else
# define geosDebug() \
	while (false) (std::cerr)
#endif
```

It not actually needed to re-define `geosDebug` and omit `#pragma once` if `debug.h` is only included once in .cpp files anyway.
`debug.h` should be somewhere in `src`, not in `include`, so it's not installed with public headers.

dr-jts:
Perhaps we should merge this PR, since it has multiple fixes.  And then hope for a new one to simplify the debugging harness?

PR #597, opened 2022-04-20
Add LineMerger directed option

mngr777:
Directed line merger implementation https://github.com/locationtech/jts/issues/773

Is it ok to start with C++ implementation and backport to JTS?


dr-jts:
It seems like it should be possible to implement this using most of the current `LineMerger` code, but with a flag indicating whether edges are allowed to be "flipped".   Have you looked into this?  Any reason that won't work?

mngr777:
@dr-jts Updated to include API header changes.

tests/unit/capi/GEOSLineMergeDirectedTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+//
+// Test Cases
+//
+
+template<>
+template<>
+void object::test<1>
+()
+{
+    auto input = GEOSGeomFromWKT("MULTILINESTRING((0 0, 0 100),(0 -5, 0 0))");
```
pramsey, line 36: If you use `geom1_` and `geom2_` as the variables, the `capitest::utility` will automatically clean them up after each test so you can drop the `GEOSGeom_destroy` calls.

PR #607, opened 2022-05-15
install geos.h header file in cmake builds

tbussmann:
since the change from autoconf to cmake this header was not longer installed with `make install`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -420,6 +420,8 @@ if(NOT DISABLE_GEOS_INLINE)
 endif()
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/capi/geos_c.h"
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+install(FILES "${CMAKE_CURRENT_LIST_DIR}/include/geos.h"
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
 add_subdirectory(tools)
 


pramsey:
On the one hand I like the idea of geos.h pulling in the "official C++ API", we've deprecated the C++ since 15 years ago.

https://github.com/libgeos/geos/blame/main/include/geos/geom.h#L144

Wonder what @dbaston thinks?

hobu:
> we've deprecated the C++ since 15 years ago.

"Hurry up, you've had 15 years to convert your code to use this slower, heavier, harder-to-use C API we've generously provided for you. Time's up."

dbaston:
Sounds fine to me, once we decide what the "C++ API" is :)

pramsey:
Yeah! I was wondering if these headers even still worked. Like, can I write the C++ example apps up in terms of just "include geos.h and go". I have no idea how deep the dependency graph reaches once you include it.

tbussmann:
I don't know if and how often that header is used for quick access to the deprecated C++ API, but no longer providing it seems to be an unintended change introduced by the switch to the cmake-only build system. If you conclude leaving this out helps to transition away of it, the [readme.md](https://github.com/libgeos/geos#using-the-c-interface-1) should be adopted, at least.

pramsey:
Taking this one in, I think it's good for the future (maybe) but good to retain continuity with the past until/if we deliberately change.

PR #610, opened 2022-05-24
Fix memory-handling bugs found by GCC 12

dr-jts:
* Avoid potential buffer overrun by copy in FixedSizeCoordinateSequence
* Disable Voronoi Diagram use of userData to avoid returning pointer to stack mem

### DIFF
--- a/include/geos/geom/FixedSizeCoordinateSequence.h
+++ b/include/geos/geom/FixedSizeCoordinateSequence.h
@@ -105,7 +105,9 @@ namespace geom {
         }
 
         void setPoints(const std::vector<Coordinate> & v) final override {
-            std::copy(v.begin(), v.end(), m_data.begin());
+            if (N > 0 && v.size() <= N) {
+                std::copy(v.begin(), v.end(), m_data.begin());
+            }
         }
 
         void apply_ro(CoordinateFilter* filter) const final override {
@@ -126,4 +128,3 @@ namespace geom {
 
 }
 }
-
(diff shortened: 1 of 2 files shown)


include/geos/geom/FixedSizeCoordinateSequence.h
```diff
@@ -105,7 +105,9 @@ namespace geom {
         }
 
         void setPoints(const std::vector<Coordinate> & v) final override {
-            std::copy(v.begin(), v.end(), m_data.begin());
+            if (N > 0 && v.size() <= N) {
```
dbaston, line 108: I would probably opt for a crash (status quo) or exception over a silent failure here.

dr-jts: Good idea.  Is there an existing pattern for an exception?

dbaston: The existing pattern (in GEOS, as well as the STL) is to crash if the caller tries to access an index that does not exist, which essentially is what is happening here. You could throw a `GEOSException` if `v.size() != N`. An `assert` would be the middle ground.

dr-jts: How would a crash work?  Does the original code crash, or just overwrite mem?  

In any case, need an explicit check for the new GCC.  So need to do something on failure.  I like `assert` best, since it makes the contract explicit.

PR #611, opened 2022-05-25
Add assert to FixedSizeCS.setPoints

dr-jts:
Improve error detection and handling in `FixedSizeCoordinateSequence`.  Check for the input to have the same size as the target with an `assert`.  Prevent running `std::copy` if the sequence has zero-length (to avoid compiler complaints).


dr-jts:
Closed due to overly-aggresive or spurious GCC 12 warnings.

PR #615, opened 2022-05-26
GEOS_DEBUG: missing header fix

mngr777:
Should `#if GEOS_DEBUG  ... #endif` be removed around `assert(...)`?


pramsey:
No, asserts aren't free.

PR #616, opened 2022-05-26
Fix MinimumDiameter rectangle for flat input

dr-jts:
Fixes `MinimumDiameter.getMinimumRectangle` to avoid an infinite loop for flat input.

Fixes bug reported in [PostGIS-5163](https://trac.osgeo.org/postgis/ticket/5163).


src/algorithm/MinimumDiameter.cpp
```diff
@@ -337,6 +340,42 @@ MinimumDiameter::getMinimumRectangle()
...
+{
+    //-- find max and min pts for X and Y
+    Coordinate ptMinX = pts->getAt(0);
+    Coordinate ptMaxX = pts->getAt(0);
+    Coordinate ptMinY = pts->getAt(0);
+    Coordinate ptMaxY = pts->getAt(0);
+
+    std::size_t const n = pts->getSize();
+    for(std::size_t i = 1; i < n; ++i) {
+      Coordinate p = pts->getAt(i);
```
dbaston, line 356:
```suggestion
      const Coordinate& p = pts->getAt(i);
```

dr-jts: Thanks!

PR #666, opened 2022-08-22
More mutable SegmentStrings

pramsey:
Current segments strings have a fairly ugly API, and quite different behaviour with regard to ownership of the CoordinateArray underlying them. The NodedSegmentString takes ownership, while the BasicSegmentString takes a const pointer.

I had hoped to cut down on copying by allowing the NodedSegmentString to take a const pointer and only start managing a read/write copy in the cases where someone called for such a copy. I removed getCoordinates() from the API and replaced it with specific getCoordinatesRO() and getCoordinatesRW(). Then I went through and changed all the calls appropriately. 

Inside both implementations, there are both a const pointer and a unique_ptr to CoordinateArray, and when the getCoordinatesRW() method is called, only then are the coordinates copied out of the const pointer. (This whole two-step is almost exclusively to support the way the ScaledNoder is implemented, as re-scaling the coordinates of an object in-place, then passing them to a nested noded. I found almost nothing else mutating the points in a SegmentString directly.)

For NodedSegmentString, I added a constructor that took at const CoordinateSequence pointer, so a NodedSegmentString could be constructed against a read-only set of points. Also added a unique_ptr constructor, for the case when the NodedSegmentString is taking ownership. This is actually quite common, it turns out, for all kinds of cases of constructive geometry.

This whole effort was started because back when doing overlayNG it became clear the first step of building up the edges was taking a complete copy of the coordinates from the input geometry, even though nothing was really being done to them. Actually, that perception was slightly off: the clipping shortcut in the overlayNG code does tend to mutate coordinates rate out of the gate; also **ever** input coordinate array has repeated points stripped out. And that routine, for simplicity, just iterates on the input copy and generates a completely new output copy without dupes. Even when there actually **are** no dupes. So there is space to get rid of a bunch of the input copying, certainly, but it will take re-working the overlayNG code a fair bit relative to the JTS baseline it seems.

So far there is not really any strong measured performance improvement from this work.


dbaston:
Thanks for digging in and documenting all this. I'll have time to start joining the effort next month - looking forward to it!

Do you have any thoughts on the viability of `std::shared_ptr<const CoordinateSequence*>` as way to store sequences in a `Geometry` and pass them pass around through operations like these?

pramsey:
In the overlay code, a key problem is that CoordinateSequence spends some time on its own, after being found in the geometry and before being incorporated into a SegmentString, and in that time it might-or-might-not need to be mutated in the course of being clipped or having repeated points removed. So it's not that a Geometry and a SegmentString might both own a CoordinateSequence for a while, so much as the life cycle of a lonely CoordinateSequence sometimes involves getting mutated (but often not).

I think even with a shared pointer the code in the EdgeNodingBuilder needs to be re-organized to bring the place where the CoordinateSequence is found in the Geometry and the place where it's incorporated into a SegmentString closer together.

pramsey:
Now just a footnote as @dbaston continues his work

Issue #669, opened 2022-09-01
WKTReader: allow TYPEZ, TYPEM, TYPEZM


lbartoletti:
GEOS can read Z/M Geometry Type. However, the parser failed to read it when there is no space between the type name and the z/m extension.

"LINESTRING Z(0 0 1, 1 1 1)" is OK but
"LINESTRINGZ(0 0 1, 1 1 1)" is not

For example, GeosOp cannot read the second one: https://github.com/libgeos/geos/pull/668

dbaston:
Somewhat related: https://trac.osgeo.org/geos/ticket/676. I will look into both.

lbartoletti:
A very quick&dirty fix is:

```cpp
std::unique_ptr<Geometry>
WKTReader::readGeometryTaggedText(StringTokenizer* tokenizer) const
{
    std::string type = getNextWord(tokenizer);
    const std::vector<std::string> ext{"", "Z", "ZM", "M"};
    for (const auto &e: ext) {
      if(type == "POINT" + e) {
          return readPointText(tokenizer);
      }
      else if(type == "LINESTRING" + e) {
          return readLineStringText(tokenizer);
      }
      else if(type == "LINEARRING" + e) {
          return readLinearRingText(tokenizer);
      }
      else if(type == "POLYGON" + e) {
          return readPolygonText(tokenizer);
      }
      else if(type == "MULTIPOINT" + e) {
          return readMultiPointText(tokenizer);
      }
      else if(type == "MULTILINESTRING" + e) {
          return readMultiLineStringText(tokenizer);
      }
      else if(type == "MULTIPOLYGON" + e) {
          return readMultiPolygonText(tokenizer);
      }
      else if(type == "GEOMETRYCOLLECTION" + e) {
          return readGeometryCollectionText(tokenizer);
      }
    }
    throw ParseException("Unknown type", type);
}
```

The problem, is that the parser wants a fixed geometry type name ; the Z and M extension are handled later by getNextEmptyOrOpener. This is OK, but you need a space between the geometry type name and the extension (or empty).

A second fix, is to check the start of the string and lets the dedicated function to handle the exception later:

```cpp
std::unique_ptr<Geometry>
WKTReader::readGeometryTaggedText(StringTokenizer* tokenizer) const
{
    std::string type = getNextWord(tokenizer);
      if(type.rfind("POINT", 0) == 0) {
          return readPointText(tokenizer);
      }
      else if(type.rfind("LINESTRING") == 0) {
[...]

PR #671, opened 2022-09-03
Resolve two bugs in WKTReader

dbaston:
This PR corrects the handling of dimension flags (Z and M) in WKT input and enforces consistent dimensionality with a geometry. It resolves https://github.com/libgeos/geos/issues/669, https://trac.osgeo.org/geos/ticket/676, https://trac.osgeo.org/geos/ticket/1080, and https://trac.osgeo.org/geos/ticket/1095


src/io/WKTReader.cpp
```diff
@@ -210,60 +254,62 @@ std::unique_ptr<Geometry>
...
-        return readGeometryCollectionText(tokenizer);
+    else if(isTypeName(type, "GEOMETRYCOLLECTION")) {
+        return readGeometryCollectionText(tokenizer, ordinateFlags);
     }
     throw ParseException("Unknown type", type);
 }
 
 std::unique_ptr<Point>
-WKTReader::readPointText(StringTokenizer* tokenizer) const
+WKTReader::readPointText(StringTokenizer* tokenizer, OrdinateSet ordinateFlags) const
```
pramsey, line 288: Are `ordinateFlags` deliberately pass-by-value? For multi-part geometries, sometimes knowing the tagged dimensionality is necessary to handle conflicts in dimensionality deeper down (`GEOMETRYCOLLECTION Z(POINT Z(0 0 0), LINESTRING M(1 1 1, 2 2 2))`)

pramsey (approved):
Very pretty, as always.

src/io/WKTReader.cpp
dbaston: This PR ensures that valid WKT with Z/M is read correctly but does not ensure that invalid WKT throws an error (the full scope of https://trac.osgeo.org/geos/ticket/1080). I can add that behavior to this PR if it's desirable.

From that ticket it sounds like nested geometries with different coordinate dimensions should _not_ be allowed. In that case, I guess we want to set the ordinate flags after reading the first coordinate of the first geometry, then "lock" them and throw a `ParseException` if subsequent coordinates do not comply? (which would require `ordinateFlags` to be pass-by-reference, as you suggest)

pramsey: This is what PostGIS parser does. For tagged dimensionality, the tag rules. For no-tag case, expect consistency with the first coordinate. I suppose throwing would be a new behaviour, in case anyone is feeding mixed dimensionality into GEOS right now. Those poor unfortunate souls.

dbaston: I've updated the PR to enforce dimensionality within a geometry. I think this wraps up https://trac.osgeo.org/geos/ticket/1080.

PR #678, opened 2022-09-12
Fix PreparedGeometry for EMPTY elements

dr-jts:
Fixes handling of EMPTY elements in PreparedGeometry methods, by add empty coordinate sequence guards in several places.

See https://github.com/locationtech/jts/pull/904

### DIFF
--- a/src/noding/MCIndexSegmentSetMutualIntersector.cpp
+++ b/src/noding/MCIndexSegmentSetMutualIntersector.cpp
@@ -45,9 +45,10 @@ MCIndexSegmentSetMutualIntersector::addToIndex(SegmentString* segStr)
 void
 MCIndexSegmentSetMutualIntersector::addToMonoChains(SegmentString* segStr)
 {
+    if (segStr->size() == 0)
+        return;
     MonotoneChainBuilder::getChains(segStr->getCoordinates(),
                                     segStr, monoChains);
-
 }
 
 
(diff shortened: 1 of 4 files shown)


src/noding/MCIndexSegmentSetMutualIntersector.cpp
```diff
@@ -45,9 +45,10 @@ MCIndexSegmentSetMutualIntersector::addToIndex(SegmentString* segStr)
 void
 MCIndexSegmentSetMutualIntersector::addToMonoChains(SegmentString* segStr)
 {
+    if (segStr->size() == 0)
```
dbaston, line 48: The new test passes without the changes to this class...is there a case where this guard would be hit?

dr-jts: Odd, that is required in JTS.  The fix seems necessary, though, since MonotoneChains should not be built on empty coordinate lists.

dr-jts: I see the cause now.  The GEOS codebase has diverged from JTS, by using the [`ChainBuilder`](https://github.com/libgeos/geos/blob/main/src/index/chain/MonotoneChainBuilder.cpp#L49) class.  Presumably this handles empty coordinate sequences correctly.  So in GEOS this is only a small performance optimization (and thus hard/unnecessary to test).

dr-jts: The JTS `MonotoneChainBuilder` should be hardened against empty inputs as well.

dbaston: Yeah, the `ChainBuilder` is one of several places where GEOS adds a `CoordinateFilter` as a workaround for slow Coordinate access. If we decide to go with #674 we can remove those workarounds.

PR #680, opened 2022-09-13
Fix RelateOp for empty geometry and closed linear geometry

dbaston:
This PR fixes the following Trac ticket: https://trac.osgeo.org/geos/ticket/1096. It ports https://github.com/locationtech/jts/pull/671.

As a side effect, it brings along the `BoundaryOp` class that has been in JTS for some time, and its associated tests.

### DIFF
--- a/include/geos/geom/GeometryFactory.h
+++ b/include/geos/geom/GeometryFactory.h
@@ -252,6 +252,8 @@ public:
     /// Construct a MultiPoint taking ownership of given arguments
     MultiPoint* createMultiPoint(std::vector<Geometry*>* newPoints) const;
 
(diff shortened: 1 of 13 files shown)


dbaston:
Thoughts on backporting, since the changeset is relatively large?

dr-jts:
> Thoughts on backporting, since the changeset is relatively large?

Always nice to backport bug fixes, especially for something as core as spatial predicates.

How far back can you just cherry-pick?

dbaston:
> How far back can you just cherry-pick?

3.11?

dr-jts:
> > How far back can you just cherry-pick?
> 
> 3.11?

Sigh... that's not great.

Oh well, I guess if someone really wants it they can complain loudly.

PR #681, opened 2022-09-14
Improve CoverageUnion performance by pre-sorting inputs

dbaston:
This PR makes `CoverageUnion` pre-sort its inputs using `HilbertEncoder`. This improves performance by about 50% on my test data set:

Before:

```
bin/geosop -t -q -a ~/data/tl_2021_counties_unsorted.wkt coverageUnion
Ran 1 coverageUnion ops ( 8,092,118 vertices)  -- 4,654,931 usec    (GEOS 3.12.0dev)
```

After:

```
bin/geosop -t -q -a ~/data/tl_2021_counties_unsorted.wkt coverageUnion
Ran 1 coverageUnion ops ( 8,092,118 vertices)  -- 2,238,353 usec    (GEOS 3.12.0dev)
```

### DIFF
--- a/include/geos/operation/union/CoverageUnion.h
+++ b/include/geos/operation/union/CoverageUnion.h
@@ -24,6 +24,7 @@ namespace geos {
     namespace geom {
         class Polygon;
         class LineString;
+        class LinearRing;
         class GeometryFactory;
     }
 }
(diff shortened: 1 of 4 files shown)


dr-jts:
Wow, that's a big difference.  So is this because the increased spatial coherency of the line segments improves the performance of the `std::unordered_set` operations `emplace` and `erase` (used to find and removed duplicate internal segments)?

I had noticed that the OvelayNG-based `CoverageUnion` seemed to be up to 50% faster than your implementation on larger datasets, and was puzzled why.  Sounds like this eliminates that difference, which is great.  (In fact, the OverlayNG CoverageUnion produces wrong results in some cases due to an unknown bug, and it's more complex than this implementation, so it's good to use this one instead.)

dr-jts:
One more thing that is on my ToDo list is to replace the current implementation of `Polygonizer` with one based on the `HalfEdge` graph structure instead.  That is a simpler structure, and more memory-efficient.

dr-jts:
It might also be worth looking at [`BoundaryChainNoder`](https://github.com/libgeos/geos/blob/main/src/noding/BoundaryChainNoder.cpp).  It reduces graph size by creating longer `SegmentStrings` for the graph edges.

dbaston:
The `Polygonizer` isn't really the bottleneck here (at least according to the profiling, which sometimes lies.) The runtime is almost all in the set operations, and I assume the sorting improves the efficiency of these by reducing the maximum size of the set (and therefore the number of allocations and rehashing cycles).

I see nearly identical runtimes between `coverageUnion` and `coverageUnionNG` and a similar performance benefit for `coverageUnionNG` with sorted inputs (tested by sorting the file rather than modifying the code)

dr-jts:
Well, that's great - Coverage Union was already pretty darn fast (under both approaches), so this makes it even better.

Now to get it rolled out into PostGIS!

PR #688, opened 2022-09-21
Add clustering functions

dbaston:
This PR is a rebase of #388 which died on the vine in 2021. It adds the DBSCAN, simple distance, and intersection clustering methods available in PostGIS. Also adds clustering based on envelope intersection and envelope distance.

### DIFF
--- a/include/geos/operation/cluster/AbstractClusterFinder.h
+++ b/include/geos/operation/cluster/AbstractClusterFinder.h
@@ -1,0 +1,136 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2020-2022 Daniel Baston
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************/
+
+#ifndef GEOS_OPERATION_CLUSTER_ABSTRACTCLUSTERFINDER
+#define GEOS_OPERATION_CLUSTER_ABSTRACTCLUSTERFINDER
+
+#include <functional>
+#include <memory>
+#include <vector>
+
+#include <geos/export.h>
+#include <geos/index/strtree/TemplateSTRtree.h>
+#include <geos/operation/cluster/Clusters.h>
+
+// Forward declarations
+namespace geos {
+namespace geom {
+    class Envelope;
+    class Geometry;
+}
+namespace operation {
+namespace cluster {
+    class UnionFind;
+}
+}
+}
+
+namespace geos {
+namespace operation {
+namespace cluster {
+
+/** AbstractClusterFinder defines an interface for bottom-up clustering algorithms,
+ * where spatial index queries can be used to identify geometries that should be
+ * clustered together.
+ */
+class GEOS_DLL AbstractClusterFinder {
+
+public:
+    /**
+     * Cluster the provided geometries, returning an object that provides access
+     * to the components of each cluster.
+     *
+     * @param g A vector of geometries to cluster
+     */
+    Clusters cluster(const std::vector<const geom::Geometry*>& g);
+
+    /**
+     * Cluster the components of the provided geometry, returning a vector of clusters.
+     * This function will take ownership of the provided geometry. Any components that
+     * are included in a cluster will be returned. Components that are not included in
+     * any cluster will be destroyed.
+     *
+     * @param g A geometry whose components should be clustered.
+     * @return a Geometry vector, with each entry representing a single cluster.
+     */
+    std::vector<std::unique_ptr<geom::Geometry>> clusterToVector(std::unique_ptr<geom::Geometry> && g);
+
+    /**
+     * Cluster the components of the provided geometry, returning a vector of clusters.
+     * The input geometry will not be modified.
+     *
+     * @param g A geometry whose components should be clustered.
+     * @return a Geometry vector, with each entry representing a single cluster.
+     */
+    std::vector<std::unique_ptr<geom::Geometry>> clusterToVector(const geom::Geometry& g);
+
+    /**
+     * Cluster the components of the provided geometry, returning a GeometryCollection.
+     * This function will take ownership of the provided geometry. Any components that
+     * are included in a cluster will be returned. Components that are not included in
+     * any cluster will be destroyed.
+     *
+     * @param g A geometry whose components should be clustered.
+     * @return a GeometryCollection, with each sub-geometry representing a single cluster.
+     */
+    std::unique_ptr<geom::Geometry> clusterToCollection(std::unique_ptr<geom::Geometry> && g);
+
+    /**
+     * Cluster the components of the provided geometry, returning a GeometryCollection.
+     * The input geometry will not be modified.
+     *
+     * @param g A geometry whose components should be clustered.
+     * @return a GeometryCollection, with each sub-geometry representing a single cluster.
+     */
+    std::unique_ptr<geom::Geometry> clusterToCollection(const geom::Geometry & g);
+
+protected:
+    /**
+     * Determine whether two geometries should be considered in the same cluster.
+     * @param a Geometry
+     * @param b Geometry
+     * @return `true` if the clusters associated with `a` and `b` should be merged.
+     */
+    virtual bool shouldJoin(const geom::Geometry* a, const geom::Geometry* b) = 0;
+
+    /**
+     * Provide an query Envelope that can be used to find all geometries possibly
+     * in the same cluster as the input.
+     * @param a Geometry
+     * @return an Envelope suitable for querying
+     */
+    virtual const geom::Envelope& queryEnvelope(const geom::Geometry* a) = 0;
+
+    /**
+     * Given a vector and index of components,
+     * @param components a vector of Geometry components
+     * @param index a spatial index storing pointers to those components
+     * @param uf a UnionFind
+     * @return a vector of with the indices of all components that should be included in a cluster
+     */
+    virtual Clusters process(const std::vector<const geom::Geometry*> & components,
+                 index::strtree::TemplateSTRtree<std::size_t> & index,
+                 UnionFind & uf);
+
+private:
+    static std::vector<std::unique_ptr<geom::Geometry>> getComponents(std::unique_ptr<geom::Geometry>&& g);
+};
+
+
+
+}
+}
+}
+
+#endif
--- a/include/geos/operation/cluster/GeometryIntersectsClusterFinder.h
+++ b/include/geos/operation/cluster/GeometryIntersectsClusterFinder.h
@@ -1,0 +1,53 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2020-2021 Daniel Baston
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************/
+
+#ifndef GEOS_OPERATION_CLUSTER_GEOMETRYINTERSECTSCLUSTERFINDER
+#define GEOS_OPERATION_CLUSTER_GEOMETRYINTERSECTSCLUSTERFINDER
+
+#include <geos/operation/cluster/AbstractClusterFinder.h>
+#include <geos/geom/Geometry.h>
+#include <geos/geom/prep/PreparedGeometry.h>
+#include <geos/geom/prep/PreparedGeometryFactory.h>
+
+namespace geos {
+namespace operation {
+namespace cluster {
+
+/** GeometryIntersectsClusterFinder clusters geometries by intersection.
+ * Any two geometries that intersect will be included in the same cluster.
+ */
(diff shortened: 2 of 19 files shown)


dr-jts:
I have thought that a nice interface for clustering would be to return a list of cluster ids for the list of input geometries.  That way, the client can (a) know which input geometry belongs to each cluster, and (b) choose whether or not to materialize the clusters as new geometries (and a utility function could support this).  This would also allow the functionality to be used to back window functions in PostGIS.

Does this PR support this kind of interface?

dr-jts:
What is the use case for the envelope clustering functions?

dr-jts:
Is ClusterGeometryIntersecting just a special case of ClusterGeometryDistance (with distance = 0) ?

dbaston:
I wouldn't expect this code to be desirable for PostGIS because PostGIS already has its own implementation that does not require converting everything into GEOS types. But maybe it would be useful for other bindings or for algorithms within GEOS.

> Does this PR support this kind of interface?

At the C++ level, it can and should be modified to support this. In the C API....do you want to post signatures for your ideal C interface? I think at the time I wrote this, I was trying to avoid introducing new types to the C API while also making the common use case of "assemble these geometries into clusters" relatively painless.

Assembling clusters from a list of IDs for each input is a actually a bit painful in C (You can see how I implemented it for PostGIS at https://github.com/postgis/postgis/blob/1dbae611ab087f5c78d94c4a8955196bf887fa9c/liblwgeom/lwgeom_geos_cluster.c#L542.). But maybe nobody is likely to be doing this; they'd be assembling the clusters in Python etc.

> What is the use case for the envelope clustering functions?

Fast alternative to distance or intersection for coarse partitioning of data.

> Is ClusterGeonmetryIntersecting just a special case of ClusterGeometryDistance (with distance = 0) ?

Like in PostGIS, it's implemented using the intersection predicate which is not always consistent with distance == 0.

dbaston:
I changed the C++ interface a bit and the following function is now public:

```
Clusters AbstractClusterFinder::cluster(const std::vector<const geom::Geometry*>& g);
```

this can be transformed into a vector of cluster IDs whose elements correspond to `g` using `Clusters::getClusterIds()`

dbaston:
which I now realize will not work for DBSCAN clustering because not all inputs are assigned a cluster. Would you want some placeholder cluster ID value used in this case? Or an array of booleans to accompany the array of IDs?

dr-jts:
> Like in PostGIS, it's implemented using the intersection predicate which is not always consistent with distance == 0.

That's not good. Do you have an example?  Is this a true semantic difference, or just a robustness issue?

dr-jts:
> which I now realize will not work for DBSCAN clustering because not all inputs are assigned a cluster. Would you want some placeholder cluster ID value used in this case? Or an array of booleans to accompany the array of IDs?

An ID indicating "no cluster assigned" seems pretty simple.  E.g. ID = -1.

dr-jts:
>  In the C API....do you want to post signatures for your ideal C interface? 

How about an array of ints?  I suppose the size is needed.  Infer from the input size?  Provide as an out parameter? Indicate via a sentinel value in the array?

dbaston:
> That's not good. Do you have an example? Is this a true semantic difference, or just a robustness issue?

```
dan=# SELECT array_length(ST_ClusterIntersecting(geom), 1) FROM tl_2016_50017_linearwater;
 array_length 
--------------
          558
(1 row)

dan=# SELECT array_length(ST_ClusterWithin(geom, 1e-300), 1) FROM tl_2016_50017_linearwater;
 array_length 

--------------
          482
(1 row)
```
[tl_2016_50017_linearwater.zip](https://github.com/libgeos/geos/files/9621101/tl_2016_50017_linearwater.zip)

That said, maybe `distance == 0` and `intersects` are equivalent in GEOS. 
 
> How about an array of ints? I suppose the size is needed. Infer from the input size? Provide as an out parameter? Indicate via a sentinel value in the array?

I'd be fine with removing the C API for now. This interface would have been painful for my use case for the reasons I described above, but maybe I'm the only one.

dr-jts:
> I'd be fine with removing the C API for now. This interface would have been painful for my use case for the reasons I described above, but maybe I'm the only one.

Not sure I understand.  A lot of use of GEOS comes through the C API, so that seems important to provide?

dr-jts:
I wonder whether the input should be provided as a array of geometries, rather than a `GeometryCollection`?  Simpler to construct, and doesn't have the nested semantics of collections to deal with.  That also nicely matches a potential return type of an array of cluster IDs.

dr-jts:
> That said, maybe `distance == 0` and `intersects` are equivalent in GEOS.

I guess you could confirm by testing using this PR?  Would be interesting to know if there is a discrepancy.  (Also why this happens in PostGIS - but I think the distance code and predicate code there are of different origins, so maybe have different contracts?)

dr-jts:
> > What is the use case for the envelope clustering functions?
> 
> Fast alternative to distance or intersection for coarse partitioning of data.

Understood.  I'm curious if there is any papers or discussion about using this technique.  Envelopes are a pretty coarse approximation, so it seems like this might have some undesirable characteristics in practice.

dbaston:
> I wonder whether the input should be provided as a array of geometries, rather than a GeometryCollection? Simpler to construct, and doesn't have the nested semantics of collections to deal with. That also nicely matches a potential return type of an array of cluster IDs.

I'm really trying to follow the contours of the existing C API, which uses `GeometryCollection` to handle for groups of geometries. I don't think they are especially complex to construct or access and unlike an array they know their own size. A non-owning variant of `GeometryCollection` could be useful. 

> A lot of use of GEOS comes through the C API, so that seems important to provide?

It's ideal to expose it to the C API, but it doesn't have to all happen at once. If you don't like the API I've proposed and you don't have a detailed alternative, I don't think there's anything wrong with punting.

> I'm curious if there is any papers or discussion about using this technique. Envelopes are a pretty coarse approximation, so it seems like this might have some undesirable characteristics in practice.

I used it to optimize raster reads from sparse polygons (think USA or France with overseas territories.) I'm not aware of any papers or discussion of it.

dbaston:
Here's an example where envelope pre-clustering improves `unaryUnion` runttime by 5% despite only dividing the data into two clusters.

```
Geometry::Ptr
Geometry::Union() const
{
    using geos::operation::geounion::UnaryUnionOp;
#ifdef DISABLE_OVERLAYNG
    return UnaryUnionOp::Union(*this);
#else

    operation::cluster::EnvelopeIntersectsClusterFinder f;
    auto clustered = f.cluster(*this);

    for (auto& g : clustered) {
        g = operation::overlayng::OverlayNGRobust::Union(g.get());
    }

    return this->getFactory()->buildGeometry(std::move(clustered));
#endif
}
```

using test `geosop unaryUnion -t -q -a ~/data/world.wkt`

dbaston:
Although swapping in a `GeometryIntersectionClusterFinder` comes up with 91 clusters and gives about a 15% performance benefit.

dbaston:
Flattening the inputs, per the latest commit, brings the performance gain to 60% for this dataset.

dbaston:
I've gone ahead and added pre-clustering by intersection to `Geometry::Union()`. I'm not seeing an adverse performance impact even in cases like Paul's watershed dataset where we end up with a single cluster. More testing would be great.

include/geos/operation/cluster/AbstractClusterFinder.h
```diff
@@ -0,0 +1,136 @@
...
+     *
+     * @param g A vector of geometries to cluster
+     */
+    Clusters cluster(const std::vector<const geom::Geometry*>& g);
+
+    /**
+     * Cluster the components of the provided geometry, returning a vector of clusters.
+     * This function will take ownership of the provided geometry. Any components that
+     * are included in a cluster will be returned. Components that are not included in
+     * any cluster will be destroyed.
```
dr-jts, line 62: Doesn't this semantic need to be documented in the C API?  Or is it inferred?

include/geos/operation/cluster/GeometryIntersectsClusterFinder.h
```diff
@@ -0,0 +1,54 @@
...
+#include <geos/operation/cluster/AbstractClusterFinder.h>
+#include <geos/geom/Geometry.h>
+#include <geos/geom/prep/PreparedGeometry.h>
+#include <geos/geom/prep/PreparedGeometryFactory.h>
+
+namespace geos {
+namespace operation {
+namespace cluster {
+
+/** GeometryIntersectsClusterFinder clusters geometries by envelope intersection.
```
dr-jts, line 27: Envelope intersection?

src/geom/Geometry.cpp
```diff
@@ -595,7 +597,13 @@ Geometry::Union() const
 #ifdef DISABLE_OVERLAYNG
     return UnaryUnionOp::Union(*this);
 #else
-    return operation::overlayng::OverlayNGRobust::Union(this);
+
+    operation::cluster::GeometryIntersectsClusterFinder f;
+    operation::cluster::DisjointOperation op(f);
+
+    return op.processDisjointSubsets(*this, [](const Geometry& g) {
```
dr-jts, line 604: If this is superior, it would be better to provide it as a separate/new operation class, rather than burying it under Geometry implementation.

src/operation/union/CoverageUnion.cpp
```diff
@@ -109,18 +111,23 @@ std::unique_ptr<Geometry> CoverageUnion::polygonize(const GeometryFactory* gf) {
...
 std::unique_ptr<geom::Geometry> CoverageUnion::Union(const geom::Geometry* geom) {
-    CoverageUnion cu;
+    double area_in = geom->getArea();
+    const auto& factory = geom->getFactory();
 
-    cu.extractRings(geom);
-    cu.sortRings();
+    operation::cluster::GeometryIntersectsClusterFinder f;
+    operation::cluster::DisjointOperation op(f);
+    auto ret = op.processDisjointSubsets(*geom, [&factory](const geom::Geometry& g) {
```
dr-jts, line 119: better to split this out into a separate PR, so it's more visible.   Would be nice to see metrics on performance increase in that PR.

src/geom/Geometry.cpp
dr-jts: Also, this is new implementation for a widely-used operation.  So better to put it in a separate PR, in case of issues.

dr-jts: Would be good to see metrics on when and how this improves performance.  Is there any negative affect for cases which have no disjoint geometry?

dr-jts: is this safe when Overlay invokes snapping or snap-rounding?  That can cause overlay results to intersect even when the input is disjoint.

dr-jts:
> I wouldn't expect this code to be desirable for PostGIS because PostGIS already has its own implementation that does not require converting everything into GEOS types. 

It would be nice if the PostGIS functions `ST_ClusterIntersecting` and `ST_ClusterWithin` could be converted to window functions (with new names of course).   To do this, would it make sense to back them onto the GEOS code?  Or just rework the PG-native code?

dbaston:
> To do this, would it make sense to back them onto the GEOS code? Or just rework the PG-native code?

I would expect PostGIS to stick with its existing code which already provides this capability (look at `ST_ClusterDBSCAN` implementation; `ST_ClusterWithin` is implemented as a special case of `ST_ClusterDBSCAN`). Switching to GEOS would require converting geometries into GEOS and dropping support for curve types that are supported by PostGIS but not GEOS.

src/geom/Geometry.cpp
dbaston: This passes all existing test cases but I suspect that it is not safe. Can you point me to an example of disjoint inputs producing an intersecting result so I can try to provoke a failure?

include/geos/operation/cluster/AbstractClusterFinder.h
dbaston: It can be inferred from the fact that we're passing a non-const pointer, but it can also be documented. But I am leaning towards removing the C API until I have a better understanding of if/how it would be used from shapely et al.

src/operation/union/CoverageUnion.cpp
dr-jts: Sorry, my comment was not clear.  I mean move all changes to `CoverageUnion` to a separate PR, as a follow-up to this PR.  Likewise for changes to Geometry unary union.

src/geom/Geometry.cpp
dr-jts: Unfortunately I don't have a test case for this situation.  I think I have seen it in the wild, however.

src/operation/union/CoverageUnion.cpp
dbaston: It's clear, I just haven't separated it yet.

src/geom/Geometry.cpp
dbaston: Could the overlay short-circuits for disjoint inputs be similarly affected? Like `OverlayUtil::isEmptyResult` ?

dr-jts: No.  When snap-rounding is required explicitly by providing a precision model, the disjoint checks are done using that precision model.  And for full precision invocations, the disjoint check is accurate as is.  

The risk occurs when a robustness error forces snapping or snap-rounding to be used internally.  In this case the envelope of the result can be larger than the envelope of the inputs.

dbaston: I guess the workaround would be to cluster based on distance. This currently has some severe performance problems that I am working to address (#691)

dbaston: Or, to verify that the results of the sub-unions remain disjoint.

dbaston:
I've removed all changes to the C API and Union operations. Will post the union changes in a separate PR.

PR #693, opened 2022-09-27
Optimize prepared polygon distance

dbaston:
The implementation of `PreparedPolygonDistance` begins with a call to `PreparedPolygonIntersects` to cover the case where a containment relationship exists between the target and test geometries. However, the segment intersection test performed by `PreparedPolygonIntersects` is unnecessary. If any segment intersect, they will be found by `IndexedFacetDistance`.

This PR improves the performance of `PreparedPolygonDistance` by performing the PIP containment tests directly within `PreparedPolygonDistance` and avoiding the call to `PreparedPolygonIntersects`. In the following performance test, which relies heavily on `PreparedPolygonDistance`, runtime is reduced by 45%.

`bin/geosop clusterWithin 1e-3 -t -q -a ~/data/world.wkt`

This should bring additional performance improvements to #692.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,7 +18,8 @@ xxxx-xx-xx
   - GeometryPrecisionReducer: Return correct dimensionality for empty results (GH-684, Dan Baston)
   - Improve performance of coverage union (GH-681, Dan Baston)
   - Improve performance of prepared polygon intersection (GH-690, Dan Baston)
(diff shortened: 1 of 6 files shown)


dbaston:
This also brings performance of prepared zero-distance check ahead of prepared intersection testing. I think this essentially means `IndexedFacetDistance` is outperforming `MCIndexSegmentSetMutualIntersector` for this dataset.

pramsey:
Good get!

PR #699, opened 2022-10-11
Fix LineSegment.orientationIndex(LineSegment)

dr-jts:
Fix `LineSegment.orientationIndex(LineSegment)` to have correct result for the case of a non-collinear segment on the right side of the target segment.

Port of https://github.com/locationtech/jts/pull/914


strk:
> Love a two character fix...

Meaning there's no unit test ? :)

pramsey:
There's several! But the fix is two-char.

PR #701, opened 2022-10-13
Add CoordinateXY type

dbaston:
This PR adds a `CoordinateXY` type and uses it throughout portions of the `algorithm` namespace that only work in 2D.

The current `Coordinate` type is changed to inherit from `CoordinateXY`. There are no virtual methods, so there is no penalty compared to the current implementation.


dbaston:
Any objections?

dr-jts:
Seems reasonable, as a step along the path to supporting multiple coordinate dimensions.

Issue #706, opened 2022-10-17
Implement LineSubstring


caspervdw:
At shapely we would like to have a "LineSubstring" method. We now have a Python-implemented one, but an implementation by GEOS would be more efficient.

See also:
https://shapely.readthedocs.io/en/stable/manual.html#shapely.ops.substring

Related work:
- https://postgis.net/docs/manual-1.5/ST_Line_Substring.html
- https://github.com/pygeos/pygeos/issues/352

@jorisvandenbossche

dbaston:
It sounds like exposing `LengthIndexedLine::extractLine` in the C API would meet this need?

https://github.com/libgeos/geos/blob/cc667261086d8e5f9906b69f4feaea5112221e92/include/geos/linearref/LengthIndexedLine.h#L96-L107

PR #708, opened 2022-10-18
Add GEOSLineSubstring

dbaston:
Resolves https://github.com/libgeos/geos/issues/706

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ xxxx-xx-xx
   - Add CoordinateSequenceIterator (GH-685, Dan Baston)
   - Geometry clustering: DBSCAN, geometry intersection/distance, envelope
     intersection/distance (GH-688, Dan Baston)
+  - CAPI: GEOSLineSubstring (GH-706, Dan Baston)
 
 - Fixes/Improvements:
   - WKTReader: Fix parsing of Z and M flags in WKTReader (#676 and GH-669, Dan Baston)
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -912,6 +912,12 @@ extern "C" {
     }
 
     Geometry*
+    GEOSLineSubstring(const Geometry* g, double start_fraction, double end_fraction)
+    {
+        return GEOSLineSubstring_r(handle, g, start_fraction, end_fraction);
+    }
+
+    Geometry*
     GEOSReverse(const Geometry* g)
     {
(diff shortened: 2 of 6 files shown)


dr-jts:
Is this good to merge?

dbaston:
I had a few questions.

I just wrote tests for the current behavior in edge cases, but I don't know if that behavior is desirable. In particular, I'm not sure silently clamping out-of-range length fractions or returning zero-length linestrings are good ideas. If we want to change the behavior, the time would be now, before we expose it in the C API.

Also not sure if anyone would prefer an interface like Shapely's where a flag indicates whether the arguments should be interpreted as length fractions or absolute lengths. I would opt for leaving it as-is.

dr-jts:
> not sure silently clamping out-of-range length fractions 

I'm +0 on whether to silently clamp or throw an exception.  I'd like to leave the option open to pass fractions outside the range [0,1] to allow extending a LineString.

> or returning zero-length linestrings 

I would like to see zero-length results be supported, since otherwise that's a corner-case that needs to be checked.  (IMO the SFS should have made zero-length lines valid)

> not sure if anyone would prefer an interface like Shapely's where a flag indicates whether the arguments should be interpreted as length fractions or absolute lengths. I would opt for leaving it as-is.

+1 to leaving as-is.

dbaston:
> I would like to see zero-length results be supported

Yes, the question is whether to follow PostGIS and emit a Point in this case.

jorisvandenbossche:
Thanks for the PR!

From a first look, I was also going to mention that in shapely we allow passing the start/end both as distance or as fraction. But, this is very easy to support on the shapely side (converting the distance to a fraction). So I think it is certainly OK to only support one option here. 
As prior art, the interpolate (ST_LineInterpolatePoint) has a `GEOSInterpolate` and `GEOSInterpolateNormalized` version (for distance and fraction, respectively)

It seems the current version in shapely also supports negative values to start measuring from the end. But again that is easily to keep supporting on the shapely side if we want that.

> or returning zero-length linestrings

The shapely version currently returns a Point in that case, but I would personally prefer what you did here (LineString with identical start and end point). It seems easier to work with if it just always returns a LineString.

As another corner case, what happens here if start > end ?

This works with both LineString and MultiLineString I see in the tests? (the PostGIS one explicitly mentions only linestrings) That would be nice to support (the shapely version currently also only supports LineStrings)

PR #709, opened 2022-10-18
Use set_target_properties to set output directories for geos and geos_c.

Civitasv:
Use `set_target_properties` to set `LIBRARY_OUTPUT_DIRECTORY`, `ARCHIVE_OUTPUT_DIRECTORY` and `RUNTIME_OUTPUT_DIRECTORY`, in this way, geos will not interfere output settings of users' project.

~Use macro `GEOS_ADD_TARGET` to add a target.~

Use `set_target_properties` to set output attrs for geos and geos_c.

```cmake
set_target_properties(geos geos_c
  PROPERTIES
  LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/lib"
  ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/lib"
  RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/bin")
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,18 +144,6 @@ if (POLICY CMP0092)
 endif()
 
 #-----------------------------------------------------------------------------
-# Setup build directories
-#-----------------------------------------------------------------------------
-# Place executables and shared libraries in the same location for
-# convenience of direct execution from common spot and for
-# convenience in environments without RPATH support.
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib)
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
-message(STATUS "GEOS: Run-time output: ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
-message(STATUS "GEOS: Archives output: ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
-
-#-----------------------------------------------------------------------------
 # Install directories
 #-----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


dbaston:
Is there a way to apply this to all of our targets without copy-paste? (benchmarks, tests, etc.)

hobu:
> Is there a way to apply this to all of our targets without copy-paste?

For PDAL we have macros that roll up all of these kind of settings for the different categories of targets. See https://github.com/PDAL/PDAL/blob/master/cmake/macros.cmake for inspiration.

Civitasv:
> Is there a way to apply this to all of our targets without copy-paste? (benchmarks, tests, etc.)

I think it can be achived by put all targets into a variable. Maybe there's a better way.

Civitasv:
> I think it can be achived by put all targets into a variable. Maybe there's a better way.

I have done that. Please consider agree my pull request.

CMakeLists.txt
```diff
@@ -169,6 +157,15 @@ message(STATUS "GEOS: Require C++${CMAKE_CXX_STANDARD}")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+#-----------------------------------------------------------------------------
+# Targets variable
+#-----------------------------------------------------------------------------
+set(_targets "")
+
+macro(GEOS_ADD_TARGET _name)
```
pramsey, line 165: Given there's only two items in this list, I find the indirection actually more harmful than helpful... when looking at the set_target_properties call I have to go back and figure out "now what targets are being set, exactly?".

Civitasv: The reason I change it to `set_target_properties` is that it won't pollute other projects' setting of output directories, including library, runtime, archive.
And yes it's not clearer than before, but I think the reason above is more important.
About the question `now what targets are being set, exactly?`, I think it may be better to include all targets?

CMakeLists.txt
```diff
@@ -300,6 +301,24 @@ endif()
...
 add_subdirectory(capi)
 
+#-----------------------------------------------------------------------------
+# Setup build directory for targets: output directories of
+# library, archive and runtime.
+# Place executables and shared libraries in the same location for
+# convenience of direct execution from common spot and for
+# convenience in environments without RPATH support.
+#-----------------------------------------------------------------------------
+set_target_properties(${_targets}
```
pramsey, line 311: This is fine, I just mean dispense with the GEOS_ADD_TARGET machinery, and just call `set_target_properties(geos geos_c ... )` without the indirection of the _targets variables.

Civitasv: Yeah, fine by me. I'll delete `GEOS_ADD_TARGET` for now.

pramsey:
If you rebase to main do we get a green CI?

Civitasv:
> If you rebase to main do we get a green CI?

I've done that.

dbaston:
This PR addresses `geos_c` and `geos` but we have many more targets for benchmarks, tests, and utilities (geosop). I would prefer not to have all of these moved.

Civitasv:
We can add more targets into it.

PR #711, opened 2022-10-21
Use CoordinateXY type for TopologyValidationError

dbaston:


dr-jts:
Validity is only checked in the XY plane.  So it's actually more appropriate if only XY is used for the location report.

dbaston:
I agree that 2D is better but the longstanding output is not _wrong_, so it doesn't seem desirable to change it.

pramsey:
I guess, but it's also only a message? Hard call. Like you said on the PostGIS ticket, at some level downstream shouldn't be spending a lot of effort testing the underlying library at very fine granularity?

dbaston:
OK, so change the message and change the  `TopologyValidationError` class to use the `CoordinateXY` type?

dr-jts:
> OK, so change the message and change the `TopologyValidationError` class to use the `CoordinateXY` type?

+1

pramsey:
Yes, onwards to our glorious future!

PR #716, opened 2022-10-25
Add OverlayNG support for simple GeometryCollection inputs

dr-jts:
Adds support for simple GeometryCollection inputs to OverlayNG.

A simple GeometryCollection is one which can be flattened to a valid Multi-geometry; i.e. it is homogeneous and has no overlapping Polygon compents.

Port of https://github.com/locationtech/jts/pull/915

### DIFF
--- a/src/operation/overlayng/OverlayMixedPoints.cpp
+++ b/src/operation/overlayng/OverlayMixedPoints.cpp
@@ -41,7 +41,7 @@ using namespace geos::geom;
  */
 class CoordinateExtractingFilter: public geom::CoordinateFilter {
 public:
-    CoordinateExtractingFilter(std::unique_ptr<CoordinateArraySequence>& p_pts, const PrecisionModel* p_pm)
+    CoordinateExtractingFilter(CoordinateArraySequence& p_pts, const PrecisionModel& p_pm)
         : pts(p_pts), pm(p_pm)
     {}
 
@@ -62,13 +62,13 @@ public:
     filter_ro(const geom::Coordinate* coord) override
     {
         Coordinate p(*coord);
-        pm->makePrecise(p);
-        pts->add(p);
+        pm.makePrecise(p);
+        pts.add(p);
     }
 
 private:
-    std::unique_ptr<CoordinateArraySequence>& pts;
-    const PrecisionModel* pm;
+    CoordinateArraySequence& pts;
+    const PrecisionModel& pm;
 };
 
 /*public*/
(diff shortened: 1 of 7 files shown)


src/operation/overlayng/OverlayMixedPoints.cpp
```diff
@@ -29,15 +29,48 @@
...
+    void
+    filter_ro(const geom::Coordinate* coord) override
+    {
+        Coordinate p(*coord);
+        pm->makePrecise(p);
+        pts->add(p);
+    }
+
+private:
+    std::unique_ptr<CoordinateArraySequence>& pts;
```
dbaston, line 70: Just store a `CoordinateArraySequence&` here.

src/operation/overlayng/OverlayMixedPoints.cpp
```diff
@@ -234,6 +267,10 @@ std::unique_ptr<CoordinateArraySequence>
 OverlayMixedPoints::extractCoordinates(const Geometry* points, const PrecisionModel* p_pm) const
 {
     std::unique_ptr<CoordinateArraySequence> coords(new CoordinateArraySequence());
+
+    CoordinateExtractingFilter filter(coords, p_pm);
+    points->apply_ro(&filter);
+    /*
```
dbaston, line 273: Remove commented-out code?

src/operation/overlayng/OverlayPoints.cpp
```diff
@@ -27,6 +27,47 @@ namespace geos {      // geos
...
+        * This provides the merging semantics of overlay
+        */
+        if (ptMap.find(p) == ptMap.end()) {
+            std::unique_ptr<Point> newPt(pt->getFactory()->createPoint(p));
+            ptMap[p] = std::move(newPt);
+        }
+    }
+
+    Coordinate
+    roundCoord(const Point* pt, const PrecisionModel* p_pm) const
```
dbaston, line 57: Looks `static` ?

src/operation/overlayng/OverlayPoints.cpp
```diff
@@ -27,6 +27,47 @@ namespace geos {      // geos
...
+struct PointExtractingFilter: public GeometryComponentFilter {
+
+    PointExtractingFilter(std::map<Coordinate, std::unique_ptr<Point>>& p_ptMap, const PrecisionModel* p_pm)
+        : ptMap(p_ptMap), pm(p_pm)
+    {}
+
+    void
+    filter_ro(const Geometry* geom)
+    {
+        const Point* pt = dynamic_cast<const Point*>(geom);
```
dbaston, line 39: Faster to use `geom->getGeometryTypeId()`, if that matters.

dr-jts: Good to use the best-practice pattern - I'll change this.

dr-jts: And then `static_cast<const Point*>(geom)`?

src/operation/overlayng/OverlayMixedPoints.cpp
dr-jts: So change the constructor to take a `CoordinateArraySequence&`?  if so, how should the client 
's `std::unique_ptr<CoordinateArraySequence>` be passed?

dbaston: Yes, the constructor can take `CoordinateArraySequence&` and the caller can construct it like `CoordinateExtractingFilter filter(*coords, p_pm)`

dr-jts: ah, ok, I see that it can be called with `*seq`.

src/operation/overlayng/OverlayPoints.cpp
dbaston: Yes

PR #727, opened 2022-11-08
Improve short-circuits in SubgraphDepthLocater

dbaston:
- Remove a seemingly incorrect short-circuit that skips a buffer subgraph if it lies to the right of a rightward-extending ray
- Add envelope short-circuits for individual DirectedEdges

Improves runtime of Australia buffer benchmark by 75%.

```
bin/geosop -q -t -a ~/data/australia.txt buffer 1e-3
```

### DIFF
--- a/src/operation/buffer/SubgraphDepthLocater.cpp
+++ b/src/operation/buffer/SubgraphDepthLocater.cpp
@@ -183,10 +183,9 @@ SubgraphDepthLocater::findStabbedSegments(const Coordinate& stabbingRayLeftPt,
 
         // optimization - don't bother checking subgraphs
         // which the ray does not intersect
-        Envelope* env = bsg->getEnvelope();
+        const Envelope* env = bsg->getEnvelope();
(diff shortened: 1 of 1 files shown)


dbaston:
The short-circuit that I propose partially removing was added here:
https://github.com/libgeos/geos/commit/aebb85462c22c1b6c62a9c85efd67eca53579f63

The test file referenced in that commit message is still in the repo under a different name: https://github.com/libgeos/geos/blob/main/tests/xmltester/tests/misc/Buffer-2.xml

pramsey:
Wow? 🌞

dbaston:
I'm not seeing a difference on less pathological buffers. But if you're someone who takes small buffers of large countries, I guess it's a "wow" 😀

Does this look OK, @dr-jts ? I'm not very familiar with the buffer internals.

dr-jts:
This change looks ok to me (especially the bug fix to the original optimization patch!)

dr-jts (approved):

Issue #606, opened 2022-05-13
heap-use-after-free in /src/geos/src/geomgraph/PlanarGraph.cpp


CCWANG19:
###### Build platform
ubuntu20.04

###### Build steps
```
CC = clang
CFLAGS=-O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link

cmake -DBUILD_SHARED_LIBS=OFF ..
make 
```

###### Test case
[poc.zip](https://github.com/libgeos/geos/files/8686576/poc.zip)

###### Execution steps
./fuzz_geo2 poc

###### Output
```
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3281110018
INFO: Loaded 1 modules   (57746 inline 8-bit counters): 57746 [0xc1e520, 0xc2c6b2), 
INFO: Loaded 1 PC tables (57746 PCs): 57746 [0xc2c6b8,0xd0dfd8), 
./fuzz_geo2: Running 1 inputs 1 time(s) each.
Running: poc
=================================================================
==13391==ERROR: AddressSanitizer: heap-use-after-free on address 0x60e000000740 at pc 0x00000068ff62 bp 0x7ffe641c7a60 sp 0x7ffe641c7a58
READ of size 8 at 0x60e000000740 thread T0
    #0 0x68ff61 in geos::geomgraph::PlanarGraph::~PlanarGraph() /src/geos/src/geomgraph/PlanarGraph.cpp:88:9
    #1 0x7ed98a in geos::operation::overlay::OverlayOp::~OverlayOp() /src/geos/src/operation/overlay/OverlayOp.cpp:184:1
    #2 0x7ec54b in geos::operation::overlay::OverlayOp::overlayOp(geos::geom::Geometry const*, geos::geom::Geometry const*, geos::operation::overlay::OverlayOp::OpCode) /src/geos/src/operation/overlay/OverlayOp.cpp:94:1
    #3 0x62fe29 in geos::geom::HeuristicOverlay(geos::geom::Geometry const*, geos::geom::Geometry const*, int) /src/geos/src/geom/HeuristicOverlay.cpp:446:19
    #4 0x616c24 in geos::geom::Geometry::Union(geos::geom::Geometry const*) const /src/geos/src/geom/Geometry.cpp:599:12
[...]

PR #730, opened 2022-11-08
Fix double-free when Node::add throws

dbaston:
Fixes https://github.com/libgeos/geos/issues/606

### DIFF
--- a/tests/unit/capi/GEOSUnionTest.cpp
+++ b/tests/unit/capi/GEOSUnionTest.cpp
@@ -43,5 +43,24 @@ void object::test<1>()
     GEOSGeom_destroy(expected);
 }
 
+// Verify that no crash occurs in overlay with NaN coordinates
+// https://github.com/libgeos/geos/issues/606
+template<>
+template<>
+void object::test<2>()
+{
+    std::string wkb1 = "010100000000000000000000000000000000000840";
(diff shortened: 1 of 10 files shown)


tests/unit/capi/GEOSUnionTest.cpp
```diff
@@ -43,5 +43,22 @@ void object::test<1>()
...
+    std::string wkb2 = "01020000000300000049544C553736090000FFFF544E494F500000000000000000FFFFFF2B2B2B2B2BFFFFFFFFFFFFFFFF00FFFFFFFFFFFFFF";
+
+    geom1_ = GEOSGeomFromHEX_buf((unsigned char*) wkb1.c_str(), wkb1.size());
+    geom2_ = GEOSGeomFromHEX_buf((unsigned char*) wkb2.c_str(), wkb2.size());
+
+    ensure(geom1_);
+    ensure(geom2_);
+
+    GEOSGeometry* result = GEOSUnion(geom1_, geom2_);
+    (void) result; // no crash
```
dr-jts, line 60: What does this do?

dbaston: Just trying to show that the union operation doesn't crash. But there is nothing particular to assert about the result.

dbaston:
Unfortunately this fix causes a leak in another test case. It looks like the `Edge*` passed to `Node::add` may or may not be owned by that function, depending on the implementation of the `EdgeEndStar`.

PR #735, opened 2022-11-15
Move readers from geosop into library

dbaston:
This enables them to be used for performance tests and other utilities


dr-jts:
+1

PR #736, opened 2022-11-16
Add STRtree queryPairs method

dbaston:
This PR adds a `queryPairs` method to encapsulate a self-join of items in an `STRtree`. It allows us to push the "don't compare pairs more than once" short-circuit from `MCIndexNoder` (and probably other places) inside the tree, where it can be a bit more effective.

This `queryPairs` implementation only prunes leaves. It seems like it should be possible to prune entire branches, but I haven't yet thought of a nice way to do it without increasing the size of a node.

This causes a failing test from, as far as I can tell, an equally valid output for https://trac.osgeo.org/geos/ticket/586.

image


brendan-ward:
What do you think about exposing this in the C API?  Could certainly be a different PR, just wanted to ask in case here it case it informs the design of this at the C++ level.

For instance, in Shapely, I could see use cases of trying to find other envelope-intersecting geometries in the tree for each geometry in the tree.  I think we'd just need a callback similar to `GEOSQueryCallback` but with a parameter for each side of the pair, plus `void userdata` so that we can pass in a data structure that allows us to accumulate info about those pairs (e.g., use the addresses of the geometries in the pair to resolve those to indices within the tree geometries).

I can't quite follow the C++ implementation, but does this exclude self-pairs?  I.e., is that what is meant by `visit leaf nodes if they have a higher address than the query node`?

dbaston:
> What do you think about exposing this in the C API? 

I don't see a reason not to, if you'd find it useful.

> does this exclude self-pairs

Yes. Hopefully the edit I made to the docstring makes this clear.

dbaston:
Looks like this may have broke the 32-bit builds.

PR #738, opened 2022-11-22
Remove obsolete code for .inl

dg0yt:


mwtoews (approved):
Thanks! This continues from #631

Issue #732, opened 2022-11-11
should this be check_symbol_exists instead of check_prototype_exists?


yurikoif:
https://github.com/libgeos/geos/blob/b602db59b5c5d0ddce1cee7dbe19d920a3359c8f/CMakeLists.txt#L196
should we use in this line check_symbol_exists instead of check_prototype_exists since we are searching in C headers instead of in c++ headers, just like searching for isfinite in line 207?

yurikoif:
created pull request https://github.com/libgeos/geos/pull/742

yurikoif:
> 

Oh sry I got you wrong in the first place. What I intended to say is, like in https://github.com/libgeos/geos/blob/b602db59b5c5d0ddce1cee7dbe19d920a3359c8f/CMakeLists.txt#L207

isn't it looking for a symbol in a header?

pramsey:
Yes, that one seems odd.

pramsey:
History seems to have caught up with this issue, I see neither test in the code base anywhere anymore.

PR #742, opened 2022-11-24
Update CMakeLists.txt

yurikoif:
responding to https://github.com/libgeos/geos/issues/732

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ if(NOT HAVE_STD_ISNAN)
       check_prototype_exists(__inline_isnand math.h HAVE_INLINE_ISNAND_XCODE)
     endif()
   else()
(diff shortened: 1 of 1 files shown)


pramsey:
Like I said, checking a prototype is something I'd expect relative to a header file, and checking a symbol makes sense for a library. Why do you think this is a necessary change (everything already currently works).

rouault:
All that code related to detecting isnan presence is likely obsolete. std::isnan() is a mandatory pat of C++11 in the <cmath> header (https://en.cppreference.com/w/cpp/numeric/math/isnan) and old usage of C isnan() could be converted to std::isnan()

pramsey:
Thanks @rouault that seems like a more fruitful PR... @yurikoif if you want to repurpose this PR to that end, please feel free, otherwise close it out.

PR #745, opened 2022-11-30
Remove raw pointer Geometry ctrs / GeometryFactory  methods

dbaston:
This PR removes remaining usages and implementations of Geometry constructors and GeometryFactory methods dealing in raw pointers instead of managed pointers.

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -2173,7 +2173,6 @@ extern "C" {
         return execute(extHandle, [&]() {
             GEOSContextHandleInternal_t* handle = reinterpret_cast<GEOSContextHandleInternal_t*>(extHandle);
             const GeometryFactory* gf = handle->geomFactory;
-            Geometry* out;
 
             // Polygonize
             Polygonizer plgnzr;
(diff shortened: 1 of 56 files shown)


capi/geos_ts_c.cpp
```diff
@@ -2752,7 +2747,7 @@ extern "C" {
             GEOSContextHandleInternal_t* handle = reinterpret_cast<GEOSContextHandleInternal_t*>(extHandle);
             const GeometryFactory* gf = handle->geomFactory;
 
-            return gf->createPoint(cs);
+            return gf->createPoint(std::unique_ptr<CoordinateSequence>(cs)).release();
```
pramsey, line 2750: Ewwww?

dbaston:
```
#define 🪂 .release()
#define 👉 std::unique_ptr
#define 🧬 CoordinateSequence

return gf->createPoint(👉<🧬>(cs)).🪂
```

pramsey: That is much prettier. However, my implicit question is "why"? It seems completely self-referential to wrap the raw pointer and then release it in one go.

dbaston: Because the C API only deals in raw pointers, so we have to either (a) create and release them at this boundary between C and C++ or (b) maintain a separate set of `GeometryFactory` methods dealing in raw pointers and hope they're not used within the C++ lib itself. This one line is ugly but I think the overall change is strongly net-positive.

pramsey: My apologies, I didn't interpret the `))` correctly. I thought you were doing
```
std::unique_ptr<CoordinateSequence>(cs).release()
```
I see now that the release is a (very important) `)` later.

Kontinuation:
We found that this patch fixed another (probably unreported) problem: `GEOSGeom_createEmptyLineString` creates a 3D geometry instead of 2D. [Commit 708d32b](https://trac.osgeo.org/geos/changeset/708d32b2cd2dd3bedc38ff1d839368656b1f9626/git) attempted to create empty geometries with 2D coordinates by default however that commit did not work for LineStrings.

dbaston:
Thanks for pointing this out, @Kontinuation . It looks like LinearRings are still 3D by default; I changed that in https://github.com/libgeos/geos/pull/819 and added tests for all types.

PR #746, opened 2022-11-30
CI: Add 32-bit build

dbaston:
This PR adds a 32-bit build to GitHub Actions. We previously had a 32-bit build using gcc 4.8 but this was recently removed from the matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,15 @@ jobs:
           cmake: 3.13.*
           os: ubuntu-18.04
 
+        - cxx_compiler: g++-5
+          c_compiler: gcc-5
+          build_type: Release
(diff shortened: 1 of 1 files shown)


pramsey:
Kind of weird that everything goes "boom" all of a sudden... was it really so long ago that we had a 32-bit CI test? Did we think we had a 32-bit CI but actually it secretly wasn't?

dbaston:
I suspect the "boom" came from d210a94fe9b46d4851115f8bb6e8395317f2a027. I noticed it on the 32-bit Jenkins builds, but those don't run until after things are merged.

dbaston:
Merging this now that 32-bit build was fixed by fafb48a0d98a4d782ad0c77363e01ab266ea762a

PR #757, opened 2022-12-07
Single-precision STRtree

dbaston:
This PR generalizes `Envelope` and `Interval` to be constructed from any numeric type, and changes the `IndexedPointInAreaLocator` to use `Interval<float>`. I tested this with `GEOSPreparedContains` because past testing has shown this to be an area where shrinking the size of a tree node has a real benefit. And it does have an impact, but only about 10% in my testing. I'm not sure that's worth it, but maybe there's a more compelling application. Rounding a query envelope from `double` to `float` in a strictly enlarging way seems to pretty expensive.

### DIFF
--- a/include/geos/geom/Envelope.h
+++ b/include/geos/geom/Envelope.h
@@ -103,7 +103,7 @@ public:
      * @param  y1  the first y-value
      * @param  y2  the second y-value
      */
-    void init(double x1, double x2, double y1, double y2)
+    void init(T x1, T x2, T y1, T y2)
     {
         if(x1 < x2) {
             minx = x1;
@@ -206,7 +206,7 @@ public:
      * Returns the Envelope maximum y-value. `min y > max y` indicates
      * that this is a null Envelope.
      */
-    double getMaxY() const
+    T getMaxY() const
     {
         assert(!isNull());
         return maxy;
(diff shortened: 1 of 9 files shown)


include/geos/geom/Envelope.h
```diff
@@ -760,26 +869,36 @@ class GEOS_DLL Envelope {
...
+        maxx = static_cast<float>(e.getMaxX());
+        miny = static_cast<float>(e.getMinY());
+        maxy = static_cast<float>(e.getMaxY());
+
+        if (static_cast<double>(minx) > e.getMinY()) {
+            minx = std::nextafter(minx, -std::numeric_limits<float>::infinity());
+        }
+        if (static_cast<double>(maxx) < e.getMaxY()) {
+            maxx = std::nextafter(maxx, std::numeric_limits<float>::infinity());
+        }
```
kylebarron, line 892: small typo:
```suggestion
        if (static_cast<double>(minx) > e.getMinX()) {
            minx = std::nextafter(minx, -std::numeric_limits<float>::infinity());
        }
        if (static_cast<double>(maxx) < e.getMaxX()) {
            maxx = std::nextafter(maxx, std::numeric_limits<float>::infinity());
        }
```

dbaston:
Doesn't seem worth it for 5-10%.

PR #761, opened 2022-12-07
Add functions to interrupt a single thread

dbaston:
This PR changes `GEOS_interruptRegisterCallback` to operate on a per-thread basis; the callback will only be registered on the thread from which `GEOS_interruptRegisterCallback` is called. This would break existing expectations of a global callback, so maybe a better solution is to add a new function that registers a thread-local callback.

I'm fuzzy on how this gets used outside of PostGIS which (I think?) would not be affected by this change.

cc @nyalldawson

### DIFF
--- a/include/geos/util/Interrupt.h
+++ b/include/geos/util/Interrupt.h
@@ -69,7 +69,7 @@ public:
      * before checking for interruption requests.
      *
      * NOTE that interruption request checking may happen
-     * frequently so the callback shoudl execute quickly.
+     * frequently so the callback should execute quickly.
      *
      * The callback can be used to call Interrupt::request()
      * or Interrupt::requestForCurrentThread().
--- a/include/geos/util/Interrupt.h
+++ b/include/geos/util/Interrupt.h
@@ -27,14 +27,24 @@ class GEOS_DLL Interrupt {
 public:
 
     typedef void (Callback)(void);
+    typedef void (ThreadCallback)(void*);
 
     /**
      * Request interruption of operations
      *
      * Operations will be terminated by a GEOSInterrupt
-     * exception at first occasion.
+     * exception at first occasion, by the first thread
+     * to check for an interrupt request.
      */
     static void request();
+
+    /**
+     * Request interruption of operations in the current thread
+     *
+     * Operations in the current thread will be terminated by
+     * a GEOSInterrupt at first occasion.
+     */
+    static void requestForCurrentThread();
 
     /** Cancel a pending interruption request */
     static void cancel();
@@ -43,16 +53,28 @@ public:
     static bool check();
 
     /** \brief
-     * Register a callback that will be invoked
+     * Register a callback that will be invoked by all threads
      * before checking for interruption requests.
      *
      * NOTE that interruption request checking may happen
-     * frequently so any callback would better be quick.
+     * frequently so the callback should execute quickly.
      *
      * The callback can be used to call Interrupt::request()
-     *
+     * or Interrupt::requestForCurrentThread().
      */
     static Callback* registerCallback(Callback* cb);
+
+    /** \brief
+     * Register a callback that will be invoked the current thread
+     * before checking for interruption requests.
+     *
+     * NOTE that interruption request checking may happen
+     * frequently so the callback shoudl execute quickly.
+     *
+     * The callback can be used to call Interrupt::request()
+     * or Interrupt::requestForCurrentThread().
+     */
+    static ThreadCallback* registerThreadCallback(ThreadCallback* cb, void* data);
 
     /**
      * Invoke the callback, if any. Process pending interruption, if any.
--- a/src/util/Interrupt.cpp
+++ b/src/util/Interrupt.cpp
@@ -16,10 +16,12 @@
 #include <geos/util/GEOSException.h> // for inheritance
 
 namespace {
-/* Could these be portably stored in thread-specific space ? */
+
+// Callback and request status for interruption of any single thread
 bool requested = false;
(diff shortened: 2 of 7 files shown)


nyalldawson:
@jorisvandenbossche 

> (so in between calls to GEOS).

The big benefit of this call is that it can potentially interrupt stupidly long geos internal methods -- ie. when a user tries to intersection a >million vertex polygon with another >million vertex polygon :rofl: . Right now there's no clean way to interrupt those operations without completely killing the whole process.

nyalldawson:
@dbaston 

> I'm fuzzy on how this gets used outside of PostGIS which (I think?) would not be affected by this change.

QGIS currently doesn't use it, because we can't isolate to individual threads. I'd *love* to use it to allow users to cancel long running geos operations though (such as the previously mentioned ridiculous intersection case)

The api for the interruption handling is a little tricky to understand. I gather I register the callback, and then from the callback I call GEOS_interruptRequest() when I want to interrupt a process. Is this correct?

(If so, this may be a nice opportunity to clean up that api and make the callback return bool for when the interruption is desired)

nyalldawson:
> In your case, would you want to register a different callback for each thread? (Does the callback need to be thread-local?) Or would you use a single callback like I did in InterruptTest.cpp to decide if a given thread should be interrupted?

I could work with either. The per-thread/context callback would be cleaner though.

jorisvandenbossche:
> > (so in between calls to GEOS).
> 
> The big benefit of this call is that it can potentially interrupt stupidly long geos internal methods

Yes, and so ideally we should certainly do both (and thus also implement the GEOS interrupt callback). 
The initial motivation of the PR I referenced was to be able to interrupt a call that would create too many geometries (and blow up memory and eventually crash), for example by accidentally creating billions of points (so many calls to a cheap GEOS function). For that we needed to use another mechanism, since (I think) the interrupt callback isn't checked in such cheap GEOS operations.

nyalldawson:
@dbaston I can work with this api :+1:  many thanks!!

dbaston:
> I could work with either. The per-thread/context callback would be cleaner though.

OK, with the current iteration you can interrupt the current thread using `GEOS_interruptThread` called from either a callback that executes on all threads (registered by the existing `GEOS_interruptRegisterCallback`) or a callback that executes on the current thread only (registered by a new `GEOS_interruptRegisterThreadCallback`). There are a few examples in `InterruptTest.cpp` and `GEOSInterruptTest.cpp`.

I used a new callback signature for `GEOS_interruptRegisterThreadCallback`, adding a `void*` so that an interrupt signal can be passed from a coordinating thread.

src/util/Interrupt.cpp
```diff
@@ -37,16 +41,23 @@ Interrupt::request()
     requested = true;
 }
 
+void
+Interrupt::requestForCurrentThread()
```
rouault, line 45: I don't understand how that can work. If GEOS is busy in a thread T1 (processing thread), and you want to interrupt it, you need to do that from another thread T2 (GUI typically). But if in T2 you call Interrupt::requestForCurrentThread(), then it will set the thread_local requested_for_thread for T2, not T1.

rouault: forget above comment... I figured it out

include/geos/util/Interrupt.h
```diff
@@ -27,33 +27,55 @@ class GEOS_DLL Interrupt {
...
+     * or Interrupt::requestForCurrentThread().
      */
     static Callback* registerCallback(Callback* cb);
 
+    /** \brief
+     * Register a callback that will be invoked the current thread
+     * before checking for interruption requests.
+     *
+     * NOTE that interruption request checking may happen
+     * frequently so the callback shoudl execute quickly.
```
rouault, line 72:
```suggestion
     * frequently so the callback should execute quickly.
```

macdrevx:
@dbaston I'm glad to see this work taking shape as it will help to address https://github.com/GEOSwift/GEOSwift/issues/190.

The proposed API lets you register a per-thread callback. I'm curious whether you've considered making it per-context rather than per-thread. My understanding is that GEOS contexts are meant to be confined to a single thread anyway, so tying the interrupt callback to the context seems like it could also solve the issue. It might also fit nicely alongside other existing context-specific APIs like the ones to register error and message handlers.

dbaston:
> I'm curious whether you've considered making it per-context rather than per-thread.

Yes, and that would certainly make for a cleaner API. The problem is that the "context" is a C API construct to which the C++ library (the code actually being interrupted) has no access, so we would need to introduce the notion of a context into the C++ library in such a way that any interruptible code can access its current context. I'm not sure of a simple way to do that, but I'm open to ideas.

dbaston:
Maybe we could have a thread-local global variable in the C++ library that stores the context, and every C API call sets that global variable before calling code in the C++ library?

rouault:
> The problem is that the "context" is a C API construct to which the C++ library (the code actually being interrupted) has no access,

That might have perhaps performance implications, but couldn't that interrupt handler be registered at the C context level, and each time a C _r function is invoked it would automatically install the per-thread interrupt handler before calling the C++ method, and uninstalls it at return / when an exception is caught ?

dbaston:
I think that would also work. Pushing the context object into the C++ library might reduce the penalty because we would be setting/checking one thread-local variable instead of two. I would need to set up a benchmark to see if that actually matters.

dbaston:
See #803 for an alternative API that is based on context handles.

robe2:
same here.  I'm assuming this should go to 3.14 so switching, @dbaston  @pramsey  switch back if you disagree.

PR #765, opened 2022-12-08
Don't initialize CoordinateSequence in constructor

### LINKED ISSUE #764: Don't initialize CoordinateSequence in constructor
# tl;dr

No need to initialize the CoordinateSequence when we construct a new one because we never used to do that.

# The new code (d463bcb6df8fc1f01f276b27719ee8ccfb2d788a)

https://github.com/libgeos/geos/blob/d463bcb6df8fc1f01f276b27719ee8ccfb2d788a/src/geom/CoordinateSequence.cpp#L68-L79

In the code above, we see that if the `CoordinateSequence` is generated by calling with two arguments, `sz` and `dim`, then `initialize` is always called.  It calls `fillVector` which calls `std::fill` and loops through the size of the new vector to initialize all elements to 0.

I think that this is not necessary.

# The old code (60edf0e5f040e065193bd5beb79f7fa8febb1cd5)

Looking at commit 60edf0e5f040e065193bd5beb79f7fa8febb1cd5 we see that creating the CoordinateSequence directly like this is instead of calling `getGeometrySequenceFactory()->create(sz, dim)`.

https://github.com/libgeos/geos/commit/60edf0e5f040e065193bd5beb79f7fa8febb1cd5#diff-95f3875df536135fe955deecfaeab0af5f6af52008d0fd5a56a3ed1b6638aca1L2427-R2429

`getCoordinateSequenceFactory` is a method of the `GeometryFactory` class:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/GeometryFactory.h#L446-L449

It returns a `CoordinateSequenceFactory`, which has a `create` method.  It's here in the old code:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/CoordinateSequenceFactory.h#L87-L98

We see that it is a pure virtual function so it must be implemented by inheritors of this class.

There are only two classes that inherit from this class, the `DefaultCoordinateSequenceFactory` and the `CoordinateArraySequenceFactory`:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/CoordinateArraySequenceFactory.h#L42

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/DefaultCoordinateSequenceFactory.h#L24

Let's look at them one-by-one

# `CoordinateArraySequenceFactory`

Here's the implementation of `create(size_t, size_t)`:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/CoordinateArraySequenceFactory.h#L69-L74

It's creating a `CoordinateArraySequence` using the constructor and returning a `unique_ptr`.  Here's that constuctor:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/src/geom/CoordinateArraySequence.cpp#L37-L42

It is resizing the vector to make it the right size but *not* initializing anything.

# `DefaultCoordinateSequenceFactory`

Here's the implementation of `create(size_t, size_t)`:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/DefaultCoordinateSequenceFactory.h#L43-L53

It's making a `FixedSizeCoordinateSequence`,templated on the number of dimensions.  It's like a fixed size list, like a tuple!  Let's look at that constuctor:

https://github.com/libgeos/geos/blob/0e8d4368b8bd72a7d361286e8523ebce5cff6146/include/geos/geom/FixedSizeCoordinateSequence.h#L35

We see that it is just coping the `dimension`.  It is not initializing anything.

# Conclusion

For all cases of `create(size_t, size_t)`, the code before 60edf0e5f040e065193bd5beb79f7fa8febb1cd5 did not initialize the coordinates.  The new code could also skip the initialization to maintain the same behavior.  This would also improve performance because we'd eliminate the slow call to `std::fill`.

eyal0:
This ought to improve performance in the common case where initialization is unnecessary.

This fixes #764

### DIFF
--- a/src/geom/CoordinateSequence.cpp
+++ b/src/geom/CoordinateSequence.cpp
@@ -75,7 +75,9 @@ CoordinateSequence::CoordinateSequence(std::size_t sz, std::size_t dim) :
     if (dim == 1 || dim > 4) {
         throw util::IllegalArgumentException("Declared dimension must be 2, 3, or 4");
     }
-    initialize();
+    if (sz > 0 && dim == 0) {
+      reinterpret_cast<Coordinate*>(m_vect.data())[0] = geos::geom::Coordinate();
+    }
 }
 
 CoordinateSequence::CoordinateSequence(const std::initializer_list<Coordinate>& list) :


eyal0:
@dbaston Please take a look, thanks.

pramsey:
@dbaston, is this a go or a no?

dbaston:
Yes, see https://github.com/libgeos/geos/blob/4c3bd72ba73e9688b0712d9878d073559c84e4d6/include/geos/geom/CoordinateSequence.h#L85-L96

robe2:
This has been sitting here for over a year are we thinking about doing anything about it @dbaston  @eyal0 @pramsey

pramsey:
Like @dbaston I hate mucking around in the initializers, because the knock-on effects are subtle. I defer to him as usual.

robe2:
I'm just going to close this one out cause it's been sitting here for 2 years and is beginning to seriously annoy me.  I think you are all agreed it's not a good idea.

PR #773, opened 2022-12-12
DouglasPeuckerSimplifier: Allow ring origin to be removed

dbaston:
This PR allows the origin of a ring to be removed during simplification with the Douglas-Peucker algorithm, as discussed in https://github.com/shapely/shapely/issues/1638.

This is performed for all closed LineStrings, but it could be made more restrictive to only operate on LinearRings.


dr-jts:
My thinking is that this should only apply to LinearRings, not to closed LineStrings, since otherwise there is no way to simplify a closed LineString and preserve the endpoints.

PR #774, opened 2022-12-12
PreparedLineStringIntersects: Fix incorrect result for mixed-dim GC with points

dbaston:
@dr-jts JTS appears to be affected by the same issue.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,6 +30,7 @@ xxxx-xx-xx
   - Improve ConvexHull radial sort robustness (GH-724, Martin Davis)
   - Use more robust Delaunay Triangulation frame size heuristic (GH-728, Martin Davis)
   - DiscreteFrechetDistance: Fix crash with empty inputs (GH-751, Dan Baston)
+  - PreparedLineStringIntersects: Fix incorrect result with mixed-dim collection (GH-774, Dan Baston)
 
 
 ## Changes in 3.11.0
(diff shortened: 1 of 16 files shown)


dr-jts:
Historically JTS has not supported predicates on GeometryCollections, including mixed-dimension ones.

One reason for this is the complexity of computing the DE-9IM over collections of overlapping polygons.

It would be nice to extend support for at least `intersects`, since that is relatively easily computable.

dbaston:
PreparedGeometry has some level of longstanding support: https://github.com/locationtech/jts/blob/master/modules/tests/src/test/resources/testxml/general/TestPreparedPredicatesWithGeometryCollection.xml

dr-jts:
> PreparedGeometry has some level of longstanding support: 

Right, I see that now.  I guess it got introduced at some point since it's easier to do with the Prepared Geometry algorithms.  So should be fixed.

dbaston:
> So should be fixed.

Are you OK with this fix?

dr-jts:
As far as I remember (and substantiated by absence of any documentation around support for Prepared Geometry with GeometryCollections), this was more of a "support by happenstance" than a planned extension.  So there may well be other issues in the code.

dr-jts:
Should add unit tests for the new `Geometry.hasDimension` method.

dr-jts:
See matching JTS PR https://github.com/locationtech/jts/pull/944

Issue #766, opened 2022-12-08
BUG: Intersects predicate error for two valid polygons


chuchulk:
``` python
from shapely import wkt
poly1 = wkt.loads('POLYGON ((26639.240191093646 6039.3615818717535, 26639.240191093646 5889.361620883223,28000.000095100608 5889.362081553552, 28000.000095100608 6039.361620882992, 28700.000190214026039.361620882992, 28700.00019021402 5889.361822800367, 29899.538842431968 5889.362160452064,32465.59665091549 5889.362882757903, 32969.2837182586 -1313.697771558439, 31715.832811969216 -1489.87008918589, 31681.039836323587 -1242.3030298361555, 32279.3890331618 -1158.210534269224, 32237.63710287376 -861.1301136466199, 32682.89764107368 -802.0828534499739, 32247.445200905553 5439.292852892075, 31797.06861513178 5439.292852892075, 31797.06861513178 5639.36178850523, 29899.538849750803 5639.361268079038, 26167.69458275995 5639.3602445643955, 26379.03654594742 2617.0293071870683, 26778.062167926924 2644.9318977193907, 26792.01346261031 2445.419086759444, 26193.472956813417 2403.5650586598513, 25939.238114175267 6039.361685403233, 26639.240191093646 6039.3615818717535), (32682.89764107368 -802.0828534499738, 32682.89764107378 -802.0828534499669, 32247.445200905655 5439.292852892082, 32247.445200905553 5439.292852892075, 32682.89764107368 -802.0828534499738))')
poly2 = wkt.loads('POLYGON ((32450.100392347143 5889.362314133216, 32050.1049555691 5891.272957209961, 32100.021071878822 16341.272221116333, 32500.016508656867 16339.361578039587, 32450.100392347143 5889.362314133216))')
poly1.intersects(poly2)
```

When I perform the above `intersects` operation,both polygons are valid, but the line of `poly1.intersects(poly2)` raise `TopologyException` and tell me the input set may be is invalid.
I am doing with Shapely 2.0b2 / GEOS 3.5

dbaston:
I can confirm this in `main` using `geosop`, which shows both inputs are valid but throws a `TopologyIntersection` during intersection testing.

dbaston:
JTS shows that the first input is invalid, so the error may be in `isValid` rather than `intersects`.

dr-jts:
The format of the first WKT is invalid, since there's a missing space between two numbers:
```
28700.000190214026039.361620882992
[...]

PR #775, opened 2022-12-13
Delegate Geometry::intersects to PreparedGeometry::intersects

dbaston:
This PR changes `Geometry::intersects` and `Geometry::disjoint` to use `PreparedGeometry` instead of `RelateOp`. Relative to the `RelateOp` implementation, the prepared intersection is more robust (e.g., https://github.com/libgeos/geos/issues/766) and is faster for both complex and simple polygons, even if the object is only used a single time.

A quick benchmark shows the performance difference in cases of relatively complex polygons (left) and triangles (right), using the GEOS `perf_intersection`, which combines index querying, intersects testing, and intersection overlay.

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@ xxxx-xx-xx
   - WKTReader: Throw exception on inconsistent geometry dimension (#1080, Dan Baston)
   - WKTReader: Throw exception if WKT contains extra text after end of geometry (#1095, Dan Baston)
   - GEOSIntersects: Fix crash with empty point inputs (#1110, Dan Baston)
+  - GEOSIntersects: Improve performance/robustness by using PreparedGeometry algorithm (GH-775, Dan Baston)
   - LineMerger: Recursively collect all components from GeometryCollections (#401, Dan Baston)
   - GeometryPrecisionReducer: Return correct dimensionality for empty results (GH-684, Dan Baston)
   - Improve performance of coverage union (GH-681, Dan Baston)
--- a/include/geos/index/strtree/Interval.h
+++ b/include/geos/index/strtree/Interval.h
@@ -17,6 +17,7 @@
 #include <geos/export.h>
 #include <algorithm>
 #include <cassert>
+#include <cmath>
 
 namespace geos {
 namespace index { // geos::index
@@ -29,7 +30,7 @@ namespace strtree { // geos::index::strtree
 class GEOS_DLL Interval {
 public:
     Interval(double newMin, double newMax) : imin(newMin), imax(newMax) {
-        assert(imin <= imax);
+        assert(std::isnan(newMin) || std::isnan(newMax) || imin <= imax);
     }
 
     double getMin() const { return imin; }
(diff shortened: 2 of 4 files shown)


dr-jts:
Nice!

Does contains/within show a similar performance boost?

dbaston:
I would expect an improvement, but I need to put together some suitable geometries for testing. Will do as a separate PR.

dr-jts:
`covers` and `coverdBy` might show even more improvement than `contains`/`within`, since they are simpler predicates.

dbaston:
So far I'm not seeing any improvement on "contains" in the initial test case (P/P with 13% of contains tests returning true). Will continue looking into it.

image

dr-jts:
You should test with Points and small (2-point) LineStrings as well, to ensure there is no performance regression in those cases (or if there is, that it's acceptable).

dr-jts:
@dbaston In JTS I'm seeing that `PreparedPolygon` is slower in single-use mode than `RelateOp` for a test with  target polygons of 1000 and 2000 pts, against a set of 1000 lines with 100 pts each:
```
-------  Running with polygon size = 10
runIntersectsNonPrep : 1921 ms
runIntersectsPrepCached : 229 ms
runIntersectsPrepNotCached : 352 ms

-------  Running with polygon size = 100
runIntersectsNonPrep : 1814 ms
runIntersectsPrepCached : 64 ms
runIntersectsPrepNotCached : 495 ms

-------  Running with polygon size = 1000
runIntersectsNonPrep : 1253 ms
runIntersectsPrepCached : 23 ms
runIntersectsPrepNotCached : 1485 ms

-------  Running with polygon size = 2000
runIntersectsNonPrep : 1300 ms
runIntersectsPrepCached : 22 ms
runIntersectsPrepNotCached : 2577 ms
```
I'm pretty sure that I carried out this kind of analysis a while ago, and decided that the tradeoff was a difficult call to make.  Perhaps what's required is a switch based on the size of the target geometry.

dr-jts:
Also, I expect that a new implementation of `RelateOp` (`RelateNG` probably) will be faster than `PreparedGeometry`, since it can choose to only index linework which might potentially interact (i.e that lies inside the intersection of the input envelopes).  It will be more robust as well.

That in itself probably doesn't weigh against this PR, since this change can be backed out when `RelateNG` arrives.  But the performance regression for large geometries requires some thought.

dbaston:
Can you point me to those benchmarks?

dbaston:
> In JTS I'm seeing that PreparedPolygon is slower in single-use mode than RelateOp for a test with target polygons of 1000 and 2000 pts, against a set of 1000 lines with 100 pts each:

I ported your benchmark and I see the same phenomenon trend in GEOS, although RelateOp doesn't become faster there are until around 3,000 points in the sine star.

dbaston:
An "incrementally prepared geometry" resolves the issue. Too hacky?

```
algorithm::locate::PointOnGeometryLocator*
PreparedPolygon::
getPointLocator() const
{
    if(! ptOnGeomLoc) {
        ptOnGeomLoc.reset(new algorithm::locate::SimplePointInAreaLocator(&getGeometry()));

        return ptOnGeomLoc.get();
    } else if (!indexedPtOnGeomLoc) {
        indexedPtOnGeomLoc.reset(new algorithm::locate::IndexedPointInAreaLocator(getGeometry()));
    }

    return indexedPtOnGeomLoc.get();
}
```

dr-jts:
> An "incrementally prepared geometry" resolves the issue. Too hacky?

Very clever!  Looks good to me.  A line or two of explanatory doco would be nice.

PR #777, opened 2022-12-15
PreparedPolygonIntersects: Use SimplePointInAreaLocator for first test

dbaston:
This provides better performance in cases where the prepared polygon is only used for one test.

I think it would be possible to take this a bit further and tell the `PreparedPolygonFactory` how many times the prepared geometry is expected to be queried. This hint could then be used for some more sophisticated algorithm selection. But this is a good improvement for now.

With this PR, the single-use prepared geometry used by `Geometry::intersects` outperforms `RelateOp` for all scenarios in the JTS benchmark.

image

### DIFF
--- a/include/geos/geom/prep/PreparedPolygon.h
+++ b/include/geos/geom/prep/PreparedPolygon.h
@@ -53,6 +53,7 @@ private:
     bool isRectangle;
     mutable std::unique_ptr<noding::FastSegmentSetIntersectionFinder> segIntFinder;
     mutable std::unique_ptr<algorithm::locate::PointOnGeometryLocator> ptOnGeomLoc;
+    mutable std::unique_ptr<algorithm::locate::PointOnGeometryLocator> indexedPtOnGeomLoc;
     mutable noding::SegmentString::ConstVect segStrings;
     mutable std::unique_ptr<operation::distance::IndexedFacetDistance> indexedDistance;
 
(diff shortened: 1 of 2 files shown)


dbaston:
@pramsey this might offer a pretty good benefit to PostGIS `ST_Intersects` which IIRC always prepares its input.

nyalldawson:
> always prepares its input

Is this a recommended practice?

dbaston:
> Is this a recommended practice?

For released versions of GEOS I would say say if you are intersecting Polygon A against many polygons the answer is "yes." If you are only intersecting Polygon A with Polygon B, the answer is "probably."

With #775 (and hopefully related follow-ups) GEOS will just use the most efficient algorithm when you call `GEOSIntersects` etc, so I would say "no."

nyalldawson:
@dbaston 

So my take away from that would be that clients should only manually prepare when they know > 1 test will be made, and otherwise just trust geos (and things will get magically faster with newer geos). Does that sound fair?

dbaston:
> Does that sound fair?

Yes

dbaston:
Here's the result of an experiment to see how many PIP tests you need to offset the cost of creating an `IndexedPointInAreaLocator`. If would be nice to pass `PreparedGeometryFactory` a hint about the number of expected queries so it could use a relationship like this to select the best algorithm.

image

nyalldawson:
@dbaston that's a very interesting result! The payoff threshold is certainly a lot higher than what I'd anticipated

PR #778, opened 2022-12-22
DOC: GEOSEquals place -> plane

visr:


mwtoews (approved):
Thanks for the correction!

Issue #515, opened 2021-11-28
Floating point errors


caspervdw:
In https://github.com/libgeos/geos/pull/485 a large amount of floating point errors were solved. However there are some that remained. I listed the ones that popped up in pygeos (with GEOS 3.10.1):

- binary predicates on empty geometries
- voronoi_polygons on a point
- hausdorff_distance (in our seemingly normal test)
- offset_curve on an empty linestring

All through the CAPI. See pygeos ​https://github.com/pygeos/pygeos/pull/441 for the specific tests where I encountered the FP “invalid” flags.

pramsey:
I'm not actually seeing fe errors when mimicking these tests...
```
template<>
template<>
void object::test<5> ()
{
    runTest(
        "LINESTRING (0 0, 100 0, 10 100, 10 100)",
        "LINESTRING (0 100, 0 10, 80 10)", 0.001, 47.89);

    // check for floating point overflow exceptions
    int raised = fetestexcept(FE_OVERFLOW);
    ensure_equals(raised & FE_OVERFLOW, 0);
}
```

caspervdw:
Thanks for looking into this issue. Sorry I didn’t provide a reproducing code snippet, I am not so comfortable writing in C++.

Looking at your example, I think you might reproduce the issue if you check for `FE_INVALID` or `FE_ALL_EXCEPT` flags. I

pramsey:
Yes, it's FE_INVALID, but I cannot track down where it is being set.

pramsey:
Even when I track it down, the placement helps not at all. Somehow this is an overflow?
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_ARITHMETIC (code=EXC_I386_SSEEXTERR, subcode=0x20)
    frame #0: 0x0000000103948014 libgeos.3.11.0.dylib`geos::geom::LineSegment::projectionFactor(this=0x00007ffeefbfeb70, p=0x00000001051b3960) const at LineSegment.cpp:79:56
   76  	    double dx = p1.x - p0.x;
   77  	    double dy = p1.y - p0.y;
   78  	    double len2 = dx * dx + dy * dy;
-> 79  	    double r = ((p.x - p0.x) * dx + (p.y - p0.y) * dy) / len2;
   80  	    return r;
   81  	}
   82  	
Target 0: (test_geos_unit) stopped.
(lldb) p len2
(double) $0 = 8100
(lldb) 
```

pramsey:
[...]

PR #790, opened 2023-01-07
Fix floating point invalid operations in Envelope (FE_INVALID)

jorisvandenbossche:
See https://github.com/libgeos/geos/issues/515 (as previous similar issue) and https://github.com/shapely/shapely/issues/1709 for context. 
Summary: in shapely 2.0, we end up catching any floating point flag set by GEOS and turn those into warnings (it's actually numpy that does that under the hood), and so that also generates warnings in cases that shouldn't warn (e.g. where there are no actual NaN coordinates from user input).

This PR only has the changes that I needed to make to get one small test passing (intersection of two non-overlapping rectangles which thus gives empty polygon as result), and so is certainly not complete at all. 
But, I also noticed that I am basically undoing some changes from https://github.com/libgeos/geos/pull/407 (which was done for performance reasons, cc @dbaston), so I assume this warrants some discussion whether changes like this PR are welcomed or if there would be another solution.


tests/unit/capi/GEOSIntersectionTest.cpp
```diff
@@ -131,6 +132,36 @@ void object::test<6>
...
+    geom2_ = GEOSGeomFromWKT("POLYGON ((1 2, 1 3, 0 3, 0 2, 1 2))");
+
+    ensure(geom1_);
+    ensure(geom2_);
+
+    // clear all floating point exceptions
+    feclearexcept (FE_ALL_EXCEPT);
+
+    int old_excepts = fegetexcept(); /* Store old exceptions */
+    feenableexcept(FE_INVALID);
```
jorisvandenbossche, line 150: This is only for debugging, and can be removed in the end, I suppose (but for debugging, it enabled to run GDB on this test and stop at all places that give an issue)

jorisvandenbossche:
Closing in favor of https://github.com/libgeos/geos/pull/791

PR #791, opened 2023-01-08
Avoid FE_INVALID on operations with empty Envelopes

dbaston:
This PR is an alternative approach to #790 that replaces comparison operators in `Envelope` with the C99 comparison functions that provide an equivalent result but do not set `FE_INVALID`.

@jorisvandenbossche 
In the process of putting this together I noticed that `Envelope::distance` sets `FE_INEXACT` in some of the test cases. I am guessing that one is not a problem for you?


jorisvandenbossche:
Ah, cool, I didn't know those functions where the "quiet" version of the operators. Thanks for taking a look!

> In the process of putting this together I noticed that `Envelope::distance` sets `FE_INEXACT` in some of the test cases. I am guessing that one is not a problem for you?

The NumPy reference (https://numpy.org/doc/stable/reference/generated/numpy.seterr.html) only speaks about division by zero, overflow, underflow and invalid errors, and so doesn't mention the fifth exception (inexact). So I suppose this is fine (given that inexact results are common with floats, it seems logical to not catch that one)

dbaston:
Thanks for debugging the issue and testing the fix, @jorisvandenbossche .

I've backported the fix to 3.11 and 3.10, which should be the only affected series.

PR #792, opened 2023-01-10
WKTWriter: Output dimension flags for empty geometries

dbaston:
These flags were omitted by design in https://trac.osgeo.org/geos/changeset/2998. I'm not sure what the motivation was for that. Including them makes GEOS output consistent with PostGIS.

Resolves https://trac.osgeo.org/geos/ticket/1129

### DIFF
--- a/src/io/WKTWriter.cpp
+++ b/src/io/WKTWriter.cpp
@@ -211,12 +211,19 @@ WKTWriter::appendGeometryTaggedText(const Geometry& geometry,
                                     int p_level,
                                     Writer& writer) const
 {
-    // evaluate the ordinates actually present in the geometry
-    CheckOrdinatesFilter cof(checkOrdinates);
-    geometry.apply_ro(cof);
+    OrdinateSet outputOrdinates = OrdinateSet::createXY();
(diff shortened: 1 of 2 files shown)


pramsey (approved):
Seems fine to me. Dimensioned empty is kind of ugly but is a "thing". Speaking of which if A.disjoint(B), and A and B are Z-dimensioned, is A.intersection(B) a Z EMPTY?

dbaston:
> and A and B are Z-dimensioned, is A.intersection(B) a Z EMPTY?

I just checked. Apparently the result is 2D.

pramsey:
Funky... I guess there is a shortcut somewhere that is building a default (non-dimensioned) empty when the disjointness is noticed. Presumably a non-disjoint result would have the extra dimension, since we have all the machinery for interpolating the Z value for introduced vertices.

jorisvandenbossche:
+1 on this change. Shapely currently has some custom code to "fix" this specific case and already write ".. Z EMPTY" wkt (https://github.com/shapely/shapely/blob/9105bd323bb255f0401ae464446d85cce284ef2f/src/geos.c#L307-L369)

Issue #704, opened 2022-10-17
Allow passing custom memory management functions


caspervdw:
Providing custom memory management functions would allow dependent libraries to use their memory management policies. Especially when small amount of bytes are repeatedly allocated/deallocated, may show a decent speedup.

Also out-of-memory situations may be handled more gracefully (specifically for pygeos/shapely, we could throw an error instead of going in to swap, see https://github.com/shapely/shapely/issues/1349) 

Ref. https://trac.osgeo.org/geos/ticket/540

One way to set up such an interface is well described in the Numpy docs (NEP49): https://numpy.org/doc/stable/reference/c-api/data_memory.html

@jorisvandenbossche @hobu

caspervdw:
I just realised that the Python allocator will require the GIL (global interpreter lock) to be held. I missed that detail when scanning the docs earlier.

So for the shapely usecase this will probably be a reason to keep using the system allocator. Or am I missing something @jorisvandenbossche

jorisvandenbossche:
Yes, that indeed seems a reason we wouldn't want to set up GEOS with Python's memory allocator in Shapely. There is also a [Raw Memory Interface](https://docs.python.org/3/c-api/memory.html#raw-memory-interface) that does not require the GIL, but it seems this is just a plain wrapper around `malloc` et al, so using this wouldn't give any advantage (it's rather meant for overriding this in Python with your custom allocator).

That said, I think this would still be an interesting enhancement to explore. Whether it is for external users of GEOS to plug in their own allocator (like PostGIS might do), or whether it is to experiment with different memory allocators in GEOS itself. 
(I would assume that the leg work for both use cases would largely overlap?)

For example, in the Arrow C++ project, we have a configurable [MemoryPool](https://arrow.apache.org/docs/cpp/memory.html#memory-pools) ([header](https://github.com/apache/arrow/blob/33f2c0ec8e281fc4fe8c03b07ed2d32e343d9b0e/cpp/src/arrow/memory_pool.h#L62-L114)), and have implementations based on the system allocator, jemalloc and mimalloc. But we default to jemalloc / mimalloc (depending on the OS) because those give better performance (if enabled at build time).

rouault:
[...]

PR #804, opened 2023-01-21
Replace all geos::detail::make_unique<> occurences with std::make_unique<>

rouault:
and remove geos::detail::make_unique


dbaston:
I was thinking about this. But if #704 is going to happen, then we'll need to keep our own `make_unique` anyway, so...

rouault:
> if #704 is going to happen, then we'll need to keep our own `make_unique` anyway, so...

good point. Feel free to close that one. Was just a sed exercice :-)

PR #806, opened 2023-01-21
Avoid FE_INVALID floating point errors in HausdorffDistance and VoronoiDiagram

jorisvandenbossche:
Similarly as https://github.com/libgeos/geos/pull/790 / https://github.com/libgeos/geos/pull/791 and fixes the remaining cases originally reported in https://github.com/libgeos/geos/issues/515

### DIFF
--- a/src/algorithm/CGAlgorithmsDD.cpp
+++ b/src/algorithm/CGAlgorithmsDD.cpp
@@ -158,6 +158,9 @@ CGAlgorithmsDD::intersection(const CoordinateXY& p1, const CoordinateXY& p2,
 CoordinateXY
 CGAlgorithmsDD::circumcentreDD(const CoordinateXY& a, const CoordinateXY& b, const CoordinateXY& c)
 {
+    if((a == b) && (a == c)){
+        return a;
+    }
     DD ax = DD(a.x) - DD(c.x);
     DD ay = DD(a.y) - DD(c.y);
     DD bx = DD(b.x) - DD(c.x);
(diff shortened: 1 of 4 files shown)


src/geom/LineSegment.cpp
```diff
@@ -55,6 +55,9 @@ LineSegment::projectionFactor(const CoordinateXY& p) const
     if(p == p1) {
         return 1.0;
     }
+    if(p0 == p1) {
+        return 0.0;
+    }
```
jorisvandenbossche, line 60: This is the case of projecting on a LineSegment where the start and end point is the same (which otherwise ends up doing a 0/0 in the code below)

src/algorithm/CGAlgorithmsDD.cpp
```diff
@@ -158,6 +158,9 @@ CGAlgorithmsDD::intersection(const CoordinateXY& p1, const CoordinateXY& p2,
 CoordinateXY
 CGAlgorithmsDD::circumcentreDD(const CoordinateXY& a, const CoordinateXY& b, const CoordinateXY& c)
 {
+    if((a == b) && (a == c)){
+        return a;
+    }
```
jorisvandenbossche, line 163: For this case, the actual floating point error happened in `DD::selfDivide`, in case of dividing DD(0) with DD(0) (which is triggered by `numx / denom` below when a, b, and c are identical).  
(I was first thinking to fix it in `DD::selfDivide`, but not directly clear what it then should return, and I assume in general to avoid the FE you need to avoid doing the 0/0 division).

(sidenote: I am not sure what the code below would actually do if only one pair of a, b, and c would be identical)

dbaston:
Thanks!

PR #810, opened 2023-01-23
Add GEOSEqualsIdentical C API function for checking binary equivalence

dbaston:
This PR adds a C API function `GEOSIdentical` that determines if two geometries are ~binary-equivalent~ pointwise-equivalent. It is distinct from other predicates such as `GEOSEquals` (which uses topological equality, returning `true` for different representations of the same structure) and `GEOSEqualsExact` (which uses pointwise equality, but allows for a distance tolerance and ignores Z/M values). This also differs from the above functions in applying a test of binary equivalence (NaN = NaN) rather than numeric equality (NaN != NaN).

I think there may _also_ be a place for some kind of `GEOSEqualsExactDims(int dims, double tol)` function that allows you to specify which dimensions you're interested in, but for the common case of "tol = 0, dims = ALL" it is worthwhile to have a specific function that can use a more efficient implementation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@ xxxx-xx-xx
   - Geometry clustering: DBSCAN, geometry intersection/distance, envelope
     intersection/distance (GH-688, Dan Baston)
   - CAPI: GEOSLineSubstring (GH-706, Dan Baston)
+  - CAPI: GEOSEqualsIdentical (GH-810, Dan Baston)
   - Voronoi: Add option to create diagram in order consistent with inputs (GH-781, Dan Baston)
 
 - Fixes/Improvements:
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -267,6 +267,12 @@ extern "C" {
         return GEOSEqualsExact_r(handle, g1, g2, tolerance);
     }
 
+    char
+    GEOSIdentical(const Geometry* g1, const Geometry* g2)
+    {
+        return GEOSIdentical_r(handle, g1, g2);
+    }
+
     int
     GEOSDistance(const Geometry* g1, const Geometry* g2, double* dist)
     {
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -268,9 +268,9 @@ extern "C" {
     }
 
     char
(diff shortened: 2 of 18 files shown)


dbaston:
Name suggestion from Slack: `GEOSEqualsIdentical`

dbaston:
@jorisvandenbossche , would this provide a solution for https://github.com/shapely/shapely/issues/1732 ?

jorisvandenbossche:
I think so, yes, at least in idea (on the short term we will need to have a fallback anyway for released GEOS vesions). 

For the exact behaviour details: naively (and based on the current docs), I would expect that `GEOSEqualsExact(geom, tolerance=0)` and `GEOSEqualsIdentical(geom)` to be equivalent (minus the dimension issue). 

But because of the implementation details of using `memcmp` that's not exactly the case. As far as I understand (no floating point expert, just googled it ;)), the bit-by-bit comparison for floats has the following consequences:

- NaNs also compare equal if they are in the same location in both geometries. I think this is something we want for this function? (and contrasts it with GEOSEqualsExact, which will return False when comparing for example `"POINT (1 NaN)"`) 
- My [understanding](https://stackoverflow.com/questions/25808445/why-cant-the-floating-point-types-compare-by-using-memcmp-function#comment40373249_25808698) is that, generally, bit-by-bit comparison will give the same as `==` for floats, with the exception of two cases: zeroes and NaNs. It's OK that NaNs compare equal (bullet point above), but you have different NaN representations that will then not compare equal (I don't know how easy it would be to end up with that through operations in GEOS). But -0.0 ad +0.0 also have different bit representation (and so would make GEOSEqualsIdentical return False), and I assume this is certainly possible to end up with in GEOS (eg reduce precision of two different geoms and then check if they are identical). 

Using the `memcmp` certainly gives an easier implementation (and faster), just wondering about the (potentially confusing) corner cases it might introduce.

dbaston:
Thanks for the feedback. Like you, I'm not worried about considering quiet and signaling NaNs to be distinct. And the only way to get a signaling NaN is by deliberately creating one. But I think the behavior with signed zeros would be a problem. I've removed the `memcmp` usage so that +/- 0.0 are equivalent. A side effect is that different NaNs are also now equivalent.

jorisvandenbossche (approved):

PR #815, opened 2023-01-31
DOC: Update docstring of GEOSSnap / GeometrySnapper to clarify behavior

### LINKED ISSUE #814: BUG: GEOSSnap creates invalid MultiPolygons
Detected using Shapely 2 / GEOS 3.12.0dev.

Snapping a very simple MultiPolygon of 2 boxes which are originally valid, to a line within tolerance between them, results in an invalid MultiPolygon because the polygons now share an edge:

```Python
>>> import shapely
>>> boxes = shapely.MultiPolygon([box(0, 0, 1, 1), box(3, 0, 4, 1)])
>>> boxes
<MULTIPOLYGON (((1 0, 1 1, 0 1, 0 0, 1 0)), ((4 0, 4 1, 3 1, 3 0, 4 0)))>
>>> shapely.is_valid(boxes)
True
>>> snapped = shapely.snap(boxes, LineString([(2, 0), (2, 1)]), tolerance=2)
>>> snapped
<MULTIPOLYGON (((2 1, 1 1, 0 1, 0 0, 2 0, 2 1)), ((4 0, 4 1, 2 1, 2 0, 4 0)))>
>>> shapely.is_valid(snapped)
False
```

Based on the description of GEOSSnap, I couldn't tell if this was expected behavior.  It doesn't look like MultiPolygons are covered in the existing test cases.

I'm also not sure what is correct behavior here: should it snap only the first and not the second?  Should it union the touching MultiPolygons into a single Polygon?  Should it do what it is doing now, in which case we should note that it may produce invalid geometries in some cases?

brendan-ward:
resolves #814 

This updates the docstrings of `GEOSSnap` and `GeometrySnapper` to better clarify behavior, especially regarding creation of invalid geometries.

### DIFF
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -4079,7 +4079,20 @@ extern GEOSGeometry GEOS_DLL *GEOSGeom_transformXY(
     void* userdata);
 
 /**
-* Snap first geometry onto second within the given tolerance.
+* Snaps the vertices and segments of the first geometry to vertices of the
+* second geometry within the given tolerance.
+*
+* Snapping one geometry to another can improve robustness for overlay operations
+* by eliminating nearly-coincident edges (which cause problems during noding and
+* intersection calculation). Because too much snapping can result in invalid
+* topology being created, heuristics are used to determine the number and
+* location of snapped vertices that are likely safe to snap. These heuristics
+* may omit some potential snaps that are otherwise within the tolerance.
+*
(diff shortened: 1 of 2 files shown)


pramsey:
Hm, the language about "improving robustness" I think dates to pre-overlayNG, we should happily chomp through nearly coincident lines now. I'm going to merge and then edit a littl emore.

Issue #808, opened 2023-01-23
Geometry created as 3D with NaN as z coordinate is 3D but WKT shows it as 2D


jorisvandenbossche:
With the latest refactors of CoordinateSequence to better support different dimensionalities (https://github.com/libgeos/geos/pull/721 et al), the CoordinateSequence now better keeps track of its dimension (how it was created), and so is this `m_hasdim`/`m_hasz` bits are set, it no longer relies on checking if the z coordinate is NaN to determine if it is a 2D or 3D geometry (since 2D coords are still stored as XYZ physically). 

But that seems to create some corner cases if you happen to explicitly create a 3D geometry but where the z coordinate is NaN .. Using shapely (built with GEOS main) here as I didn't have time to create a C reproducer, but mentioned the C APIs that it corresponds to:

```python
>>> import shapely
>>> from numpy import NaN
# creating a Point with `GEOSCoordSeq_create_r` passing dims=3 -> `CoordinateSequence(1, 3)`
>>> p = shapely.Point(1, 2, NaN)
# this now knows it is 3D using `GEOSHasZ_r` 
>>> p.has_z 
True
# or using `GEOSGeom_getCoordinateDimension_r`
>>> shapely.get_coordinate_dimension(p)
3
# but printing it with WKT shows it as 2D (with `GEOSWKTWriter_setOutputDimension_r` set to 3)
>>> p
<POINT (1 2)>
```

So the geometry now keeps track of that it is 3D, but the WKT shows it as 2D (probably still checking if the z-coordinate is Nan?).

While WKB seems to follow the dimension and return 3D:

```python
>>> shapely.Point(1, 2).wkb.hex()
'0101000000000000000000f03f0000000000000040'
>>> shapely.Point(1, 2, NaN).wkb.hex()
'0101000080000000000000f03f0000000000000040000000000000f87f'
```

cc @dbaston

dbaston:
This is intended behavior for the `WKTWriter`, though I'm not sure what the motivation for that behavior is. It omits Z and M if they do not have at least one non-NaN value.

We could remove this behavior (I'm not opposed, but others might be) or add a flag to disable it.

dbaston:
> But that might also be from the time that checking for NaN was the only way to check if a geometry was 2D or 3D? (so not necessarily "intended" for removing that for "actual" 3D geometries).

[...]

PR #817, opened 2023-02-02
WKTWriter: Disable removal of empty dimensions

dbaston:
The removal of empty dimensions (JTS behavior) is still available by a flag that defaults to off.

Also fix a case in `BufferBuilder` where a geometry declared to be 2D was storing 3D coordinates. Correct output of this case depended on `WKTWriter` ignoring the declared dimensionality.

Resolves https://github.com/libgeos/geos/issues/808


jorisvandenbossche:
Thanks for this fix, looks good!

PR #824, opened 2023-02-14
Remove lazy calculation of Geometry envelope

dbaston:
This PR updates `Geometry` so that envelopes are calculated eagerly rather than on their first use. This avoids the need to "prime" geometries before they can safely be used by the predicate and overlay operators in multiple threads. (See https://trac.osgeo.org/geos/ticket/855)

Also, `Polygon` is updated to avoid storing an envelope, because we can reference the shell's envelope directly.

Performance-wise, it seems to be a wash. Clients like PostGIS that store the Envelope may now want to request it from geometries constructed in GEOS to avoid re-calculating it.

image

### DIFF
--- a/include/geos/geom/GeometryCollection.h
+++ b/include/geos/geom/GeometryCollection.h
@@ -199,6 +199,7 @@ public:
 protected:
 
     GeometryCollection(const GeometryCollection& gc);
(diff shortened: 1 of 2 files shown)


dr-jts:
Seems like a good idea.  It's simpler, and makes the library closer to being thread-safe.

Issue #830, opened 2023-02-23
BUG: segfault on unary union of empty point and line


brendan-ward:
First reported in [Shapely #1769](https://github.com/shapely/shapely/issues/1769).

The following segfaults for me on macOS 12.5 for GEOS 3.9 - 3.12.0dev:

```bash
geosop unaryUnion -a "GEOMETRYCOLLECTION (POINT EMPTY, LINESTRING (0 0, 1 1))"                                                                                                                                                                                   
```

Note that the order of items within the collection does not matter.  This apparently only affects a combination of an empty point and a linestring (empty or non-empty); it does not happen for an empty point an other geometry types that I've tested so far.

brendan-ward:
Specifically, this is appears to only happen when the only point is an empty point:

```bash
> geosop unaryUnion -a "GEOMETRYCOLLECTION (LINESTRING EMPTY, POINT EMPTY, POINT (0 0))"
GEOMETRYCOLLECTION (POINT (0 0), LINESTRING EMPTY)
```

PR #831, opened 2023-02-23
Fix crash on GC union with empty points

dbaston:
Fixes https://github.com/libgeos/geos/issues/830


brendan-ward:
Thanks for the quick fix @dbaston !  Confirmed that this resolves the error in shapely.

PR #834, opened 2023-02-27
GEOSSTRtree_iterate: Do not return removed items

### LINKED ISSUE #833: 3.11.0: STRtree: removed items are still returned by `GEOSSTRtree_iterate_r`
With GEOS version 3.11.0, consider the following code which creates an `GEOSSTRtree`, adds a single item to it, removes that item, and calls `GEOSSTRtree_iterate_r`:

```c
#include <geos_c.h>
#include <stdio.h>
#include <stdlib.h>

void callback(void *item, void *userdata) {
  printf("found %s\n", (char *)item);
}

int main(int argc, char *argv[]) {
  GEOSContextHandle_t handle = GEOS_init_r();
  GEOSWKTReader *wktReader = GEOSWKTReader_create_r(handle);
  GEOSSTRtree *tree = GEOSSTRtree_create_r(handle, 8);

  GEOSGeometry *g = GEOSWKTReader_read_r(handle, wktReader, "POINT (0 0)");
  char *item = "item 1";

  printf("inserting %s:\n", item);
  GEOSSTRtree_insert_r(handle, tree, g, item);

  printf("removing %s:\n", item);
  switch (GEOSSTRtree_remove_r(handle, tree, g, item)) {
  case 0:
    printf("%s was not removed\n", item);
    break;
  case 1:
    printf("%s was removed\n", item);
    break;
  default:
    printf("exception\n");
    break;
  }

  printf("iterating:\n");
  GEOSSTRtree_iterate_r(handle, tree, callback, NULL);

  return 0;
}
```

The iteration should find no items, but instead `"item 1"` is returned:

```console
$ gcc main.c $(pkg-config --cflags --libs geos) && ./a.out
inserting item 1:
removing item 1:
item 1 was removed
iterating:
found item 1
```

Surely `"item 1"` should not be returned as it was removed from the tree?

dbaston:
Fixes #833

### DIFF
--- a/include/geos/index/strtree/TemplateSTRtree.h
+++ b/include/geos/index/strtree/TemplateSTRtree.h
@@ -330,7 +330,9 @@ public:
     void iterate(F&& func) {
         auto n = built() ? numItems : nodes.size();
         for (size_t i = 0; i < n; i++) {
-            func(nodes[i].getItem());
+            if (!nodes[i].isDeleted()) {
+                func(nodes[i].getItem());
+            }
         }
     }
 
--- a/tests/unit/capi/GEOSSTRtreeTest.cpp
+++ b/tests/unit/capi/GEOSSTRtreeTest.cpp
@@ -468,6 +468,48 @@ void object::test<13>()
     GEOSSTRtree_destroy(tree);
 }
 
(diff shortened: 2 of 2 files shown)


twpayne:
Thank you for the speedy fix!

Issue #821, opened 2023-02-10
MaximumInscribedCircle crashes on polygon with infinity coordinates


pramsey:
From https://trac.osgeo.org/postgis/ticket/5320
```
SELECT ST_MaximumInscribedCircle(foo1.the_geom)  As result
FROM ((SELECT '0106000020E61000000100000001030000000100000005000000000000000000F07F000000000000F07F000000000000F07F000000000000F07F000000000000F07F000000000000F07F000000000000F07F000000000000F07F000000000000F07F000000000000F07F'::geometry AS the_geom ) ) As foo1 ;
```

PR #837, opened 2023-02-27
IndexedFacetDistance: Avoid crash with Inf coordinates

dbaston:
Fixes https://github.com/libgeos/geos/issues/821


dbaston:
Committed as 115404455e879bf01989864f80342497352368d4 and 29baedbf8b3ad3d392ea7448cefb61e448167088

PR #845, opened 2023-03-09
CI: Use ccache for Linux builds

dbaston:


mwtoews:
#573 added a CMake `-D USE_CCACHE=ON` convenience, which could be used here.

mwtoews:
MSVC would work. Folks have attempted this with https://github.com/frerich/clcache but this seems to be an abandoned effort.

For mingw64, perhaps try [cygpath](https://cygwin.com/cygwin-ug-net/cygpath.html):
```
$ cygpath -u "D:\\a\\geos\\geos"
/cygdrive/d/a/geos/geos
```

.github/workflows/ci.yml
```diff
@@ -238,102 +266,103 @@ jobs:
...
+
     steps:
     - name: 'Check Out'
       uses: actions/checkout@v3
 
     - name: 'Setup'
       uses: msys2/setup-msys2@v2
       with:
-        install: mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake make
+        install: mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake make ccache
```
mwtoews, line 285: This should fix the `windows-mingw` job, it doesn't seem to mind which direction the slashes go:
```suggestion
        install: mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake make mingw-w64-x86_64-ccache
```
see [mingw-w64-x86_64-ccache](https://packages.msys2.org/package/mingw-w64-x86_64-ccache) vs [ccache](https://packages.msys2.org/package/ccache).

Also L298 might not be needed.

.github/workflows/ci.yml
```diff
@@ -238,102 +266,103 @@ jobs:
...
       with:
-        install: mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake make
-        update: true
+        path: .ccache
+        key: windows-mingw-${{ matrix.build_type}}-${{ github.ref_name }}-${{ github.run_id }}
+        restore-keys: windows-mingw-${{ matrix.build_type}}
 
     - name: 'Build'
       run: |
+       export CCACHE_BASE_DIR=$(pwd)
```
mwtoews, line 298: probably unnecessary, but harmless to keep

mwtoews (approved):
Nicely done!

Issue #662, opened 2022-08-16
GEOSGeom_setPrecision flattens multi-geometries


pramsey:
So no-op precision reductions end up having a spurious type change as well:
```
geosop -a "MULTILINESTRING((0 0,10 10))" -f wkt reducePrecision 1

LINESTRING (0 0, 10 10)
```

pramsey:
@dr-jts thoughts?

pramsey:
Reported in https://trac.osgeo.org/postgis/ticket/5208

dr-jts:
It seems reasonable to preserve the collection type of the input.

PR #846, opened 2023-03-09
GeometryPrecisionReducer: preserve collection-level of the input geometry 

pramsey:
References GH-662.

### DIFF
--- a/src/precision/GeometryPrecisionReducer.cpp
+++ b/src/precision/GeometryPrecisionReducer.cpp
@@ -53,6 +53,32 @@ GeometryPrecisionReducer::reduce(const Geometry& geom)
         reduced = PrecisionReducerTransformer::reduce(geom, targetPM, removeCollapsed);
     }
 
+    // Match the collection level of the output to the input
+    // if necessary
+    if (geom.isCollection()
+        && ! reduced->isCollection()
+        && (geom.getCoordinateDimension() == reduced->getCoordinateDimension())
+        && reduced->getGeometryTypeId() != GEOS_GEOMETRYCOLLECTION)
+    {
+        GeometryTypeId typeId = reduced->getGeometryTypeId();
+        if (reduced->isEmpty()) return geom.getFactory()->createEmpty(geom.getGeometryTypeId());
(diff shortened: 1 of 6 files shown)


src/precision/GeometryPrecisionReducer.cpp
```diff
@@ -53,6 +53,32 @@ GeometryPrecisionReducer::reduce(const Geometry& geom)
         reduced = PrecisionReducerTransformer::reduce(geom, targetPM, removeCollapsed);
     }
 
+    // Match the collection level of the output to the input
+    // if necessary
+    if (geom.isCollection()
+        && ! reduced->isCollection()
+        && (geom.getCoordinateDimension() == reduced->getCoordinateDimension())
+        && reduced->getGeometryTypeId() != GEOS_GEOMETRYCOLLECTION)
```
dbaston, line 61: This is implied by line 59, no?

src/precision/GeometryPrecisionReducer.cpp
```diff
@@ -53,6 +53,32 @@ GeometryPrecisionReducer::reduce(const Geometry& geom)
...
+    {
+        GeometryTypeId typeId = reduced->getGeometryTypeId();
+        if (reduced->isEmpty()) return geom.getFactory()->createEmpty(geom.getGeometryTypeId());
+        std::vector<std::unique_ptr<Geometry>> subgeoms;
+        subgeoms.push_back(std::move(reduced));
```
dbaston, line 66: Thoughts on putting this logic into a method of `GeometryFactory` ?

src/precision/GeometryPrecisionReducer.cpp
pramsey: Yes, true.

src/precision/GeometryPrecisionReducer.cpp
pramsey: Makes sense... with a no-op pass through for already-collections?

pramsey:
Suggestions from @dbaston incorporated.

PR #851, opened 2023-03-18
CAPI fix : change int param to char for consistency

darkblue-b:
minor param change to new CAPI call

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -486,7 +486,7 @@ extern "C" {
     Geometry*
     GEOSConcaveHullByLength(const Geometry* g,
                     double length,
-                    unsigned int allowHoles)
+                    char allowHoles)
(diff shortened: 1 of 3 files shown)


dbaston:
While the overall level of inconsistency in the C API is regrettable, I think it makes sense for `GEOSConcaveHullByLength` to stick with the same types as `GEOSConcaveHull`.

dr-jts:
> While the overall level of inconsistency in the C API is regrettable, I think it makes sense for `GEOSConcaveHullByLength` to stick with the same types as `GEOSConcaveHull`.

Definitely.

But we need to try and improve consistency going forward.  Not sure if we have documented coding standards for GEOS?

pramsey:
No standards, just copying whatever happens to be closest to hand, generally speaking.

PR #858, opened 2023-03-30
Add linear referencing functions "interpolate" and "project" to geosop

mwtoews:
I noticed these functions were missing from geosop.

A careful review of my novice C++ would be appreciated.

Other name suggestions are welcome, e.g. `interpolate` vs `extractPoint`.

A `substring`/`extractLine` operator would be nice, but `geosop` is currently limited to one argument, so this presently isn't possible.


dr-jts:
Looks good to me.  Thanks for the contribution.

I will look into an enhancement to allow multiple operation arguments.

PR #861, opened 2023-04-03
Improve geosop performance and refactor

dr-jts:
Improves `geosop` performance by creating operations on an as-needed basis. 
Provide precise operation function signatures.
Rename `GeometryOp` class.


dr-jts:
@mwtoews This PR provides the basis of supporting a second parameter for operations.  Stay tuned - I will implement that and use your `extractLine` suggestion to test it.

PR #864, opened 2023-04-09
Add Point::getM() to C++ API and GEOSGeomGetM to C API

mwtoews:
- Add `GEOSGeomGetM` (and `GEOSGeomGetM_r`) to C API
- Add `Point::getM()` to C++ API
- Refactor `Point::getZ()` to use `CoordianteSequence::getOrdinate(0, CoordinateSequence::Z)`


src/geom/Point.cpp
```diff
@@ -180,6 +180,19 @@ Point::getZ() const
...
+double
+Point::getM() const
+{
+    if(isEmpty()) {
+        throw util::UnsupportedOperationException("getM called on empty Point\n");
+    }
+    if(getCoordinatesRO()->hasM()) {
+        return coordinates.getAt<CoordinateXYZM>(0).m;
+    } else {
+        return std::numeric_limits<double>::quiet_NaN();
```
mwtoews, line 192: I'm not sure if this is the correct approach, as I could not identify how `getZ()` returned nan for non-Z geometries. Suggestions welcome here.

tests/unit/capi/GEOSGeomGetXYZTest.cpp
mwtoews, line 1: Should this file be renamed to `GEOSGeomGetXYZMTest.cpp`?

src/geom/Point.cpp
dbaston: I'd be inclined to delegate this (and `getZ()`) to `CoordianteSequence::getOrdinate`.

That function looks like it may not handle XYM correctly -- it should dispatch on `CoordinateType`, not `hasZ` and `hasM`.

tests/unit/capi/GEOSGeomGetXYZTest.cpp
dbaston: Sure

src/geom/Point.cpp
mwtoews: Thanks, `coordinates.getOrdinate(0, CoordinateSequence::M)` is much simpler, but correct that XYM is not correctly handled. (XYZM is fine). I'll also refactor `GetZ` to use `getOrdinate`.

src/geom/CoordinateSequence.cpp
```diff
@@ -288,7 +288,7 @@ CoordinateSequence::getOrdinate(std::size_t index, std::size_t ordinateIndex) co
         case CoordinateSequence::Z:
             return hasZ() ? getAt<Coordinate>(index).z : DoubleNotANumber;
         case CoordinateSequence::M:
-            return hasM() ? (hasZ() ? getAt<CoordinateXYZM>(index).m : getAt<CoordinateXYM>(index).m) : DoubleNotANumber;
+            return hasM() ? getAt<CoordinateXYZM>(index).m : DoubleNotANumber;
```
mwtoews, line 291: This works, as there seems to be an issue with `CoordinateXYM`.

mwtoews: @dbaston if you want to investigate the issue, re-set this line to the previous edit to see the issue via `ctest --output-on-failure -R GEOSGeomGetXYZM`. If the issue is simple, feel free to push a commit here on in another PR/main.

dbaston: In current GEOS, `POINT M (1 2 3)` is backed by `CoordinateXYZM`. So to check if we can access the coordinate as an XYZM or an XYM we need to ask that question directly, not infer it from hasZ/hasM.

src/geom/CoordinateSequence.cpp
```diff
@@ -288,7 +288,9 @@ CoordinateSequence::getOrdinate(std::size_t index, std::size_t ordinateIndex) co
         case CoordinateSequence::Z:
             return hasZ() ? getAt<Coordinate>(index).z : DoubleNotANumber;
         case CoordinateSequence::M:
-            return hasM() ? (hasZ() ? getAt<CoordinateXYZM>(index).m : getAt<CoordinateXYM>(index).m) : DoubleNotANumber;
+            return getCoordinateType() == CoordinateType::XYZM ? getAt<CoordinateXYZM>(index).m :
+                                                                 getCoordinateType() == CoordinateType::XYM ? getAt<CoordinateXYM>(index).m :
```
mwtoews, line 292: Thanks for the input! I don't see any test coverage with `getCoordinateType() == CoordinateType::XYM`. The tests appear to only cover `CoordinateType::XYZM`.

dbaston: That's right. It's not possible to construct a `CoordinateSequence` that is backed by `XY` or `XYM` unless you change this `#define`:

https://github.com/libgeos/geos/blob/main/src/geom/CoordinateSequence.cpp#L40

The `main` version of line 291 is only correct if the define is _off_ (bad). The version you proposed is correct if the define is _on_, which it is (better). This version is correct either way.

dbaston (approved):

PR #871, opened 2023-04-14
Change WKTWriter Old3D option to only add "M" for XYM

mwtoews:
The WKTWriter Old3D option previously only modified the behaviour of XYZ geometries, rendering "POINT Z (1 2 3)" as "POINT (1 2 3)". This PR changes geometries with M coordinates to also remove dimensionality tag, as I suppose this was the original meaning of the option.

<s>For instance, "POINT M (1 2 4)" would be formatted as "POINT (1 2 4)". I've added a note that 3D XYZ and XYM geometries are indistinguishable.</s>

Furthermore, "POINT ZM (1 2 3 4)" is now formatted as "POINT (1 2 3 4)" with Old3D enabled. Is this the same output as older PostGIS versions?

### DIFF
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -5407,8 +5407,9 @@ extern int  GEOS_DLL GEOSWKTWriter_getOutputDimension(GEOSWKTWriter *writer);
 
 /**
 * Sets the format for 3D outputs. The "old 3D" format does not
-* include a dimensionality tag, eg. "POINT(1 2 3)" while the new (ISO)
-* format does includes a tag, eg "POINT Z (1 2 3)".
+* include a Z dimension tag, e.g. "POINT (1 2 3)", except for XYM,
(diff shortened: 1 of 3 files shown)


mwtoews:
Looking further at [PostGIS 1.3](https://postgis.net/docs/manual-1.3/ch04.html#id434922) (the oldest manual I can find) it seems that the "old 3D" for XYM was more like "POINTM(0 0 0)", and XYZM was "POINT(0 0 0 0)". Perhaps this option should actually reflect these styles? Opinions welcome!

mwtoews:
Updated old3D to add "M" without a space, only for XYM coordinate types. This should resemble old PostGIS docs.

mwtoews:
Last change renders a space before "M" (e.g.) "POINT M (1 2 4)", which is slightly different than old PostGIS docs. This is because "POINTM" is not understood by older GEOS versions. while "POINT M" is (even if it is interpreted as XYZ, due to older behaviour).

PR #878, opened 2023-04-26
Remove Orientation.isCCW exception

dr-jts:
Change behaviour of `Orientation.isCCW` and `GEOSCoordSeq_isCCW` to avoid throwing an exception when the input sequence is short (< 4 vertices).  This aligns with JTS semantics, and simplifies logic in internal routines.  In particular, `normalize` now works on invalid (short) rings without additional checks.


dbaston:
Should probably be noted in `NEWS` as a change in behavior.

dr-jts:
> Should probably be noted in `NEWS` as a change in behavior.

Yes, definitely.  Thanks for the reminder.

PR #883, opened 2023-04-28
Fix MaximumInscribedCircle and LargestEmptyCircle performance and memory issues

### LINKED ISSUE #875: MaximumInscribedCircle goes into long loop and fails
As noted in -  https://lists.osgeo.org/pipermail/postgis-devel/2023-April/029923.html

MaximumInscribedCircle cries in tears when faced with this geometry

```
SELECT ST_MaximumInscribedCircle( 'POLYGON ((59.3 100.00000000000001, 99.7 100.00000000000001, 99.7 100, 59.3 100, 59.3 100.00000000000001))') ;
```

This has been tested on GEOS 3.9, GEOS 3.10.2, GEOS 3.11.2, GEOS 3.12 (main branch).

FWIW my tests on 3.11.2 and GEOS 3.12, the query eventually completed in 3 minutes (this is testing under CentOS 8 and MingW64)
 but with an error:

```ERROR:  Error calculating GEOSMaximumInscribedCircle.```

Others with more powerful systems did not fair so well.  Not sure if it was postgis that realized something was wrong or GEOS.

### PR DESCRIPTION
**Pull request by dr-jts**
This PR adds heuristics to MaximumInscribedCircle and LargestEmptyCircle to prevent long-running (and memory-intensive) processing for nearly flat and very thin inputs.

There are two problems solved:

* if the input extent is very small in one dimension (i.e. almost flat), the cell grid was initialized with cells of very small size, causing a memory explosion.  This is avoided by using a single initial grid cell covering the extent.
* Very thin (low-area) geometries (or LEC boundaries) cause a long search to find a cell with a centre inside the geometry (since very few cells in the search space satisfy this condition). A heuristic is used to determine a maximum number of iterations permitted. This means that the computed result may lie outside the input geometry. This can be mitigated by using a smaller tolerance distance; but there is no way to guarantee the centre will lie inside.

Fixes #875.

Post of https://github.com/locationtech/jts/pull/978 and https://github.com/locationtech/jts/commit/a45d9888a28cdb069a9de7674012a92f204770da

### DIFF
--- a/src/algorithm/construct/LargestEmptyCircle.cpp
+++ b/src/algorithm/construct/LargestEmptyCircle.cpp
@@ -18,6 +18,7 @@
  **********************************************************************/
 
 #include <geos/algorithm/construct/LargestEmptyCircle.h>
+#include <geos/algorithm/construct/MaximumInscribedCircle.h>
 #include <geos/geom/Coordinate.h>
 #include <geos/geom/CoordinateSequence.h>
 #include <geos/geom/Envelope.h>
@@ -28,6 +29,7 @@
 #include <geos/geom/MultiPolygon.h>
 #include <geos/algorithm/locate/IndexedPointInAreaLocator.h>
 #include <geos/operation/distance/IndexedFacetDistance.h>
+#include <geos/util/Interrupt.h>
 
 #include <typeinfo> // for dynamic_cast
 #include <cassert>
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/algorithm/construct/LargestEmptyCircle.cpp [hunk 1]
```diff
@@ -231,12 +232,18 @@ LargestEmptyCircle::compute()
      * Carry out the branch-and-bound search
      * of the cell space
      */
-    while (!cellQueue.empty()) {
+    std::size_t maxIter = MaximumInscribedCircle::computeMaximumIterations(boundary.get(), tolerance);
+    std::size_t iterationCount = 0;
+    while (!cellQueue.empty() && iterationCount < maxIter) {
```
dbaston, line 237: Should any action be taken if `maxIter` is hit while the queue is non-empty?

dr-jts: For most practical cases the iteration converges very fast to a good approximation to the LEC centre.  The only other option I can see is to throw an exception warning that the result could be inaccurate.

dbaston: It seems like without an exception, you don't know if the result is correct or not?

dr-jts: The algorithm is a heuristic approximation, so it will almost never find a perfectly correct answer anyway.  If the max iterations are exceeded, the result should be "close".  

I think there's an improvement that could be made to ensure that the computed point always lies inside the boundary/obstacle convex hull.  That fulfills the most important part of the contract.

Issue #665, opened 2022-08-22
SingleSided buffer shows artifacts


hmundle:
When I want to create a single sided buffer, GEOS shows strange result even for easy line inputs.

This is my input (`geom`): `LINESTRING (50 50, 150 150, 150 100, 150 0)`
![image](https://user-images.githubusercontent.com/1384773/185927212-6d55d89b-0c46-43fa-ac47-f72f4bffc950.png)

This is my implementation:
``` c++
geos::operation::buffer::BufferParameters bp;
bp.setSingleSided(true);
geos::operation::buffer::BufferOp ob(geom.get(), bp);
std::unique_ptr<Geometry> g3 = ob.getResultGeometry(d);
```
with `d = -21.0`

The resulting output shows artifacts from the seconds line segment:
`POLYGON ((50 50, 129 129, 129 150, 129.25925925925927 150, 143.92517768386912 143.92517768386912, 150 150, 150 141.408903891521, 164.66591842460988 135.33408157539012, 164.8492424049175 135.1507575950825, 150 120.30151519016499, 150 100, 150 0, 129 0, 129 99.30151519016499, 64.8492424049175 35.1507575950825, 50 50))`
![image](https://user-images.githubusercontent.com/1384773/185927459-c6ec5f98-3c74-4f8c-a784-62afcd287961.png)

If I am removing the vertex in the vertical line (`geom` = `LINESTRING (50 50,150 150, 150 0)`) the output is valid:
`POLYGON ((50 50, 150 150, 150 0, 129 0, 129 99.30151519016499, 64.8492424049175 35.1507575950825, 50 50))`
![image](https://user-images.githubusercontent.com/1384773/185927687-3ce481d7-4d6b-44f7-b1f0-49d7d98193c9.png)

If I am changing to `d = -20.0`, the output is valid as well:
POLYGON ((50 50, 150 150, 150 100, 150 0, 130 0, 130 101.7157287525381, 64.14213562373095 35.85786437626905, 50 50))
![image](https://user-images.githubusercontent.com/1384773/185928236-4ee3a93e-49ea-4557-bd02-5fe34692b285.png)

I am using GEOS 3.11.0.

Why do tiny changes result in a big change in the output?
I can imagine, that it is difficult to implement a single sided buffer algorithm for any strange input line. But this one is not strange, is it?

Some reference I found on the Internet:
https://trac.osgeo.org/geos/ticket/712 (closed with wontfix)
[...]

PR #891, opened 2023-05-04
Reduce artifacts in single-sided buffer output

strk:
Ports heuristic from deprecated BufferBuilder::bufferLineSingleSided code into 

See GH-665


strk:
continues in #895

PR #895, opened 2023-05-04
Reduce artifacts in single-sided buffer output

strk:
References GH-665

Continues from #891 which was made unusable by a force-push to the associated branch ( maybe I should stop using github for pull requests .. )


strk:
The approach in this PR is too destructive as removing edges can result in ending up with a disconnected graph which in turn throws TopologyException.

A different approach is needed.

strk:
Closing to reopen with a new approach

strk:
New approach is in  #896

PR #896, opened 2023-05-08
Single sided buffer artifacts removal

strk:
Only keep larger polygon from polygonization of single-sided buffer and input line.
References #665  implementing @hmundle approach


strk:
This is how a left-buffer with distance 50000 looks like with https://trac.osgeo.org/geos/ticket/810 input:

image

strk:
And this is the result of the buffer from https://trac.osgeo.org/geos/ticket/712 

image

strk:
Pushed to main branch as 091f6d99fd2d6485c8f1be1106f914c4bb4c2b84

PR #901, opened 2023-05-09
Fix reading MultiPoint Z/M WKT

### LINKED ISSUE #886: MultiPoint ZM WKT does not round-trip
This WKT parse and output is unexpected:
```bash
$ ./bin/geosop -a "MULTIPOINT ZM (0 0 4 3, 1 2 4 5)" -f wkt
MULTIPOINT Z (0 0 4, 1 2 4)
```
The expected result is created with:
```bash
$ ./bin/geosop -a "MULTIPOINT ZM ((0 0 4 3), (1 2 4 5))" -f wkt
MULTIPOINT ZM (0 0 4 3, 1 2 4 5)
```
This means than WKT does not round-trip:
```bash
$ ./bin/geosop -a "MULTIPOINT ZM ((0 0 4 3), (1 2 4 5))" -f wkt | ./bin/geosop -a stdin -f wkt
MULTIPOINT Z (0 0 4, 1 2 4)
```
WKB seems to round-trip OK:
```bash
$ ./bin/geosop -a "MULTIPOINT ZM ((0 0 4 3), (1 2 4 5))" -f wkb | ./bin/geosop -a stdin.wkb -f wkt
MULTIPOINT ZM (0 0 4 3, 1 2 4 5)
```

mwtoews:
Closes #886

Also, re-word some comments around "deprecated" or "correct" forms for multipoint. I can't see any official source for the correct form. Looking at OGC 06-103r4 for [SFA](https://www.ogc.org/standard/sfa) 1.2.1 it has both:

- page 18: `multipoint m(1 0 4, 3 1 4, 5 3 4)`
- page LXI (aka 61): `MultiPoint ((10 10), (20 20))`

The BNF description implies the second form is more "correct". Both GEOS and PostGIS WKTWriters only write the first form, so there is not clear "correct" form.

Also note that this PR does not attempt to fix an outstanding issue with EMPTY described in #902.

### DIFF
--- a/src/io/WKTReader.cpp
+++ b/src/io/WKTReader.cpp
@@ -338,9 +338,8 @@ WKTReader::readMultiPointText(StringTokenizer* tokenizer, OrdinateSet& ordinateF
     int tok = tokenizer->peekNextToken();
 
     if(tok == StringTokenizer::TT_NUMBER) {
-
-        // Try to parse deprecated form "MULTIPOINT(0 0, 1 1)"
-        auto coords = detail::make_unique<CoordinateSequence>();
+        // Try to parse "MULTIPOINT (0 0, 1 1)"
+        auto coords = detail::make_unique<CoordinateSequence>(0u, ordinateFlags.hasZ(), ordinateFlags.hasM());
 
         CoordinateXYZM coord(0, 0, DoubleNotANumber, DoubleNotANumber);
         do {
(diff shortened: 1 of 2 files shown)


dr-jts:
The general consensus is to accept the SF WKT BNF grammar as normative for the MultiPoint syntax.  So, each point has parens around it.  PostGIS and JTS have been emitting this form for years now.  I'm surprised that GEOS is using the old form. That should be fixed!

mwtoews:
Are you sure about PostGIS? Both `ST_AsText` and `ST_AsEWKT` seem to emit without parens.

I don't mind switching GEOS to emit with parens, as JTS does.

Issue #902, opened 2023-05-09
MultiPoint WKT with EMPTY parts does not round-trip


mwtoews:
MultiPoint geometries may contain EMPTY parts, however it is only preserved in the bracketed multipoint form. In the coordinate-sequence form it raises a ParseException:
```bash
$ ./bin/geosop -a "MULTIPOINT (EMPTY, (1 2))" -f wkt
MULTIPOINT (EMPTY, 1 2)
$ ./bin/geosop -a "MULTIPOINT (EMPTY, 1 2)" -f wkt
ParseException: Expected word but encountered number: '1'
$ ./bin/geosop -a "MULTIPOINT (EMPTY, (1 2))" -f wkt | ./bin/geosop -a stdin -f wkt
ParseException: Expected word but encountered number: '1'
```

There is no issue with WKB:
```bash
$ ./bin/geosop -a "MULTIPOINT (EMPTY, (1 2))" -f wkb | ./bin/geosop -a stdin.wkb -f wkt
MULTIPOINT (EMPTY, 1 2)
```
PostGIS is able to round-trip these WKT experiments.

dr-jts:
It might be nice to fix this, but since this bug hasn't been reported before it's probably not essential.  #903 will mean that standard GEOS output does round-trip.

mwtoews:
Closing as not planned, since `MULTIPOINT (EMPTY, 1 2)` is using a deprecated form for MultiPoint -- see #903

PR #903, opened 2023-05-09
Change MultiPoint WKT to use parentheses in sub-members

mwtoews:
This is a potentially breaking change that switches the WKT output of MultiPoint from:

    MULTIPOINT (2 8, 3 9)

to

    MULTIPOINT ((2 8), (3 9))

This aligns with JTS' outputs of MultiPoint and with ISO/OGC specs.

See also https://github.com/postgis/postgis/pull/682 which was implemented for PostGIS 3.3


pramsey:
I wonder how many downstream projects have tests against the WKT output of MULTIPOINT.

mwtoews:
Possibly a few, for instance I'll try shapely later. Another strategy is to modify WKTWriter to enable an option, keeping the default to use old behaviour.

dr-jts:
> I wonder how many downstream projects have tests against the WKT output of MULTIPOINT.

I think we have to note this as a breaking change and move on.  It's more important to be standards-compliant.

dr-jts:
`WKTReader` can still handle both forms, right?

It's good to update the unit tests that aren't specifically checking this behaviour.  But are there still tests for `WKTReader` that check both formats?

mwtoews:
> `WKTReader` can still handle both forms, right?

Yes, and I intentionally left a few test cases with the previous form. The only exception that I'm aware of #902 which will be closed as "not planned".

PR #908, opened 2023-05-14
Change WKBWriter default output dimension to 4

mwtoews:
This potentially breaking change is to use output dimension 4 with the WKBWriter.

For example, `POINT Z (1.1 2.2 3.3)` via writeHEX using defaults:

- Current: `01010000009A9999999999F13F9A99999999990140` or `POINT (1.1 2.2)`
- Proposed: `01010000809A9999999999F13F9A999999999901406666666666660A40` same as input WKT

Also add a few additional tests related to `setOutputDimension` that didn't seem to be covered.


pramsey:
Seems reasonable to me, although maybe a little less user facing than the WKT writer change. As long as people are only ever getting the dimensionality they put in, seems like a net reduction in user surprise, which is +1 :)

PR #915, opened 2023-05-24
Change WKTWriter default output dimension to 4 and trim to 'on'

mwtoews:
This potentially breaking change with the WKTWriter is to use output dimension 4 and trim 'on'.

For example, `POINT Z (1.1 2.2 3.3)` output using defaults are:

- Current: `POINT (1.1000000000000001 2.2000000000000002)`
- Proposed: `POINT Z (1.1 2.2 3.3)`

Other details:

- From the C++ API, both `toText()` and `toString()` will also use the new defaults; add `tests/unit/geom/Geometry/toTextTest.cpp` for this coverage
- Deprecated `GEOSGeomToWKT` (and `GEOSGeomToWKT_r`) should keep the previous defaults showing untrimmed 2D outputs, which is checked with `tests/unit/capi/GEOSGeomToWKTTest.cpp`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -74,6 +74,7 @@ xxxx-xx-xx
   - Remove Orientation.isCCW exception to simplify logic and align with JTS (GH-878, Martin Davis)
   - Change MultiPoint WKT to use parentheses in sub-members (GH-903, Mike Taves)
   - Change WKBWriter default output dimension to 4 (GH-908, Mike Taves)
+  - Change WKTWriter defaults output dimension to 4 and trim to 'on' (GH-915, Mike Taves)
 
 ## Changes in 3.11.0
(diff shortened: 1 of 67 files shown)


pramsey:
I would tend to think not? We have behaviour changes every release, but what I've internalize as "breaking" change has been C API change. Once you cannot plunk a GEOS binary on top of another one. Things like "what outputs are produced" are in flux (behaviour of concave hull, etc) and there's really no way to work on the library without that happening. I sympathize with your point of view, but I don't know if I quite agree.

BuonOmo:
Hm okay this make sense for sure. I'll subscribe to the gh releases then. In case there is some kind of information given before releasing, my email is _buonomo dot ulysse at gmail dot com_!

Issue #897, opened 2023-05-08
Offset curve has Weird Geometry


MrShutCo:
Output of offsetting this curve by 15 with CAP_ROUND and JOIN_ROUND is unexpected 
```
LINESTRING(292671.05 7336734.42,292624.85 7336739.72,292623.53 7336739.92,292622.36 7336740.2,292621.35 7336740.57,292620.94 7336740.78,292620.61 7336740.97,292620.33 7336741.14,292620.13 7336741.29,292619.79 7336741.57,292619.48 7336741.86,292619.34 7336742.01,292619.21 7336742.17,292619.09 7336742.34,292618.98 7336742.51,292618.8 7336742.88,292618.63 7336743.25,292618.57 7336743.43,292618.52 7336743.58,292618.48 7336743.71,292618.46 7336743.8,292618.44 7336743.98,292618.42 7336744.15,292618.41 7336744.33,292618.4 7336744.51,292618.39 7336744.69,292618.38 7336744.88,292618.38 7336744.91,292618.38 7336744.95,292618.38 7336744.99,292618.38 7336745.02,292618.38 7336745.06,292618.38 7336745.1,292618.38 7336745.13,292618.38 7336745.17,292618.38 7336745.21,292618.38 7336745.25,292618.38 7336745.43,292618.41 7336746.92)
```

Which gives 

```
LINESTRING(292672.7595667329 7336749.322260955,292626.8289702739 7336754.5913553545,292626.5791762319 7336754.629202936,292626.5097220095 7336754.654646562,292626.50454360223 7336754.64051091,292626.4042523117 7336754.65570656,292626.5015777726 7336754.632414997,292626.44602173776 7336754.480762037,292628.06232136535 7336753.77393037,292629.15317889326 7336753.036285378,292629.3464127592 7336752.884382651,292629.7456107097 7336752.533905098,292630.23746757495 7336752.041483207,292630.77071670454 7336751.429334497,292631.26624786 7336750.77544772,292631.44290077855 7336750.514153615,292632.18637824676 7336749.195972605,292632.6413907597 7336748.0441531595,292632.7640483846 7336747.639042543,292632.95704258484 7336746.627106525)
```

And looks like
<img width="809" alt="Screenshot 2023-05-08 at 10 04 16 AM" src="https://user-images.githubusercontent.com/12452116/236845044-3a720420-8ce6-4b2c-83a5-58a88cb7f3d5.png">

Is this expected behaviour?

[...]

PR #918, opened 2023-05-26
Fix OffsetCurve to use a minimum QuadrantSegs value

pramsey:
* Port https://github.com/locationtech/jts/pull/981
* References https://github.com/qgis/QGIS/issues/53165 
* References https://github.com/libgeos/geos/issues/897


tests/unit/capi/GEOSOffsetCurveTest.cpp
```diff
@@ -104,7 +104,7 @@ void object::test<3>()
 {
      checkOffset(
         "LINESTRING(0 0, 10 0, 10 10)",
-        "LINESTRING (12 10, 12 0, 10 -2, 0 -2)",
+        "LINESTRING (0 -2, 10 -2, 10.390180644 -1.9615705608, 10.7653668647 -1.847759065, 11.111140466 -1.6629392246, 11.4142135624 -1.4142135624, 11.6629392246 -1.111140466, 11.847759065 -0.7653668647, 11.9615705608 -0.390180644, 12 0, 12 10)",
         -2, 1, GEOSBUF_JOIN_ROUND, 2,
         0.000001
```
strk, line 109: Note that the last argument is a tolerance, so there's no need for more than 6 decimal digits with the current tolerance.

strk (approved):
Beside nitpicking, is always good to port. Maybe add the port info in the source files though, so we know where we left ? See BufferBuilder.cpp for an example (includes the commit hash of the last port)

PR #919, opened 2023-06-06
Expose geos ryu fork

Maxxen:
This PR exposes the modified GEOS ryu-based double formatting functionality in the c-api, as well as as a static function on the WKTWriter. The change itself is pretty small, but I also ran the bundled `astyle` on the modified files which seems to have formatted lots of other lines. Let me know if you want me to undo it.

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -25,6 +25,8 @@
 #include <geos/io/GeoJSONWriter.h>
 #include <geos/util/Interrupt.h>
 
+#include <ryu/ryu.h>
+
 #include <stdexcept>
 #include <new>
 
@@ -1441,6 +1443,11 @@ extern "C" {
     GEOSWKBWriter_setIncludeSRID(GEOSWKBWriter* writer, const char newIncludeSRID)
     {
         GEOSWKBWriter_setIncludeSRID_r(handle, writer, newIncludeSRID);
+    }
+
+    int
+    GEOS_d2sfixed_buffered_n(double d, unsigned int precision, char *result) {
+        return geos_d2sfixed_buffered_n(d, precision, result);
     }
 
     /* GeoJSON Reader */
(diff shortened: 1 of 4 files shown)


pramsey:
I'd rather not have the spurious astyle in the PR, if you can revert that (good reminder that a new astyle pass would probably be good at release time).
You have win32 build issues in CI to fix.
Why `GEOS_d2sfixed_buffered_n`? `GEOS_printDouble()` maybe? 
Also why expose our double printing at all? I get that maybe some people want ryu without themselves including ryu, but ... anyways, it's a small thing so I can see no real issue with ading it.
Release time is coming so if you can address sooner than later if you want this in next release.

Maxxen:
> I'd rather not have the spurious astyle in the PR, if you can revert that (good reminder that a new astyle pass would probably be good at release time).

Will do!
> You have win32 build issues in CI to fix.

Yeah I guess this is due to ryu not being linked into the clib, I should probably just call the WKTWriter static method instead.

> Release time is coming so if you can address sooner than later if you want this in next release.

Thanks for the heads up! I'll try to have it fixed by the end of the week.

> Also why expose our double printing at all? 

I think the changes made to the ryu code makes a lot of sense and I haven't been able to find something akin to it anywhere else or been able to replicate it using e.g fmtlib. My first thought was to install and link against the ryu code distributed in this repo separately but its not clear to me what license that part falls under, I assume the changes are LGPL like the rest of GEOS? Any thoughts about separating the GEOS changes into a proper fork with the original MIT license (or maybe make another attempt at upstreaming)? Given that GEOS, PostGIS and now DuckDB seem to prefer this version I'm sure there are others who would feel the same way. I could probably hack together my own patch with some work and include ryu separately, but since I'm already linking against GEOS that seems kind of wasteful.

pramsey:
I'm a little leery of this as being off the reservation, but it's small, so...

Issue #885, opened 2023-05-01
Handling of nonfinite ordinate values


caspervdw:
My PR in https://github.com/libgeos/geos/pull/881 sparked some discussion, so I think it is good to transfer it to an issue.

Has there been a discussion about blocking nonfinite values altogether?

@dr-jts wrote:

>Yes, several times (probably on some issue, but don't have a reference to hand).
> The conclusion is always to follow [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle): Be lenient in what you accept, and strict in what you emit.
> The rationale is that the library should be able to read everything, so that it can be validated and subsequently fixed as and how determined.

That's clear, so GEOS needs to deal with nonfinite values.

> As Paul, says, there is a combinatorial explosion of possibilities if the lib tries to do clever things with handling non-finite values. It would be best to take a rigorous and simple approach to determining how to decide if things are EMPTY internally. It might be simplest to specify that any non-finite X or Y ordinate in the first coordinate indicates an EMPTY CoordinateSequence. That way most geometry operations will Just Work, since they generally always check for EMPTY and take some default action.

I am certainly on board with a rigorous and simple approach. Some thoughts:

- My first thought is a `hasNonFinite` property. It's rigorous and simple. However it will impose an extra loop and it might cause performance degradation. And there is no chance of caching here as coordinate buffers may be user controlled.
- If we would go with @dr-jts proposal and process geometries as EMPTY geometries if the first coordinate contains a nonfinite, wouldn't this cause wrong assumptions? EMPTY means: no coordinates, while there are coordinates, they are just nonfinite.
- Secondly, it's not complete as a second coordinate might be nan and result in undefined behavior too.
- What to do with Z coordinates being nan? If a user provides an XYZ buffer with NaN Z coordinates, should GEOS act like it is empty? Or coerce it to 2D? What if only *some* Z coordinates are none?

pramsey:
In the limit, the existing rules of the road are fine: check for validity with isValid(); expect things to fail in exciting ways if your geometry is not valid; fix nonvalid things with makeValid().

The only downside I have see with this so far, is that non-finite values can actually crash some of the algorithms. So we are having to check the coordinates before starting, which is a full pass through. Even having a flag isn't free, since some of the best performing constructors we have now, the copying from buffers, are straight mem copies, so a "quick" finite check would involve, yep, a whole scan.

[...]

PR #927, opened 2023-06-23
Remove NaN point special case logic

caspervdw:
Addresses #885


pramsey (approved):

mwtoews:
The NEWS items will need to be moved from 3.12.0 to ... 3.13.0? Other?

PR #928, opened 2023-06-26
CPack: ignore other files/dirs starting with .; use verbatim option

mwtoews:
This PR uses [CPACK_VERBATIM_VARIABLES](https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_VERBATIM_VARIABLES) to simplify the escapes with the `CPACK_SOURCE_IGNORE_FILES` list for CPack. A new entry is added to ignore all files/directories that start with '`.`'. Only a few items that start with '`.`' are removed, otherwise the list is alphasorted.

Looking at results from `cmake --build . --target dist`, the only differences is that `.astylerc`, `.editorconfig` and `.mailmap` are now ignored.

Also fix HOWTO_RELEASE to the current main repo.


pramsey:
Feel free to merge.

Issue #740, opened 2022-11-23
BUG: inconsistency between prepared and unprepared polygon contains predicate


brendan-ward:
I'm seeing cases where the contains predicate is true when the polygon is not prepared, but false when the polygon is prepared.  The polygon is valid, so I don't think that is the issue.  In this case, the line is created by clipping a larger line to the polygon.  Example WKT files attached.

In contrast, the line is always within the polygon, regardless of whether or not the line is prepared.

I didn't see a way to use prepared predicates with `geosop`, but here is what I am doing with Shapely 2.0b2 / GEOS 3.11:

```Python
>>> import shapely
>>> poly = shapely.from_wkt(open("test_poly.wkt").read())
>>> line = shapely.from_wkt(open("test_line.wkt").read())
>>> shapely.contains(poly, line)  # calls GEOSContains_r
True
>>> shapely.prepare(poly)  # calls GEOSPrepare_r
>>> shapely.contains(poly, line)  # calls GEOSPreparedContains_r with prepared polygon
False
>>> shapely.within(line, poly)  # calls GEOSWithin_r
True
>>> shapely.prepare(line)
>>> shapely.within(line, poly)  # calls GEOSPreparedWithin_r
True
```

This might be a special case where the endpoint of the line is within the boundary of the polygon (fails ContainsProperly), which perhaps is causing one of the speedups in the prepared Contains predicate to fail when it shouldn't.

[test_poly.wkt.zip](https://github.com/libgeos/geos/files/10078170/test_poly.wkt.zip)
[test_line.wkt.zip](https://github.com/libgeos/geos/files/10078172/test_line.wkt.zip)

dr-jts:
In fact the test_line is NOT contained by the test_poly, since an endpoint of the line falls outside the polygon.
![image](https://user-images.githubusercontent.com/3529053/203664415-1ec50f13-c22c-4df0-ad5f-d2f3759c606d.png)
So the prepared contains result is correct, and the Relate result is wrong.

This is likely due to the issue described in https://github.com/locationtech/jts/issues/396.  Because the current Relate algorithm nodes the input before computing the DE-9IM topology, it can happen that due to robustness the noding causes the line endpoint and the computed node to be (incorrectly) identical.

Prepared Contains doesn't node the input, but uses (more) robust topological predicates.  So it computes the correct answer.

Fixing this is going to require a new Relate algorithm, unfortunately.

brendan-ward:
[...]

PR #937, opened 2023-07-10
Make line intersection computation more robust

### LINKED ISSUE #933: MultiLineString.contains produces incorrect result
I'm working on https://github.com/rapidsai/cuspatial/issues/1063, which are GPU-accelerated multi-geometry binary predicate operations. I discovered a test case that fails in comparison with GeoPandas/Shapely/GEOS results that I thought might interest you:

```
import shapely
from shapely.geometry import LineString, MultiLineString
b1 = LineString([(0, 0), (1, 1)])
b2 = LineString([(0.5, 0.5), (1, 0.1), (-1, 0.1)])
m = MultiLineString([b1, b2])
print(m.contains(b1))
print(m.contains(b2))
```
```
False
True
```

Expected result is that `m.contains(b1) == True`. I'm very interested to learn if I'm doing something wrong or if there's a bug here that has since gone unnoticed. I'll be keeping an eye over here! Thanks!

dr-jts:
Changes the computation of line intersections to use DoubleDouble arithmetic, for improved accuracy. This replaces the previous floating-point implementation.

The main reason for doing this is to fix some failures in `relate` computation (such as #933).  These are cases where both inputs contain the same line segment, but input A has a self-intersection in the line segment. The `relate` algorithm computes a new node point at the self-intersection.  The FP intersection algorithm computes an intersection point which falls slightly off the parent line segment.  The noded input A then no longer matches input B exactly, resulting in Exterior(A) / Interior(B) = 1 (rather then the expected result of F). The DD intersection computation is accurate enough to avoid this problem.

`CGAlgorithmsDD.intersection` is slower than the original FP implementation. But in typical use intersection calculation is a small fraction of overall processing, so there is minor performance impact in practice.

This changes the results of some overlay cases by a small amount. A few unit tests had to be changed to reflect this.

Fixes #933
See also https://github.com/locationtech/jts/issues/396. for another relate case which is fixed by this change.

> _Note: Even with this fix in place, having the relate algorithm relying on intersection computation accuracy is still inherently brittle.  A future redevelopment of the relate predicate algorithm may be able to handle this in a more robust way._


tests/xmltester/tests/issue/issue-geos-398.xml
```diff
@@ -32,9 +32,8 @@ POLYGON ((50.0044466166182886 0.0,
   </b>
 <test>
  <op name="union" arg1="A" arg2="B">
-MULTIPOLYGON (((0 25.950779066386122, 0 26.860827855779647, 0 29.83879230192533, 0 35.73921397193218, 60 13.183894681853793, 60 10.39852836763784, 60 7.802134559422377, 60 6.657099879646016, 60 6.510515132098641, 60 0, 55.31611974965083 0, 50.00444661661829 0, 43.31611974965083 0, 0 0, 0 14.034613342373582, 0 17.92266612923108, 0 21.587486526024364, 0 25.950779066386122),
-  (0 21.587486526024364, 34.025852439655786 6.898140262297274, 34.02585243965579 6.898140262297273, 0 21.587486526024364)),
-  ((13.44557253233479 36, 60 36, 60 16.794451829809802, 60 16.36440115550932, 60 14.043996030454757, 2.9187843276549756 36, 11.8945390820011 36, 13.44557253233479 36)))
+MULTIPOLYGON (((0 26.860827855779647, 0 29.83879230192533, 0 35.73921397193218, 60 13.183894681853793, 60 10.39852836763784, 60 7.802134559422377, 60 6.657099879646016, 60 6.510515132098641, 60 0, 55.31611974965083 0, 50.00444661661829 0, 43.31611974965083 0, 0 0, 0 14.034613342373582, 0 17.92266612923108, 0 21.587486526024364, 0 25.950779066386122, 0 26.860827855779647)),
```
pramsey, line 35: First coordinate in ring is different? Seems odd?

tests/unit/algorithm/RobustLineIntersectionTest.cpp
```diff
@@ -452,7 +445,7 @@ void object::test<10>
         "LINESTRING (-58.00593335955 -1.43739086465, -513.86101637525 -457.29247388035)",
         "LINESTRING (-215.22279674875 -158.65425425385, -218.1208801283 -160.68343590235)",
         1,
-        "POINT ( -215.22279674875 -158.65425425385 )",
+        "POINT ( -215.22279674875003 -158.65425425385004 )",
```
pramsey, line 448: Feels like a case for rounding? I guess not until we have some other platform that's got a slightly different answer...?

tests/xmltester/tests/issue/issue-geos-398.xml
dr-jts: Might just be a bit of non-determinism.  Will investigate.

tests/unit/algorithm/RobustLineIntersectionTest.cpp
dr-jts: Agreed, this would be more robust if the test specified a precision for the result.  More so in GEOS than JTS, since  there is more chance of platform-related discrepancy.

dr-jts:
Unfortunately this does not fix #740.  Perhaps the larger size of the ordinate values defeats the greater accuracy of the DD intersection computation.

tests/unit/algorithm/RobustLineIntersectionTest.cpp
dr-jts: Fixed by adding a distance tolerance.

tests/xmltester/tests/issue/issue-geos-398.xml
dr-jts: I think there was a report that overlay would "rotate" results by a single coordinate over repeated operations.  Perhaps this is the same phenomenon at work.

dr-jts:
This does not fix #766.

Issue #946, opened 2023-08-18
Crash in concaveHullByLengthRatio


pramsey:
Reported on [postgis-users](https://lists.osgeo.org/pipermail/postgis-users/2023-August/046039.html) and confirmed directly in GEOS. Here's a test case
```patch
diff --git a/tests/unit/algorithm/hull/ConcaveHullTest.cpp b/tests/unit/algorithm/hull/ConcaveHullTest.cpp
index f805a3d79..49e476300 100644
--- a/tests/unit/algorithm/hull/ConcaveHullTest.cpp
+++ b/tests/unit/algorithm/hull/ConcaveHullTest.cpp
@@ -298,6 +298,16 @@ void object::test<19>()
         "POLYGON ((20 90, 40 96, 56 95, 80 90, 90 70, 95 45, 90 20, 80 10, 60 15, 45 5, 20 10, 10 20, 5 40, 11 60, 20 70, 20 90), (40 80, 15 45, 21 30, 40 20, 70 20, 80 40, 80 60, 70 80, 40 80))" );
 }
 
+// postgis-users report
+template<>
+template<>
+void object::test<20>()
+{
+    checkHullByLengthRatio(
+        "MULTIPOINT ((113.56577197798602 22.80081530883069),(113.565723279387 22.800815316487014),(113.56571548761124 22.80081531771092),(113.56571548780202 22.800815317674463),(113.56577197817877 22.8008153088047),(113.56577197798602 22.80081530883069))",
+        0.75,
+        "POLYGON EMPTY"); // Change this
+}
 
```

dr-jts:
Confirmed this is an issue in JTS as well.

The problem is that the input points are almost "flat".  This causes very thin triangles, and the triangulation algorithm is unable to form a valid triangulation.

For such a small geometry it's hard to see why a concave hull would be required, as opposed to simply the convex hull.  Perhaps this is simply a single pathological case in a set of more reasonable inputs?

The problem should be detectable and be converted to an exception.  Given this, potential workarounds are:
* simply return the convex hull
* use a more robust (but slower) triangulation algorithm (this is experimental)

dr-jts:
The crash stack:
```
#0  0x0000ffff84e02210 in geos::triangulate::tri::Tri::getIndex(geos::triangulate::tri::Tri const*) const () from /home/[erchen.cz/pgsql15/lib/libgeos.so.3.11.2](http://erchen.cz/pgsql15/lib/libgeos.so.3.11.2)
[...]

Issue #719, opened 2022-11-01
geos::algorithm::hull::ConcaveHull::concaveHullByLength segfault


orca-hydromancer:
In libgeos v3.11.0 the following code results in segfault.

```c++
auto g = geos::io::WKTReader().read("MULTIPOINT( 1139294.6389832512941211 8201313.5346954688429832, 1139360.8549531854223460 8201271.1898052766919136, 1139497.5995843114797026 8201199.9955425458028913, 1139567.7837303513661027 8201163.3485335074365139, 1139635.3942210066597909 8201119.9025274068117142 )");
geos::algorithm::hull::ConcaveHull::concaveHullByLength( g.get(), 240, false );
```

Callstack:
```
geos::triangulate::tri::Tri::getIndex(const geos::triangulate::tri::Tri * const this, const geos::triangulate::tri::Tri * tri) (/externals/geos/src/triangulate/tri/Tri.cpp:312)
geos::algorithm::hull::HullTriangulation::nextBorderTri(geos::algorithm::hull::HullTri * triStart) (/externals/geos/src/algorithm/hull/HullTriangulation.cpp:161)
geos::algorithm::hull::HullTriangulation::traceBoundary(geos::triangulate::tri::TriList<geos::algorithm::hull::HullTri> & triList) (/externals/geos/src/algorithm/hull/HullTriangulation.cpp:129)
geos::algorithm::hull::HullTriangulation::traceBoundaryPolygon(geos::triangulate::tri::TriList<geos::algorithm::hull::HullTri> & triList, const geos::geom::GeometryFactory * factory) (/externals/geos/src/algorithm/hull/HullTriangulation.cpp:105)
geos::algorithm::hull::ConcaveHull::toGeometry(geos::algorithm::hull::ConcaveHull * const this, geos::triangulate::tri::TriList<geos::algorithm::hull::HullTri> & triList, const geos::geom::GeometryFactory * factory) (/externals/geos/src/algorithm/hull/ConcaveHull.cpp:375)
geos::algorithm::hull::ConcaveHull::getHull(geos::algorithm::hull::ConcaveHull * const this) (/externals/geos/src/algorithm/hull/ConcaveHull.cpp:152)
geos::algorithm::hull::ConcaveHull::concaveHullByLength(const geos::geom::Geometry * geom, double maxLength, bool holesAllowed) (/externals/geos/src/algorithm/hull/ConcaveHull.cpp:80)
```

This is what the multipoint looks like
[...]

PR #953, opened 2023-09-05
Fix IncrementalDelaunayTriangulator to ensure triangulation boundary is convex

dr-jts:
Port of https://github.com/locationtech/jts/pull/1004

This adds logic to `IncrementalDelaunayTriangulator` to ensure that the triangulation boundary is always convex (as per specification). This is required because the "frame" created to contain the triangulation construction is not infinitely distant from the input points. This has been causing robustness failures in cases where a site lies very close to the convex hull of the input.

This is a robust enhancement that replaces the frame size heuristic change in #728.

Also fixes issues reported for ConcaveHull ( #719, #946)


src/triangulate/IncrementalDelaunayTriangulator.cpp
```diff
@@ -82,25 +91,76 @@ IncrementalDelaunayTriangulator::insertSite(const Vertex& v)
...
+    }
+    return false;
+}
+
+bool 
+IncrementalDelaunayTriangulator::isConcaveAtOrigin(const QuadEdge& e) 
+{
+    Coordinate p = e.orig().getCoordinate();
+    Coordinate pp = e.oPrev().dest().getCoordinate();
+    Coordinate pn = e.oNext().dest().getCoordinate();
```
dbaston, line 150: Use `const Coordinate&` here

src/triangulate/IncrementalDelaunayTriangulator.cpp
```diff
@@ -82,25 +91,76 @@ IncrementalDelaunayTriangulator::insertSite(const Vertex& v)
...
+    bool isConcave = Orientation::COUNTERCLOCKWISE == Orientation::index(pp, pn, p);
+    return isConcave;
+}
+
+bool 
+IncrementalDelaunayTriangulator::isBetweenFrameAndInserted(const QuadEdge& e, const Vertex& vInsert) 
+{
+    const Vertex v1 = e.oNext().dest();
+    const Vertex v2 = e.oPrev().dest();
```
dbaston, line 159: `const Vertex&`

include/geos/triangulate/IncrementalDelaunayTriangulator.h
```diff
@@ -77,6 +91,14 @@ class GEOS_DLL IncrementalDelaunayTriangulator {
      * @return a quadedge containing the inserted vertex
      */
     quadedge::QuadEdge& insertSite(const quadedge::Vertex& v);
+
+private:
+
+    bool isConcaveBoundary(const quadedge::QuadEdge& e);
```
dbaston, line 97: methods can be `const`

dr-jts:
`const` comments fixed in https://github.com/libgeos/geos/commit/1033161eca3684e0d9479daeafd35a4e0fb1800e.

Thanks.

PR #956, opened 2023-09-08
Improve scale handling for GEOS_setPrecision

dr-jts:
## Problem

`GEOS_setPrecision` accepts a parameter specifying the **grid size** of the desired precision.  The most common usage is to reduce the number of decimal places in a geometry (e.g. to 3 decimal places).  In this case, the grid size is a negative power of 10 (e.g. 0.001).  However, this can result in inexact rounding of ordinate values.

For example, this `geosop` command reproduces this [PostGIS issue](https://trac.osgeo.org/postgis/ticket/5520#comment:2) and shows how reducing the precision of a geometry using grid size 0.001 produces unwanted "extra decimals" (note that in `geosop` the `reducePrecision` operation parameter is negative to indicate the value is a grid size):
```
bin/geosop -a "LINESTRING (657035.913 6475590.114,657075.57 6475500)" reducePrecision N-0.001

LINESTRING (657035.9130000001 6475590.114, 657075.5700000001 6475500)
```
The same issue can happen when using a **scale factor** that is a negative power of 10:
```
bin/geosop -a "POINT (674169.89 198051.619820510769063)" reducePrecision .00001
LINESTRING (700000 199999.99999999997, 700000 1400000)
```

## Cause

There are two causes of this problem:
* computing the scale factor as the reciprocal of the grid size may be inexact for some values (i.e. the reciprocal of a fractional power of 10 may not be an exact integral value)
* the [arithmetic](https://github.com/libgeos/geos/blob/main/src/geom/PrecisionModel.cpp#L62) used to scale values is subject to rounding issues when non-integral values are used:
```
return util::round(val / gridSize) * gridSize;
```
Because negative powers of 10 [cannot be represented exactly in binary FP numbers](https://stackoverflow.com/questions/1089018/why-cant-decimal-numbers-be-represented-exactly-in-binary)  this can cause the resulting "rounded" ordinate values to be inexact.

In contrast, if an integral scale factor is used directly, the [rounding arithmetic](https://github.com/libgeos/geos/blob/main/src/geom/PrecisionModel.cpp#L59) is:
```
return util::round(val * scale) / scale;
```
This produces the desired result:
```
bin/geosop -a "LINESTRING (657035.913 6475590.114,657075.57 6475500)" reducePrecision 1000   

LINESTRING (657035.913 6475590.114, 657075.57 6475500)
```

> It appears that in FP arithmetic multiplying by an integer value is more accurate than dividing by an (approximately equal) fractional value

## Proposed Fix
This PR proposes a fix which has **no impact** on the GEOS API or semantics (and hence on downstream usage).  

The fix does two things:
* It detects when the PrecisionModel scale factor OR the equivalent grid size is "essentially integral" (i.e. very close to an integer).  If one of these is the case, then the exact integer value of grid size or scale factor is used. 
* The arithmetic to adjust ordinate values uses the larger of the scale factor and grid size.  This ensures that an integral grid size or scale factor is used if present (which is almost always the case).  This produces a more exact scaled value (thanks to the magic voodoo of FP arithmetic).

Almost all (or all?) uses of `GEOS_setPrecision` use either grid sizes equivalent to integral scale factors OR grid sizes which are obviously NOT integral (e.g. say `0.5`),  so the heuristic of deciding "close" is fine for practical purposes.

In the very rare cases where both grid size and scale factor are fractional, the original values are used for scaling.  Since they are fractional already, the output won't be exact in any case.

As a nice side effect, this fix allows dropping support for negative scale factors, which was introduced solely to accommodate the use of grid size rather than scale factor.  (This PR will be extended to delete this code if that is agreed to).

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -2980,12 +2980,14 @@ extern "C" {
         return execute(extHandle, [&]() {
             PrecisionModel newpm;
             if (gridSize != 0) {
-                //-- check for an integral scale
                 double scale = 1.0 / gridSize;
-                double scaleInt = std::round(scale);
-                //-- if scale factor is essentially integral, use the exact integer value
-                if (std::abs(scale - scaleInt) < GRIDSIZE_INTEGER_TOLERANCE) {
-                    scale = scaleInt;
+                //-- check if the gridsize corresponds to an integral scale
+                if (gridSize < 1) {
+                    double scaleInt = std::round(scale);
+                    //-- if scale factor is essentially integral, use the exact integer value
+                    if (std::abs(scale - scaleInt) < GRIDSIZE_INTEGER_TOLERANCE) {
+                        scale = scaleInt;
+                    }
                 }
                 newpm = PrecisionModel(scale);
             }
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -2982,8 +2982,7 @@ extern "C" {
             if (gridSize != 0) {
                 //-- check for an integral scale
                 double scale = 1.0 / gridSize;
-                //-- add a small "bump" to ensure flooring works
-                double scaleInt = std::floor(scale + 2 * GRIDSIZE_INTEGER_TOLERANCE);
+                double scaleInt = std::round(scale);
                 //-- if scale factor is essentially integral, use the exact integer value
                 if (std::abs(scale - scaleInt) < GRIDSIZE_INTEGER_TOLERANCE) {
                     scale = scaleInt;
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -2978,9 +2978,16 @@ extern "C" {
         return execute(extHandle, [&]() {
             PrecisionModel newpm;
             if(gridSize != 0) {
-                // Use negative scale to indicate you actually want a gridSize
-                double scale = -1.0 * std::abs(gridSize);
-                newpm = PrecisionModel(scale);
+                double scaleNom = 1.0 / gridSize;
+                double scaleInt = std::floor(scaleNom);
+                if (std::abs(scaleNom - scaleInt) < 1e-5) {
+                    //-- if scale factor is nearly integral, use an exact integer value
+                    newpm = PrecisionModel(scaleInt);
+                }
+                else {
+                    //-- otherwise, just use the reciprocal of the grid size
+                    newpm = PrecisionModel(scaleNom);
+                }
             }
 
             const PrecisionModel* pm = g->getPrecisionModel();
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -2979,17 +2979,16 @@ extern "C" {
 
         return execute(extHandle, [&]() {
             PrecisionModel newpm;
-            if(gridSize != 0) {
-                double scaleNom = 1.0 / gridSize;
-                double scaleInt = std::floor(scaleNom);
-                if (std::abs(scaleNom - scaleInt) < GRIDSIZE_INTEGER_TOLERANCE) {
-                    //-- if scale factor is nearly integral, use an exact integer value
-                    newpm = PrecisionModel(scaleInt);
+            if (gridSize != 0) {
+                //-- check for an integral scale
+                double scale = 1.0 / gridSize;
+                //-- add a small "bump" to ensure flooring works
+                double scaleInt = std::floor(scale + 0.1);
+                //-- if scale factor is essentially integral, use the exact integer value
+                if (std::abs(scale - scaleInt) < GRIDSIZE_INTEGER_TOLERANCE) {
+                    scale = scaleInt;
                 }
-                else {
-                    //-- otherwise, just use the reciprocal of the grid size
-                    newpm = PrecisionModel(scaleNom);
-                }
+                newpm = PrecisionModel(scale);
             }
 
             const PrecisionModel* pm = g->getPrecisionModel();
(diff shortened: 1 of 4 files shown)


dr-jts:
I wonder if the inverse problem may occur as well: an integral grid size (e.g. 1000) would produce undesired "slop" when converted to a reciprocal scale factor?

This might require building a fuzzer to test a lot of values to see if this occurs.  It's possible that the guard digits in FP arithmetic will handle this problem.

capi/geos_ts_c.cpp
```diff
@@ -2977,9 +2979,15 @@ extern "C" {
...
         return execute(extHandle, [&]() {
             PrecisionModel newpm;
-            if(gridSize != 0) {
-                // Use negative scale to indicate you actually want a gridSize
-                double scale = -1.0 * std::abs(gridSize);
+            if (gridSize != 0) {
+                //-- check for an integral scale
+                double scale = 1.0 / gridSize;
+                //-- add a small "bump" to ensure flooring works
+                double scaleInt = std::floor(scale + 2 * GRIDSIZE_INTEGER_TOLERANCE);
```
dbaston, line 2986: I'm probably missing something obvious, but why not `std::round` here?

dr-jts: yes, much better idea.

pramsey (approved):
Looks OK to me. How do we feel about the behaviour when gridSize is > 0 (1, 10, 100, etc)?

dr-jts:
> How do we feel about the behaviour when gridSize is > 0 (1, 10, 100, etc)?

Hmm..  maybe should only do the integral test when the gridSize is < 1.

strk:
Seems ok to me, now I wonder how results compare with ST_SnapToGrid

dr-jts:
>  I wonder how results compare with ST_SnapToGrid

`ST_SnapToGrid` has a similar issue - see [PG-5425](https://trac.osgeo.org/postgis/ticket/5425).

I will add that data as a test case.

mwtoews:
> `ST_SnapToGrid` has the same issue - see [PG-5425](https://trac.osgeo.org/postgis/ticket/5425) and [PG-3929](https://trac.osgeo.org/postgis/ticket/3929). That PostGIS code could benefit from a similar fix.

Also recently reported on [postgis-users](https://lists.osgeo.org/pipermail/postgis-users/2023-September/046107.html).
As @pramsey [noted](https://lists.osgeo.org/pipermail/postgis-users/2023-September/046109.html) a workaround for PostGIS is to use `ST_AsText(g1, maxdecimaldigits = 3)`.
A similar workaround for GEOS is to use `GEOSWKTWriter_setRoundingPrecision()` / `writer->setRoundingPrecision()`

mwtoews:
Another quirk that could be worth fixing here or in a separate issue is `0.001` or `N-1000`:
```
$ ./bin/geosop -a "LINESTRING (657035.913 6475590.114,657075.57 6475500)" reducePrecision 0.001
LINESTRING Z EMPTY
```
where the expected result should have been:

    LINESTRING (657000 6476000, 657000 6476000)

And what should happen with `N-100000000`? `LINESTRING EMPTY`? or `LINESTRING (0 0, 0 0)`?

dr-jts:
> Also recently reported on [postgis-users](https://lists.osgeo.org/pipermail/postgis-users/2023-September/046107.html). 

Good find. There are lots of examples of this occurring in `ST_SnapToGrid`!  All solved by this fix.

dr-jts:
> Another quirk that could be worth fixing here or in a separate issue 

@mwtoews This is a different problem, so should be a separate issue.  (Is there already an issue for this?)

Technically zero-length linestrings are invalid, but it seems good to bend the rules in this case.  

> what should happen with N-100000000? LINESTRING EMPTY? or LINESTRING (0 0, 0 0)?

Probably the latter - so as to not lose information.

dr-jts:
> How do we feel about the behaviour when gridSize is > 0 (1, 10, 100, etc)?

I have found a case where a large gridsize (100,000) causes rounding error when using the reciprocal scale factor.

Using the gridsize 100,000 directly works:
```
bin/geosop -a "LINESTRING(674169.89 198051.619820510769063, 674197.71234 1448065.55674200)" -f wkt reducePrecision N-100000
LINESTRING (700000 200000, 700000 1400000)
```
Using the reciprocal scale factor of 0.00001 causes rounding error:
```
bin/geosop -a "LINESTRING(674169.89 198051.619820510769063, 674197.71234 1448065.55674200)" -f wkt reducePrecision .00001
LINESTRING (700000 199999.99999999997, 700000 1400000)
```

To resolve this, gridsizes > 1 need to be computed using the more stable arithmetic formula:
```
return util::round(val / gridSize) * gridSize;
```
Also, it's best to move the integral rounding logic into `PrecisionModel`, so it is widely available.

capi/geos_ts_c.cpp
```diff
@@ -2969,6 +2969,8 @@ extern "C" {
         });
     }
 
+    //const double GRIDSIZE_INTEGER_TOLERANCE = 1e-5;
```
pramsey, line 2972: Remove dead code.

src/geom/PrecisionModel.cpp
```diff
@@ -85,7 +91,7 @@ PrecisionModel::PrecisionModel(Type nModelType)
     :
     modelType(nModelType),
     scale(1.0),
-    gridSize(0.0)
+    gridSize(1.0)
```
pramsey, line 94: This has no knock-on effects? Did 0.0 have a special meaning before?

capi/geos_ts_c.cpp
dr-jts: gotcha

src/geom/PrecisionModel.cpp
dr-jts: Yes, before gridsize = 0.0 indicated that the scale value should be used.  But that logic has been changed to just detect which of scale or gridsize are larger, and use that.
 
Scales/gridsizes of 0 are disallowed implicitly.  But the code handles a bad input of 0 as a no-op

PR #959, opened 2023-09-17
Fix PreparedLineStringDistance for lines within envelope and polygons

### LINKED ISSUE #958: GEOSPreparedDistanceWithin fails when comparing an interior subset of a line to the full line
Given two lines where the subset is at least 3 vertices in from the endpoints of the larger line, GEOSPreparedDistanceWithin returns false when querying the subset line against the larger line, whereas GEOSDistanceWithin returns true.

Example using Shapely 2.0:

```Python
line1 = shapely.from_wkt('LINESTRING (0 0, 1 1, 2 2, 3 3, 4 4, 5 5, 6 6, 7 7, 8 8, 9 9)')
line2 = shapely.from_wkt('LINESTRING (2 2, 3 3, 4 4, 5 5, 6 6, 7 7)')

print(shapely.dwithin(line2, line1, 1))  # uses GEOSDistanceWithin_r
# True

shapely.prepare(line2)  # uses GEOSPrepare_r
print(shapely.dwithin(line2, line1, 1))  # uses GEOSPreparedDistanceWithin_r
# False

# longer subset works fine
line3 = shapely.from_wkt('LINESTRING (2 2, 3 3, 4 4, 5 5, 6 6, 7 7, 8 8)')

print(shapely.dwithin(line3, line1, 1))
# True

shapely.prepare(line3)
print(shapely.dwithin(line3, line1, 1))
# True
```

First observed in shapely [#1887](https://github.com/shapely/shapely/issues/1887) using GEOS `main` (3.13.0dev), Shapely `main`, on MacOS.

### LINKED ISSUE #960: PreparedLineStringDistance.isWithinDistance fails for containing polygons
As a follow-on to #958, `PreparedLineStringDistance.isWithinDistance` (and hence `GEOSPreparedDistanceWithin`) does not work for LineStrings when the test geometry is a Polygon which contains the indexed line(s).  This is because the `PreparedLineStringDistance.isWithinDistance` does not check if the test geometry contains any points of the indexed geometry.

Here's a test case:
```
template<>
template<>
void object::test<13>
()
{
    checkDistanceWithin(
        "LINESTRING (30 30, 70 70)",
        "POLYGON ((0 100, 100 100, 100 0, 0 0, 0 100))",
        1,
        1
    );
}
```

dr-jts:
This PR fixes `PreparedLineStringDistance` to handle the cases:

1. the prepared line is contained within the envelope of the test geometry
2. the prepared line is contained within a polygonal geometry

Fixing the first situation requires removing an invalid heuristic check from `IndexedFacetDistance.isWithinDistance` [code](https://github.com/libgeos/geos/blob/main/src/operation/distance/IndexedFacetDistance.cpp#L69):
```
    auto env2 = g->getEnvelope();
    if (distance(env2.get()) > maxDistance) {
        return false;
    }
```
The check attempts to short-circuit the withinDistance test by checking the distance to the envelope polygon of the tested geometry. However, `IndexedFacetDistance` only computes distances to linework, not to polygonal geometries. So the test is incorrect in some cases (such as #958).

This heuristic is not in the JTS [code](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/distance/IndexedFacetDistance.java#L183).

Fixing the second situation requires an additional point-in-polygon check in 
`PreparedLineStringDistance.distance` and `PreparedLineStringDistance.isWithinDistance` to test for the prepared line (or element of) lying within a test polygonal geometry.

Fixes #958.
Fixes #960.

### DIFF
--- a/src/geom/prep/PreparedLineStringDistance.cpp
+++ b/src/geom/prep/PreparedLineStringDistance.cpp
@@ -32,12 +32,18 @@ PreparedLineStringDistance::distance(const geom::Geometry* g) const
         return DoubleInfinity;
     }
 
-    // TODO: test if this shortcut be any useful
-    //if ( prepLine.intersects(g) ) return 0.0;
-
-    /* Not intersecting, compute distance from facets */
+    /* Compute potential distance from facets */
     operation::distance::IndexedFacetDistance *idf = prepLine.getIndexedFacetDistance();
-    return idf->distance(g);
+    double dist = idf->distance(g);
+
+    // If any point from prepLine is contained by g, the distance is zero
+    // Do this last because this PIP test is not indexed.
+    if ( g->getDimension() == 2 
+            && dist > 0 
+            && prepLine.isAnyTargetComponentInTest(g)) {
+        return 0.0;
+    }
+    return dist;
 }
 
 bool
--- a/src/geom/prep/PreparedLineStringDistance.cpp
+++ b/src/geom/prep/PreparedLineStringDistance.cpp
@@ -49,7 +49,15 @@ PreparedLineStringDistance::isWithinDistance(const geom::Geometry* g, double d) 
     }
 
     operation::distance::IndexedFacetDistance *idf = prepLine.getIndexedFacetDistance();
-    return idf->isWithinDistance(g, d);
+    bool withinDistance = idf->isWithinDistance(g, d);
+
+    // If any point from prepLine is contained by g, the distance is zero
+    // Do this last because this PIP test is not indexed.
+    if ( g->getDimension() == 2 && ! withinDistance) {
+        return prepLine.isAnyTargetComponentInTest(g);
+    }
+
+    return withinDistance;
 }
 
 
--- a/src/geom/prep/PreparedLineStringDistance.cpp
+++ b/src/geom/prep/PreparedLineStringDistance.cpp
@@ -35,11 +35,12 @@ PreparedLineStringDistance::distance(const geom::Geometry* g) const
     /* Compute potential distance from facets */
     operation::distance::IndexedFacetDistance *idf = prepLine.getIndexedFacetDistance();
     double dist = idf->distance(g);
+    if (dist == 0.0)
+        return 0.0;
 
     // If any point from prepLine is contained by g, the distance is zero
     // Do this last because this PIP test is not indexed.
     if ( g->getDimension() == 2 
-            && dist > 0 
             && prepLine.isAnyTargetComponentInTest(g)) {
         return 0.0;
     }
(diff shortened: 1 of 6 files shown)


dr-jts:
The heuristic could be kept if it is only applied to test geometries whose envelope does not contain the indexed geometry envelope.  Not sure if this effort is worth it though?

dbaston:
Is it sufficient to guard the test with

```
    if (!g->getEnvelopeInternal()->contains(baseGeometry.getEnvelopeInternal()))
    {
        auto env2 = g->getEnvelope();
        if (distance(env2.get()) > maxDistance) {
            return false;
        }
    }
```

dr-jts:
> Is it sufficient to guard the test with...

No, because the indexed geometry `baseGeometry` might have two elements.  So its envelope might not be contained in the envelope of a polygon `g`, but one of the elements could still be contained in `g` and hence have distance = 0.

dbaston:
Do you have a test case that would fail? I'm sorry, I've spent a good 20 minutes going over this and I can't understand the problem.

dr-jts:
> Do you have a test case that would fail? 

My bad - this does work, with the addition of the PIP check in `PreparedLineStringDistance.isWithinDistance`.  I will add the test case in `GEOSPreparedDistanceWithin`.

```
    checkDistanceWithin(
        "MULTILINESTRING ((1 6, 1 1), (14 16, 16 14))",
        "MULTIPOLYGON (((10 20, 20 20, 20 10, 10 10, 10 20)), ((30 20, 40 20, 40 10, 30 10, 30 20)))",
        1,
        1
    );
```
image

It's a bit confusing, because the `IndexedFacetDistance.isWithinDistance` actually returns `false` for this case.  It would be good to have unit tests for `IndexedFacetDistance.isWithinDistance`.

dr-jts:
> Do you have a test case that would fail? 

Actually while this heuristic works in the context of `PreparedLineStringDistance` (which has the extra PIP test), I'm not convinced it works in all cases for `IndexedFacetDistance.isWithinDistance` itself.  I'll explore this further and make a `IndexedFacetDistance` test case.

dbaston:
I think `baseGeom` is either:
- inside the envelope of `g`: skip the heuristic. This check was missing in the original implementation.
- partially inside the envelope of `g`: computed distance will be zero, which is not greater than`maxDistance`
- outside the envelope of `g`: if the envelope of `g` is > `maxDistance` from `baseGeom`, than any point on `g` is >= `maxDistance` from `baseGeom` so we can return false and avoid building the whole tree on `g`.

dr-jts:
> * partially inside the envelope of `g`: computed distance will be zero, which is not greater than`maxDistance`

If the indexed geometry has multiple elements, the computed distance to `g.env` may be > 0.

dr-jts:
I'd like to commit this particular fix, and move the issue of the correctness of `IndexedFacetDistance` to a new issue.  For now I will remove the heuristic (since IMO there is still a question of correctness).  It can be reinstated if it can be demonstrated to work later on.

dbaston:
> If the indexed geometry has multiple elements, the computed distance to g.env may be > 0.

You're right, the heuristic can be wrong if `baseGeom` is a multipart. I would suggest changing it to

```
    if (baseGeometry.getNumGeometries() == 1 && !g->getEnvelopeInternal()->contains(baseGeometry.getEnvelopeInternal()))
    {
        auto env2 = g->getEnvelope();
        if (distance(env2.get()) > maxDistance) {
            return false;
        }
    }
```

and adding the following test to `IndexedFacetDistance` (demonstrating the need for the `getNumGeometries()` condition)

```
template<>
template<>
void object::test<13>()
{
    auto g1 = _wktreader.read("MULTILINESTRING ((6 6, 6.1 6.1), (100 100, 101 101))");
    auto g2 = _wktreader.read("LINESTRING (0 10, 10 0)");

    IndexedFacetDistance ifd(g1.get());

    ensure(ifd.isWithinDistance(g2.get(), 2));
}
```

PR #961, opened 2023-09-19
fix View Source link on home page (points to postgis)

mdsumner:
View Source on home page https://libgeos.org/  points to postgis


mdsumner:
thank you

PR #966, opened 2023-09-27
python-software-properties has been deprecated

JamesParrott:
I suggest it is replaced with https://packages.ubuntu.com/kinetic/software-properties-common

python3-software-properties may be preferrable (less to download) but I've not tested it.

https://packages.ubuntu.com/kinetic/python3-software-properties


pramsey:
Thanks!

Issue #979, opened 2023-10-31
ST_Distance doesn't return the minimum distance.


cuteDen-ECNU:
Consider this statement:
```sql
SELECT ST_Distance(a2, a1), ST_Distance(a1, a2)
FROM ST_GeomFromText('MULTIPOINT((-2 0), EMPTY)') As a1
,ST_GeomFromText(' GEOMETRYCOLLECTION(POINT(1 0),LINESTRING(0 0,1 0))') As a2;
--actual{3,2}
--expected{2,2}
```
According to the following definition, ST_Distance returns the minimum distance between two geometries :
> returns the minimum 2D Cartesian (planar) distance between two geometries

The minimum distance between a1 and a2 is 2 instead of 3. But the result of ST_Distance(a2, a1) is 3, which is not the minimum distance.
Meanwhile,  the result of ST_Distance(a2, a1) is not same as ST_Distance( a1, a2).
So I believe it is a functional issue that the current test suites haven't explored.

According to the crash https://github.com/libgeos/geos/pull/977, the function related to the GEOMETRYCOLLECTION with EMPTY seems not to be explored by before test suites?

The version is the newest in GitHub:
POSTGIS="3.5.0dev 3.4.0rc1-705-g5c3ec8392" [EXTENSION] PGSQL="170" GEOS="3.13.0dev-CAPI-1.18.0" PROJ="8.2.1 NETWORK_ENABLED=OFF URL_ENDPOINT=https://cdn.proj.org/ USER_WRITABLE_DIRECTORY=/tmp/proj DATABASE_PATH=/usr/share/proj/proj.db" LIBXML="2.9.13"

pramsey:
Distance in PostGIS is not delegated to GEOS, so this issue should properly be ticketed there. That said, I will add a test case here before closing, just in case it's also an issue in GEOS.

pramsey:
Well, your particular example crashes GEOS, and in general the question of "what is the distance to EMPTY" seems a little fraught. Distance between `POINT EMPTY` and `LINESTRING EMPTY` is zero. Distance between `POINT EMPTY` and 'POINT(1 1)' is zero. Distance between `GEOMETRYCOLLECTION(LINESTRING EMPTY, POINT(1 1))` and `POINT(2 1)` is ... one.

https://trac.osgeo.org/postgis/wiki/DevWikiEmptyGeometry

[...]

PR #980, opened 2023-10-31
Skip over testing empty distances for mixed collections

pramsey:
This leaves the inconsistency that pure empty/empty distance tests will return zero while mixed tests with some real distances will return those non-zero distances. References #979


dr-jts:
JTS fix is https://github.com/locationtech/jts/pull/1010

Might want to pull the updated XML test file into GEOS?

PR #994, opened 2023-11-15
Support id for GeoJSONFeature

kunlinyu:
1. Add id and getter for GeoJSONFeature
2. Modify the copy constructor of GeoJSONFeature
3. Check "id" in GeoJSONReader
4. Check "id" in GeoJSONWriter
5. Modify unit test for id

### DIFF
--- a/src/io/GeoJSONReader.cpp
+++ b/src/io/GeoJSONReader.cpp
@@ -102,7 +102,7 @@ GeoJSONFeature GeoJSONReader::readFeature(const geos_nlohmann::json& j) const
     std::string id = "";
     if (j.contains("id") && !j.at("id").is_null()) {
         if (j.at("id").is_string()) id = j.at("id").get<std::string>();
-        if (j.at("id").is_number()) id = std::to_string(j.at("id").get<int>());
+        if (j.at("id").is_number()) id = j.at("id").dump();
     }
 
     return GeoJSONFeature{readGeometry(geometryJson), readProperties(properties), id};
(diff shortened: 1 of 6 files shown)


src/io/GeoJSONReader.cpp
```diff
@@ -98,7 +98,11 @@ GeoJSONFeature GeoJSONReader::readFeature(const geos_nlohmann::json& j) const
 {
     const auto& geometryJson = j.at("geometry");
     const auto& properties = j.at("properties");
-    return GeoJSONFeature{readGeometry(geometryJson), readProperties(properties)};
+
+    std::string id = "";
+    if (j.contains("id")) id = j.at("id").get<std::string>();
```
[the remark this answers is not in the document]
pramsey: yes, seems worrisome :)

src/io/GeoJSONReader.cpp
```diff
@@ -98,7 +98,14 @@ GeoJSONFeature GeoJSONReader::readFeature(const geos_nlohmann::json& j) const
 {
     const auto& geometryJson = j.at("geometry");
     const auto& properties = j.at("properties");
-    return GeoJSONFeature{readGeometry(geometryJson), readProperties(properties)};
+
+    std::string id = "";
+    if (j.contains("id") && !j.at("id").is_null()) {
+        if (j.at("id").is_string()) id = j.at("id").get<std::string>();
+        if (j.at("id").is_number()) id = std::to_string(j.at("id").get<int>());
```
rouault, line 105: does get<int>() works fine on floating-point numbers ?

kunlinyu: It works fine but I think it is not correct. According to the geojson standard, all number is valid. So we have to support float point. I will upload a patch soon.

rouault (approved):

Issue #993, opened 2023-11-15
Empty LineString created with HasZ() is true


ddowling:
I am trying to track down an issue in recent versions of GeoDjango where empty `LineString` objects are being created with the `hasz` property True. The following code is outputting True with libgeos version 3.12.0_CAPI_1.18.0. When run with 3.11.2-CAPI-1.17.2 the code putputs False

```python
from django.contrib.gis.geos import LineString, geos_version
print(geos_version())
ls = LineString()
print(ls.hasz)
```

I have traced this down to the C `libgeos` API where `GEOSGeom_createLineString` is getting called with a `NULL` argument. 
Following is a short C program that demonstrates this. When run with version 3.12.0 of libgeos it outputs GEOSHasZ() = 1. With version 3.11.0 it outputs GEOSHasZ() = 0.

```c
#include <stdio.h>
#include <stdarg.h>

#include <geos_c.h>

static void geos_msg_handler(const char* fmt, ...)
{
    va_list ap;
    va_start(ap, fmt);
    vprintf (fmt, ap);
    va_end(ap);
}

int main()
{
    /* Send notice and error messages to the terminal */
    initGEOS(geos_msg_handler, geos_msg_handler);

    printf("GEOS Version %s\n", GEOSversion());

    GEOSGeometry* geom = GEOSGeom_createLineString(NULL);

    GEOSWKTWriter* writer = GEOSWKTWriter_create();
    char* wkt = GEOSWKTWriter_write(writer, geom);
    printf("Geometry: %s\n", wkt);
    printf("GEOSHasZ() = %d\n", GEOSHasZ(geom));

    /* Clean up allocated objects */
    GEOSWKTWriter_destroy( writer);
    GEOSGeom_destroy(geom);

    /* Clean up the global context */
    finishGEOS();

    return 0;
}

````

**Comment by ddowling** [comment]
Tested this with the code from the main branch
```
GEOS Version 3.13.0dev-CAPI-1.18.0
Geometry: LINESTRING Z EMPTY
GEOSHasZ() = 1
```

I would expect this code to create a 2D geometry.

pramsey:
I'm not able to replicate this on main or on 3.12 branch.

pramsey:
[...]

PR #998, opened 2023-11-20
Return 2D empty linestring on GEOSGeom_createLineString(NULL)

pramsey:
References GH-993

### DIFF
--- a/tests/unit/capi/GEOSHasZMTest.cpp
+++ b/tests/unit/capi/GEOSHasZMTest.cpp
@@ -106,14 +106,6 @@ void object::test<9>()
     ensure_equals(GEOSHasM(input_), 1);
 }
 
-template<>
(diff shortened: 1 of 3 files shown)


ddowling:
I can confirm that this change fixes the issue that I was originally seeing in Django with the default construction of LineString() objects.

pramsey:
OK w/ @dbaston ?

tests/unit/capi/GEOSHasZMTest.cpp
```diff
@@ -106,5 +106,14 @@ void object::test<9>()
     ensure_equals(GEOSHasM(input_), 1);
 }
 
+template<>
```
dbaston, line 109: Consider putting this in `GEOSGeom_createTest.cpp` ?

PR #1014, opened 2023-12-09
geos_c.cpp: fix declarations to avoid C++ One Definition Rule violations

rouault:
Currently building with GCC link-time-optimization with -flto -Wodr leads to the following type of warnings:
```
/home/even/geos/geos/build_cmake/capi/geos_c.h:665:31: warning: ‘GEOSBufferWithParams_r’ violates the C++ One Definition Rule [-Wodr]
  665 | extern GEOSGeometry GEOS_DLL *GEOSBufferWithParams_r(
      |                               ^
/home/even/geos/geos/capi/geos_ts_c.cpp:4034:5: note: type mismatch in parameter 3
 4034 |     GEOSBufferWithParams_r(GEOSContextHandle_t extHandle, const Geometry* g1, const BufferParameters* bp, double width)
      |     ^
/home/even/geos/geos/capi/geos_ts_c.cpp:4034:5: note: type name ‘geos::operation::buffer::BufferParameters’ should match type name ‘GEOSBufParams_t’
/home/even/geos/geos/capi/geos_ts_c.cpp:4034:5: note: ‘GEOSBufferWithParams_r’ was previously declared here
```

This is due to geos_c.cpp using slightly different definitions of the aliases between C API and C++ classes than the ones set in geos_ts_c.cpp.
Use exactly the ones from geos_tc_c.cpp to fix that issue

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -23,6 +23,7 @@
 #include <geos/io/WKBWriter.h>
 #include <geos/io/GeoJSONReader.h>
 #include <geos/io/GeoJSONWriter.h>
+#include <geos/operation/buffer/BufferParameters.h>
 #include <geos/util/Interrupt.h>
 
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


mwtoews:
I've drafted an unmerged post-install test for GEOS, and am intending to re-write another sometime. I suppose this link-time-optimization check could fit in that process as a CI check?

rouault:
> I suppose this link-time-optimization check could fit in that process as a CI check?

yes potentially, but LTO is for the main build, not a post-install test

PR #1025, opened 2024-01-09
Fix StringTokenizer::peekNextToken

### LINKED ISSUE #1013: Cannot parse EMPTY token with white space
This does not parse
```
MULTIPOINT( EMPTY, (10 10), (20 20))
```
this does
```
MULTIPOINT(EMPTY, (10 10), (20 20))
```
Here's a test case
```c++
// EMPTY token with some white space
template<>
template<>
void object::test<24>
()
{
    GeomPtr geom(wktreader.read("MULTIPOINT( EMPTY, (10 10), (20 20))"));
    auto ngeoms = geom->getNumGeometries();

    ensure(ngeoms == 3);
}
```

mwtoews:
This fixes `StringTokenizer::peekNextToken` so that both `(EMPTY` and `( EMPTY` are `StringTokenizer::TT_WORD`.

The issue was that the second assignment of `pos` (from `str.find_first_of`) was less than the first (from `str.find_first_not_of`), meaning that it started searching from the wrong position. The fix is simply to set the starting position to `pos + 1`.

Closes #1013

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@
   - Fix DiscreteHausdorffDistance for LinearRing (GH-1000, Martin Davis)
   - Fix IsSimpleOp for MultiPoint with empty element (GH-1005, Martin Davis)
   - Fix PreparedPolygonContains for GC with MultiPoint (GH-1008, Martin Davis)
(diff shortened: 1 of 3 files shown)


dbaston (approved):

mwtoews:
Backported to 3.12, 3.11, 3.10

Issue #1031, opened 2024-01-23
Unsupported geometry types in coverageSimplify result in segfaults


martinfleis:
`coverageSimplify` has a tendency to produce segfaults when given unsupported geometry types. I ran into it while doing the shapely wrapper in https://github.com/shapely/shapely/pull/1969.

See examples below:

```sh
% geosop -a "MULTILINESTRING ((200 100, 100 100, 200 200), (200 200, 200 100), (200 200, 300 100, 200 100))" -v -e coverageSimplify 10
Input A: WKT literal
Read 1 geometries, 8 vertices  -- 174 usec
zsh: segmentation fault  geosop -a  -v -e coverageSimplify 10
```

```sh
% geosop -a "MULTIPOINT ((0 0), (0 1), (1 1), (1 0))" -v -e coverageSimplify 2
Input A: WKT literal
Read 1 geometries, 4 vertices  -- 59 usec
zsh: segmentation fault  geosop -a "MULTIPOINT ((0 0), (0 1), (1 1), (1 0))" -v -e coverageSimplify 2
```

```sh
% geosop -a "GEOMETRYCOLLECTION (POLYGON ((0 0, 20 0, 20 10, 10 5, 0 10, 0 0)), LINEARRING (0 10, 10 5, 20 10, 20 20, 0 20, 0 10))" -v -e coverageSimplify 20 
Input A: WKT literal
Read 1 geometries, 12 vertices  -- 40 usec
zsh: segmentation fault  geosop -a  -v -e coverageSimplify 20
```

`CoverageUnion` results in `IllegalArgumentException: Unhandled geometry type in CoverageUnion.` which is what I'd expect here as well.

Using GEOS 3.12.1.

PR #1039, opened 2024-02-05
Fix segfault when coverage simplify called with nonpolygons

pramsey:
References GH-1031

### DIFF
--- a/src/coverage/CoverageSimplifier.cpp
+++ b/src/coverage/CoverageSimplifier.cpp
@@ -20,6 +20,7 @@
 #include <geos/geom/Geometry.h>
 #include <geos/geom/GeometryFactory.h>
 #include <geos/geom/MultiLineString.h>
+#include <geos/util/IllegalArgumentException.h>
 
 
 using geos::geom::Geometry;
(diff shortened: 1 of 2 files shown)


src/coverage/CoverageSimplifier.cpp
```diff
@@ -83,7 +84,12 @@ CoverageSimplifier::simplifyInner(
 CoverageSimplifier::CoverageSimplifier(std::vector<const Geometry*>& coverage)
     : m_input(coverage)
     , m_geomFactory(coverage.empty() ? nullptr : coverage[0]->getFactory())
-    {}
+    {
+        for (const Geometry* g: m_input) {
+            if (!g->isPolygonal())
```
jorisvandenbossche, line 89: I think `isPolygonal` can return true for GeometryCollections that contain polygons. But the simplifier code will still segfault for this as well (noticed while implementing the bindings in shapely).

Using geosop:

```bash
# GeometryCollection of polygons -> fine
$ geosop -a "GEOMETRYCOLLECTION (POLYGON ((0 10, 10 10, 10 7, 10 3, 10 0, 0 0, 0 10)), POLYGON ((10 10, 20 10, 20 0, 10 0, 10 3, 10 7, 10 10)))" coverageSimplify 1.0
POLYGON ((10 10, 10 0, 0 0, 0 10, 10 10))
POLYGON ((10 10, 20 10, 20 0, 10 0, 10 10))
# GeometryCollection of polygon and collection -> segfaults
$ geosop -a "GEOMETRYCOLLECTION (POLYGON ((0 10, 10 10, 10 7, 10 3, 10 0, 0 0, 0 10)), GEOMETRYCOLLECTION (POLYGON ((10 10, 20 10, 20 0, 10 0, 10 3, 10 7, 10 10))))" coverageSimplify 1.0
Segmentation fault (core dumped)
```

(and using a nested GeometryCollection here because we first collect the array of input geometries (which can contain a collection) in a collection)

jorisvandenbossche: Opened a new issue to track this -> https://github.com/libgeos/geos/issues/1250

PR #1042, opened 2024-02-06
CI: replace macos-11/xcode 13.2.1 config by ARM64 M1/xcode 14.3

rouault:
Some testing of non-Intel architecture...


pramsey:
Thanks! Good plan.

PR #1043, opened 2024-02-06
ci.yml: add automatic cancellation of current builds if new commits are pushed, and tighten permissions

rouault:


mwtoews (approved):
Note this PR should improve the "Token-Permissions" from https://securityscorecards.dev/viewer/?uri=github.com/libgeos/geos, which are currently zero.

PR #1044, opened 2024-02-06
CI: increase -j to take into account that recently GitHub action workers have bumped to 4 vCPUs

rouault:
Cf https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,7 +241,7 @@ jobs:
           -DUSE_CCACHE=ON \
           -DBUILD_DOCUMENTATION=YES \
           -DCMAKE_BUILD_TYPE=${{ matrix.ci.build_type }} ..
-        make -j 2
+        make -j $(nproc)
         cmake --build . --target docs
         ccache -s
 
@@ -322,7 +322,7 @@ jobs:
         -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
         -DUSE_CCACHE=ON \
         -G"MSYS Makefiles" ..
-       cmake --build . -j 2
+       cmake --build . -j 4
 
     - name: Save build cache
       uses: actions/cache/save@v3
@@ -370,7 +370,7 @@ jobs:
        cd build
        cmake --version
        cmake -DCMAKE_BUILD_TYPE=${{ matrix.ci.build_type }} -DCMAKE_CXX_STANDARD=${{ matrix.ci.cxxstd }} -DBUILD_SHARED_LIBS=ON -DUSE_CCACHE=ON ..
-       cmake --build . --config ${{ matrix.ci.build_type }} -j 2
+       cmake --build . --config ${{ matrix.ci.build_type }} -j 4
 
     - name: Save build cache
       uses: actions/cache/save@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -241,7 +241,7 @@ jobs:
           -DUSE_CCACHE=ON \
           -DBUILD_DOCUMENTATION=YES \
           -DCMAKE_BUILD_TYPE=${{ matrix.ci.build_type }} ..
-        make -j 2
+        make -j $(nproc)
```
mwtoews, line 244:
```suggestion
        cmake --build . --parallel
```

.github/workflows/ci.yml
```diff
@@ -322,7 +322,7 @@ jobs:
         -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
         -DUSE_CCACHE=ON \
         -G"MSYS Makefiles" ..
-       cmake --build . -j 2
+       cmake --build . -j 4
```
mwtoews, line 325:
```suggestion
       cmake --build . --parallel
```

.github/workflows/ci.yml
```diff
@@ -370,7 +370,7 @@ jobs:
        cd build
        cmake --version
        cmake -DCMAKE_BUILD_TYPE=${{ matrix.ci.build_type }} -DCMAKE_CXX_STANDARD=${{ matrix.ci.cxxstd }} -DBUILD_SHARED_LIBS=ON -DUSE_CCACHE=ON ..
-       cmake --build . --config ${{ matrix.ci.build_type }} -j 2
+       cmake --build . --config ${{ matrix.ci.build_type }} -j 4
```
mwtoews, line 373:
```suggestion
       cmake --build . --config ${{ matrix.ci.build_type }} --parallel
```

.github/workflows/ci.yml
```diff
@@ -439,7 +439,7 @@ jobs:
           -DCMAKE_CXX_STANDARD=${CXX_STANDARD} \
           -DUSE_CCACHE=ON \
           -DCMAKE_BUILD_TYPE=${BUILD_TYPE} ..
-        cmake --build . --config ${BUILD_TYPE} -j 3
+        cmake --build . --config ${BUILD_TYPE} -j 4
```
mwtoews, line 442:
```suggestion
        cmake --build . --config ${BUILD_TYPE} --parallel
```

.github/workflows/ci.yml
```diff
@@ -496,6 +496,6 @@ jobs:
         cp geos/examples/capi_read.c .
         cmake --version
         cmake -S . -B build
-        cmake --build build -j 2
+        cmake --build build -j $(nproc)
```
mwtoews, line 499:
```suggestion
        cmake --build build --parallel
```

mwtoews (commented):
The number of jobs argument is optional ([ref](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-j)), so perhaps these build commands could be simpler?

.github/workflows/ci.yml
rouault: I've some bad experience of "make -j" without values on directories with a very large number of .cpp files where it forks a gcc for each file, which exhausts the RAM and in the best case trigger OOM killer and in the worse case forces me to reboot...

mwtoews: Yup, on second thought I've had these issues too (and I normally hardcode this number too, leaving "1 cpu" remaining to "do other things"). In this case, you may reject my suggestions.

mwtoews (approved):

PR #1045, opened 2024-02-07
CI: add CodeQL static analyzer

rouault:
We've added this guy recently in GDAL, PROJ and MapServer. I've given it a try in my rouault/geos fork and it found zero issues... ( https://github.com/rouault/geos/security/code-scanning?query=is%3Aopen+branch%3Aci_codeql+ ). Which is both a bit surprising (I had to fix hundreds in GDAL), and expected (GEOS shouldn't do anything dangerous).

Not sure how much appetite there is to add this here. Adding it will increase the GEOS score on https://securityscorecards.dev/viewer/?uri=github.com/libgeos/geos (SAST category)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,68 @@
+name: "CodeQL"
+
+on:
+  push:
+  pull_request:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


pramsey:
I think I'm fine with this, particularly if we are already clean and it's not too chatty.

PR #1050, opened 2024-02-24
Fix overlay op regression with mixed GeometryCollection on 3.9 branch

### LINKED ISSUE #1023: Regression on 3.9.5: difference overlay op with mixed dimension collection is failing
In the shapely test suite, we discovered a new failure when updating our CI from 3.9.4 to the latest 3.9.x bugfix release 3.9.5 (https://github.com/shapely/shapely/pull/1942).

Reproducing it with `geosop`. Checking out 3.9.4 and building locally, and using the built geosop:

```
$ ./build/bin/geosop
geosop - GEOS 3.9.4
Executes GEOS geometry operations
...
$ ./build/bin/geosop difference -a "GEOMETRYCOLLECTION (POINT (51 -1), LINESTRING (52 -1, 49 2))" -b "POINT (2 3)" -f wkt
GEOMETRYCOLLECTION (POINT (51 -1), LINESTRING (52 -1, 49 2))
```

Then checking out 3.9.5, rebuilding, and rerunning the above:

```
$ ./build/bin/geosop difference -a "GEOMETRYCOLLECTION (POINT (51 -1), LINESTRING (52 -1, 49 2))" -b "POINT (2 3)" -f wkt
terminate called after throwing an instance of 'geos::util::IllegalArgumentException'
  what():  IllegalArgumentException: Overlay input is mixed-dimension
Aborted (core dumped)
```

From the 3.9 new file (https://github.com/libgeos/geos/blob/3.9/NEWS) it's not directly obvious to me which patch might have caused this.

I don't know if it's worth looking into this (it's for an old release, and for mixed dimension collections (which might not very well supported in general anyway?), but it _is_ a regression compared to 3.9.4), and it also only happens for just 3.9.5. None of the other new point releases we are testing (3.8.x to 3.12.x) shows the same failure.

mwtoews:
This fixes a regression in the 3.9 branch (only the 3.9.5 release), which permits overlay ops GEOSDifference with mixed-type GeometryCollection inputs. It more-or-less reverts ad62ddf5d97beecc193756c99f6c076963d36536 but limits the caught exception to `geos::util::IllegalArgumentException`.

The added CAPI test passes on newer branches too (with some mods), and I'll possibly add this test into other branches after this PR looks good.

Closes #1023

---

There's just one outstanding thing that I don't fully understand. By design, the low-level C++ level does not support overlay ops with mixed GeometryCollection types. But it doesn't appear to be consistent, e.g. (here "GC" is a mixed GeometryCollection and "PT" is a Point):

- GC.difference(PT) // raises IllegalArgumentException
- PT.difference(GC) // doesn't raise, result is PT

demonstrated with `tests/unit/operation/overlayng/OverlayNGMixedPointsTest.cpp`:
```c++
/**
* Mixed collections not supported at a low-level
*/
template<>
template<>
void object::test<12> ()
{
    std::string a = "GEOMETRYCOLLECTION (POINT (51 -1), LINESTRING (52 -1, 49 2))";
    std::string b = "POINT (2 3)";
    std::unique_ptr<Geometry> geom_a = r.read(a);
    std::unique_ptr<Geometry> geom_b = r.read(b);

    // A.difference(B) does not work
    try {
        std::unique_ptr<Geometry> geom_result = OverlayNG::overlay(geom_a.get(), geom_b.get(), OverlayNG::DIFFERENCE);
        fail();
    } catch(geos::util::IllegalArgumentException const& e) {
        ensure_equals(std::string(e.what()), "IllegalArgumentException: Overlay input is mixed-dimension");
    }
 
    // B.difference(A) works (?!)
    testOverlay(b, a, b, OverlayNG::DIFFERENCE, 1);
} 
```
is this expected, or unusual?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -17,6 +17,7 @@ xxxx-xx-xx
   - Remove undefined behaviour in use of null PrecisionModel (GH-931, Jeff Walton)
   - Skip over testing empty distances for mixed collections (GH-979, Paul Ramsey)
   - Add missing cstdint headers (GH-990, GH-743, Regina Obe, Sergei Trofimovich)
+  - HeuristicOverlay: Fix overlay op regression (GH-1050, Mike Taves)
 
 
 ## Changes in 3.9.4
--- a/src/geom/HeuristicOverlay.cpp
+++ b/src/geom/HeuristicOverlay.cpp
@@ -421,7 +421,8 @@ HeuristicOverlay(const Geometry* g0, const Geometry* g1, int opCode)
 
         return ret;
     }
-    catch(const geos::util::TopologyException& ex) {
+    catch(const geos::util::IllegalArgumentException& ex) {
+        // Mixed GeometryCollection; see GH-1050
         ::geos::ignore_unused_variable_warning(ex);
 
 #if GEOS_DEBUG_HEURISTICOVERLAY
--- a/tests/unit/capi/GEOSDifferenceTest.cpp
+++ b/tests/unit/capi/GEOSDifferenceTest.cpp
@@ -40,5 +40,33 @@ void object::test<1>()
     GEOSGeom_destroy(result);
 }
 
+/**
+* Mixed GeometryCollection types permitted at a high-level
+*/
+template<>
+template<>
+void object::test<2>()
+{
+    GEOSGeometry* a = GEOSGeomFromWKT("GEOMETRYCOLLECTION (POINT (51 -1), LINESTRING (52 -1, 49 2))");
+    GEOSGeometry* b = GEOSGeomFromWKT("POINT (2 3)");
+
+    ensure(a);
+    ensure(b);
+
+    GEOSGeometry* ab = GEOSDifference(a, b);
+    GEOSGeometry* ba = GEOSDifference(b, a);
+
+    ensure(ab);
+    ensure(ba);
+
+    ensure_geometry_equals(ab, a);
+    ensure_geometry_equals(ba, b);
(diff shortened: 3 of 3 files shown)


mwtoews:
Also curious to know about the broader implications such as https://trac.osgeo.org/postgis/ticket/5401

pramsey:
Sorry, broader implications? I mean, we could try to characterize how different versions deal with mixed inputs, ending up at the current state in which they are explicitly handled and passed through the NG engine entirely... at some level we want to just try and not knock over the Jenga tower, since retrofitting main/3.12 behaviour back onto stable branches would cause other consistency changes in output.

mwtoews:
Ok, perhaps we'll let the xref'ed broader implications slide for now, as we are essentially restoring the behaviour to the same as four releases (3.9.0 .. 3.9.4).

pramsey:
Good plan. I don't see a decent way to "fix" what was reported in [5401](https://trac.osgeo.org/postgis/ticket/5401) and their eventual move to a more modern release will clear things up for them.

mwtoews:
Backports of the test are pushed to be62defd9a1ffe81163bcb038e1edc1b93dd3be2 (3.10), dc4c486c169d5444fe49ec678a149659678f1371 (3.11), 22fe1a6cf245316c93eeec69836c3a5d378ce220 (3.12), and bdceea69174e80cf2cc2959d6da8d9dd0e766e3e (main). No other code changes are needed for these newer branches.

PR #1056, opened 2024-03-15
Fix buffer Inverted Ring Removal check

### LINKED ISSUE #984: Negative buffer operation returns EMPTY geometry
This bug was originally reported here: https://github.com/shapely/shapely/issues/1932

The input MultiPolygon is valid, and identical behaviour is obtained with a single Polygon too.

With a recent build of GEOS 3.13.0dev:
```bash
$ cat > mp.wkt
MULTIPOLYGON (((833454.7163917861 6312507.405413097, 833455.3726665961 6312510.208920742, 833456.301153878 6312514.207390314, 833492.2432584754 6312537.770332065, 833493.0901320165 6312536.098774815, 833502.6580673696 6312517.561360772, 833503.9404352929 6312515.0542803425, 833454.7163917861 6312507.405413097)))
$ ./bin/geosop -a mp.wkt buffer N-3.7
POLYGON ((833459.564533443 6312511.903163322, 833459.5698499765 6312511.926058625, 833490.8265097003 6312532.417314233, 833498.3074367424 6312517.923378301, 833459.564533443 6312511.903163322))
$ ./bin/geosop -a mp.wkt buffer N-3.8
POLYGON EMPTY
```
A buffer of -3.8 should be fine. Here is the visual result with -3.7:
image

This bug applies to JTS too (using a recent-ish 1.20.0 SNAPSHOT).

dr-jts:
This fixes the `BufferCurveSetBuilder.isRingCurveInverted` check to avoid removing a class of raw offset curves which are in fact valid buffer curves. 

The check was introduced in https://github.com/locationtech/jts/pull/706 and improved in https://github.com/locationtech/jts/pull/878 (ported as https://github.com/libgeos/geos/commit/85ead60acc547328217f41d1be3d721e24753300).  But the issues reported in #984 revealed that there were still valid cases being removed incorrectly.

Fixes #984

### DIFF
--- a/include/geos/geom/LineSegment.h
+++ b/include/geos/geom/LineSegment.h
@@ -250,16 +250,14 @@ public:
     //
     /// @param ret will be set to the midpoint of the segment
     ///
-    Coordinate midPoint() const
+    CoordinateXY midPoint() const
     {
-        return Coordinate(
-            (p0.x + p1.x) / 2,
-            (p0.y + p1.y) / 2);
+        return midPoint(p0, p1);
     };
 
-    static Coordinate midPoint(const Coordinate& pt0, const Coordinate& pt1)
+    static CoordinateXY midPoint(const CoordinateXY& pt0, const CoordinateXY& pt1)
     {
-        return Coordinate(
+        return CoordinateXY(
             (pt0.x + pt1.x) / 2,
             (pt0.y + pt1.y) / 2);        
     }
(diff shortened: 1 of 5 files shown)


include/geos/geom/LineSegment.h
```diff
@@ -250,13 +250,20 @@ class GEOS_DLL LineSegment {
     //
     /// @param ret will be set to the midpoint of the segment
     ///
-    void midPoint(Coordinate& ret) const
+    Coordinate midPoint() const
```
dbaston, line 253: Suggest using `CoordinateXY` anywhere we are not accessing `z`.

dr-jts: Good point.

PR #1057, opened 2024-03-19
CMakeLists.txt: add compatibility with latest Intel Compiler

rouault:
With those tunings (mostly forcing -fno-fast-math), build passes without warnings with ICC 2024.0.2.29 in Release mode, and tests pass as well.


rouault:
clang++-14 ASAN issue is unrelated since it also occurs on latest master: https://github.com/libgeos/geos/actions/runs/8335308670/job/22811441575

Issue #1066, opened 2024-04-15
Performance optimisation of point-to-point distance


martinfleis:
I've been wondering, if there is a space to optimise the distance measurement when all geometries are simple points. When using shapely and extracting coordinates to compute distance using numpy, I can get about 50x speedup.

See the example of measuring distance between 4k of points as posted in https://github.com/shapely/shapely/issues/2038.

```py
import shapely
import numpy as np

points = shapely.points(np.random.rand(4000, 2))

%%timeit
geos = shapely.distance(points, points.reshape(-1, 1))
# 4 s ± 21.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
# linalg.norm is convenient and commonly used for this
assert (shapely.get_type_id(points) == 0).all()
coords = shapely.get_coordinates(points)
linalg = np.linalg.norm(coords[:, None] - coords, axis=2)
# 328 ms ± 4.44 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
# but we can also be explicit, and faster
(shapely.get_type_id(points) == 0).all()
coords = shapely.get_coordinates(points)
sqrt = np.sqrt((coords[:, None][:, :, 0] - coords[:,0])**2 + (coords[:, None][:, :,1] - coords[:,1])**2)
# 71.2 ms ± 1.73 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

While we could do this in shapely, maybe there's a space to do it in GEOS?

kadyb:
In R-spatial, we also noted that some alternatives can be faster than GEOS.

```r
library("geos") # 3.11.1
library("Rfast")

n = 4000
df = data.frame(x = rnorm(n), y = rnorm(n))
pts = as_geos_geometry(sf::st_as_sf(df, coords = c("x", "y")))

geos_dist = function(x) {
  mat = matrix(0.0, nrow = length(x), ncol = length(x))
  for (i in seq_along(x)) mat[i, ] = geos::geos_distance(x[i], x)
  return(mat)
}

bench::mark(
  check = FALSE,
  geos = geos_dist(pts),
  R = as.matrix(dist(df, method = "euclidean")),
  Rfast = Rfast::Dist(df, method = "euclidean")
)
#>   expression      min   median `itr/sec` mem_alloc
#> 1 geos          6.87s    6.87s     0.146    1.01GB
#> 2 R          465.05ms 477.14ms     2.10   671.42MB
[...]

PR #1067, opened 2024-04-15
DistanceOp: Add short-circuit for point-point distance

dbaston:
Relates to https://github.com/libgeos/geos/issues/1066

Add a short circuit to reduce point-point computation time by ~90%.

### DIFF
--- a/benchmarks/operation/CMakeLists.txt
+++ b/benchmarks/operation/CMakeLists.txt
@@ -10,3 +10,13 @@
 ################################################################################
 add_subdirectory(buffer)
 add_subdirectory(predicate)
+
+if (benchmark_FOUND)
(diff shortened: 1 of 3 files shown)


dr-jts:
Good find, @dbaston .  Can you add the benchmark output to this PR for ease of seeing it?

dr-jts:
I guess there could be some other optimizations added as well, such as Point-Linestring, or perhaps LineString-LineString?   But maybe the returns diminish very quickly.

dbaston:
It looks like there is opportunity for more improvements to the general case. For example, in the point-point case we are spending ~50% of the total time in `PolygonExtracter` and `PointExtracter`:

image

Issue #1071, opened 2024-04-19
Segmentation fault in polygon.cpp


gabe-sherman:
A segmentation fault occurs in the below program. This behavior occurs at line 192 in polygon.cpp
```
#include <stdio.h>
#include <stdarg.h>
#include <string.h>
#include <stdlib.h>
#include <geos_c.h>

static void handler(const char *fmt, ...){
    exit(0);
}

int main(int argc, char *argv[])
{
    initGEOS(handler, handler);
    GEOSGeometry *v0 = GEOSGeomFromWKT("POLYGON((0 0, 00 0,   0 0))");
    GEOSGeometry *v1 = GEOSConcaveHullOfPolygons(v0, 0.7, NULL, NULL);
    return 0;
}
```

### Test Environment

Ubuntu 22.04, 64bit
### How to trigger

./filename
### Version

Latest: c8b889be9e8fa22de8a34bea50fec3bb073f6898

### Address Sanitizer Output
```
/home/gabesherman/harness_test/AutoHarn-Evaluation/geos/lib_asan/src/triangulate/polygon/PolygonHoleJoiner.cpp:139:46: runtime error: member call on null pointer of type 'geos::geom::Polygon'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/gabesherman/harness_test/AutoHarn-Evaluation/geos/lib_asan/src/triangulate/polygon/PolygonHoleJoiner.cpp:139:46 in 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==308818==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x555555aa1ba4 bp 0x7fffffffcfe0 sp 0x7fffffffce30 T0)
==308818==The signal is caused by a READ memory access.
==308818==Hint: address points to the zero page.
    #0 0x555555aa1ba4 in geos::geom::Polygon::getExteriorRing() const /home/gabesherman/harness_test/AutoHarn-Evaluation/geos/lib_asan/src/geom/Polygon.cpp:192:12
    #1 0x555555cbe649 in geos::triangulate::polygon::PolygonHoleJoiner::extractOrientedRings(geos::geom::Polygon const*) /home/gabesherman/harness_test/AutoHarn-Evaluation/geos/lib_asan/src/triangulate/polygon/PolygonHoleJoiner.cpp:139:46
[...]

PR #1076, opened 2024-04-20
GEOSConcaveHullOfPolygons: Avoid crash on zero-area input

dbaston:
Resolves https://github.com/libgeos/geos/issues/1071


dr-jts (commented):
It might be better to go further than this and return an empty hull for any input with zero area (not just a zero-area envelope).

src/algorithm/hull/ConcaveHullOfPolygons.cpp
```diff
@@ -153,7 +153,7 @@ ConcaveHullOfPolygons::setTight(bool p_isTight)
 std::unique_ptr<Geometry>
 ConcaveHullOfPolygons::getHull()
 {
-    if (inputPolygons->isEmpty()) {
+    if (inputPolygons->isEmpty() || inputPolygons->getEnvelopeInternal()->getArea() == 0) {
```
dbaston, line 156:
```suggestion
    if (inputPolygons->isEmpty() || inputPolygons->getArea() == 0) {
```

tests/unit/capi/GEOSConcaveHullOfPolygonsTest.cpp
```diff
@@ -62,6 +62,15 @@ void object::test<3>()
     result_ = GEOSConcaveHullOfPolygons(input_, 0.7, false, false);
     ensure("curved geometry not supported", result_ == nullptr);
 }
+  
+template<>
+template<>
+void object::test<3>()
```
dbaston, line 68:
```suggestion
void object::test<4>()
```

PR #1083, opened 2024-04-23
Add PointLocation.isOnSegment function

dr-jts:
Add `PointLocation.isOnSegment` function, and use it instead of `LineIntersector.hasIntersection`.

Also remove unused `LineIntersector.computeIntersection(p, p1, p2)`.

### DIFF
--- a/src/algorithm/PointLocation.cpp
+++ b/src/algorithm/PointLocation.cpp
@@ -20,15 +20,31 @@
 #include <vector>
 
 #include <geos/algorithm/LineIntersector.h>
+#include <geos/algorithm/Orientation.h>
 #include <geos/algorithm/PointLocation.h>
 #include <geos/algorithm/RayCrossingCounter.h>
 #include <geos/geom/CoordinateSequence.h>
 #include <geos/geom/Coordinate.h>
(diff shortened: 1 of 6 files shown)


src/algorithm/PointLocation.cpp
```diff
@@ -42,7 +58,7 @@ PointLocation::isOnLine(const geom::CoordinateXY& p, const geom::CoordinateSeque
     const geom::CoordinateXY* pp = &(pt->getAt<geom::CoordinateXY>(0));
```
pramsey, line 58: Maybe change this to a reference so you dispense with the `*` further down,

pramsey (approved):

src/algorithm/PointLocation.cpp
dr-jts: I tried that but was getting compile errors.  Does it need to be a pointer because it gets revalued at the end of the loop?

dr-jts: Another option is to simply define references for both points each time through the loop.

pramsey: Oh yeah, const references hate being redefined. 🤷  Nm.

Issue #1084, opened 2024-04-26
stack-buffer-overflow related to non-DE9IM argument for GEOSRelatePatternMatch_r


stevenagy:
The following program:
```
#include <geos_c.h>
int main(int argc, char *argv[])
{
      GEOSContextHandle_t handle = GEOS_init_r();
      GEOSRelatePatternMatch_r(handle, "0000000000",  "111111111");
      return 0;
}
```
Results in the following stack buffer overflow:

```=================================================================
==920298==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7fffffffdde4 at pc 0x555555a7f3ee bp 0x7fffffffdcb0 sp 0x7fffffffdca8
WRITE of size 4 at 0x7fffffffdde4 thread T0
    #0 0x555555a7f3ed in geos::geom::IntersectionMatrix::set(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /home/geos/lib_asan/src/geom/IntersectionMatrix.cpp:152:26
    #1 0x55555593506d in GEOSRelatePatternMatch_r::$_11::operator()() const /home/geos/lib_asan/capi/geos_ts_c.cpp:646:32
    #2 0x55555593506d in decltype(fp0) execute<GEOSRelatePatternMatch_r::$_11>(GEOSContextHandle_HS*, std::conditional<std::is_same<decltype(std::declval<GEOSRelatePatternMatch_r::$_11>()()), bool>::value, char, decltype(std::declval<GEOSRelatePatternMatch_r::$_11>()())>::type, GEOSRelatePatternMatch_r::$_11&&) /home/geos/lib_asan/capi/geos_ts_c.cpp:406:16
    #3 0x555555934dc1 in GEOSRelatePatternMatch_r 
    #4 0x555555930744 in main 
    #5 0x7ffff7029d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #6 0x7ffff7029e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #7 0x5555558704b4 in _start (BuildId: f2f0d6aa8fcc88fc49bab5d846274260243ce696)

Address 0x7fffffffdde4 is located in stack of thread T0 at offset 228 in frame
[...]

PR #1089, opened 2024-04-30
IntersectionMatrix: Ignore patterns longer than 9 characters

dbaston:
Fixes https://github.com/libgeos/geos/issues/1084

### DIFF
--- a/tests/unit/capi/GEOSRelatePatternTest.cpp
+++ b/tests/unit/capi/GEOSRelatePatternTest.cpp
@@ -33,7 +33,7 @@ void object::test<1>()
 
 template<>
 template<>
-void object::test<11>()
+void object::test<2>()
 {
     geom1_ = fromWKT("CIRCULARSTRING (0 0, 1 1, 2 0)");
     geom2_ = fromWKT("LINESTRING (1 0, 2 1)");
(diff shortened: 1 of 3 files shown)


tests/unit/capi/GEOSRelatePatternTest.cpp
```diff
@@ -45,5 +45,26 @@ void object::test<11>()
...
+    ensure(geom2_);
+
+    // pattern too long
+    ensure_equals(2, GEOSRelatePattern(geom1_, geom2_, "0FFFFF2120000000000000000000"));
+
+    // pattern too short
+    ensure_equals(2, GEOSRelatePattern(geom1_, geom2_, "0F"));
+
+    // pattern has invalid characters
+    ensure_equals(0, GEOSRelatePattern(geom1_, geom2_, "123456789"));
```
dr-jts, line 66: @dbaston: is it possible the args to `ensure_equals` are in the wrong order here? (as per https://github.com/libgeos/geos/blob/main/tests/unit/tut/tut_assert.hpp#L159)

dbaston: Yes, as with many of our tests

Issue #1064, opened 2024-04-07
RelateCompute intersection report inconsistent for lines and boundary points


strk:
RelateComputer says there's an intersection between line A interior and and line B boundary but says there is no intersection 
between B boundary and A.

Line A: 010200000002000000B02EA9C2AE5430400E3994256830514037F136E82055304039FAE86851305140
Line B: 01020000000200000064AB380D714E3040A38E9F0E1F305140E798A2D4E85430405F5969945C305140

### A Interior intersects B boundary in dim = 0
```
   I B E
I: F 0 1
B: F F 0
E: 1 0 2
```  

### A has no intersection with Boundary(B) 
```
   I B E
I: F F 1
B: F F 0
E: 0 F 2 
```

Reported in https://lists.osgeo.org/pipermail/geos-devel/2024-March/011012.html and confirmed in https://github.com/libgeos/geos/issues/968#issuecomment-2041134101

dr-jts:
This is not a bug, but a limitation in the ability to compute the relate matrix due to numeric rounding.  In order to correctly compute the relate matrix for two lines the node point must be computed (see https://github.com/locationtech/jts/issues/396 for an example of a case which requires noding for correct evaluation). However, due to numeric roundoff the computed node point equals an **endpoint** of line A, even though the robust orientation test reports that it lies in the **interior** of A.

It's not clear that there is an algorithm which can avoid noding if this would cause an inconsistency.

Another future option is to allow evaluating predicates with a distance tolerance.  This will essentially allow avoiding inconsistency due to roundoff.

strk:
For the record, GEOS-3.8 gives a different intersection matrix for the two lines, saying there is no interior/boundary intersection. Did 3.8 not have this limitation ?

```
   I B E
I: F F 1
B: F F 0
E: 1 0 2
```

strk:
An XML test for this case was added in #1092 - the test passes in 3.8 branch, fails in main branch:
```
build/bin/test_xmltester -v ../3.8/tests/xmltester/tests/issue/issue-geos-gh1064.xml 
../3.8/tests/xmltester/tests/issue/issue-geos-gh1064.xml: case1: test1: relatestring(A, B): failed. (0 ms)
        Description: See https://github.com/libgeos/geos/issues/1064
        Geometry A: LINESTRING (16.330791631988802 68.75635661578073, 16.332533372319826 68.75496886016562)
[...]

PR #1092, opened 2024-05-07
Gh1064 robust relate

strk:
See GH-1064


strk:
No CI for 3.8 branch here ? What's the rationale of that ?

strk:
CI runs but `make check` does not seem to be running any XML test

strk:
> CI runs but `make check` does not seem to be running any XML test

I was wrong, it's just harder to spot them, but they are there:
https://github.com/libgeos/geos/actions/runs/8982102259/job/24668979871?pr=1092#step:4:6239

strk:
Expected result should be updated as reported in https://github.com/libgeos/geos/pull/1093#issuecomment-2098495000

robe2:
@strk Does this need to be added to other branches.  GEOS 3.8 is EOL - https://libgeos.org/usage/download/

pramsey:
This PR should be closed as it adds a test that does not conform to what we think the answer should be, from discussion on GH-1064. Or it should be altered to have an answer that we agree with.

Issue #968, opened 2023-10-03
Potential bug in function covers


cuteDen-ECNU:
Consider the following code:
```python
from shapely.geometry import LineString, Point

a = LineString([(1, 0), (0, 2)])

b = Point(0.9, 0.2)

covers = a.covers(b)
# expected - {True}; actual - {False}

a1 = LineString([(10, 0), (0, 20)])

b1 = Point(9, 2)

covers = a1.covers(b1)
# True
```
Line a should cover point b because the distance between point b and line a is 0.
But Shapely doesn't consider line a covers point b.
The issue cannot be reproduced with integers only, indicating a potential floating-point issue. After all coordinate points are magnified ten times, the results are in line with expectations.

dr-jts:
This is indeed a bug.  Ironically it's due to the high-precision evaluation of the line-point orientation predicate. 

Not sure what a suitable fix is at the moment.  Likely it will involve some sort of precision rounding, but that needs to be done carefully to ensure there are no other impacts. 

See also https://trac.osgeo.org/postgis/ticket/5563

cuteDen-ECNU:
Thank you for your prompt response. It's greatly appreciated. 
Could you help me see whether [this potential bug](https://tarry-goal-613.notion.site/Potential-bug-in-the-ST_3DIntersects-function-d356d2cae4c448059ea1b2f88b24ca66?pvs=4) has the same core reason as it?
Once again, thank you for your expertise and assistance in addressing this bug. :·)

dr-jts:
> Could you help me see whether [this potential bug](https://trac.osgeo.org/postgis/ticket/5557) has the same core reason as it? 

The PostGIS `ST_3DIntersects` issue has the same basic cause: the inaccuracy of finite-precision floating point arithmetic.  But it's much less surprising in that case, since there's a lot of arithmetic being carried out, so the likelihood of 2 lines NOT intersecting in 3D is very high.

The bottom line is that you can't rely on topological relationships being preserved through floating-point operations.  This is the essence of the well-known issue of geometrical robustness failures.  The solution is usually either to use exact arithmetic (often impractical) or use tolerance values (not a panacea, but very often works).

cuteDen-ECNU:
Thank you for your prompt response. 
The explanation of geometrical robustness failures is very clear. 
The 'tolerance values' recommendation is very inspiring to me. 😊

dr-jts:
>  The 'tolerance values' recommendation is very inspiring to me. 😊

[...]

PR #1093, opened 2024-05-07
Add test for GH issue 1064, for main branch

strk:
See GH-1064

And see GH-1092 for the 3.8 runs

### DIFF
--- a/tests/xmltester/tests/issue/issue-geos-gh1064.xml
+++ b/tests/xmltester/tests/issue/issue-geos-gh1064.xml
@@ -1,0 +1,19 @@
+<run>
+  <precisionModel type="FLOATING" />
+
+<case>
+  <desc>See https://github.com/libgeos/geos/issues/1064</desc>
+  <a>
+    010200000002000000B02EA9C2AE5430400E3994256830514037F136E82055304039FAE86851305140
+  </a>
+  <b>
+    01020000000200000064AB380D714E3040A38E9F0E1F305140E798A2D4E85430405F5969945C305140
+  </b>
+<test>
+  <op name="relatestring" arg1="A" arg2="B">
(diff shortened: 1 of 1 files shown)


dr-jts:
AFAIK there is no known fix for this.  It's caused by rounding in the calculation of the intersection point (note, not the `intersects` test, which is accurate).  So this test can't be added to CI at this time, since it will just keep failing.

strk:
Rather than adding a broken expectance I'd expect the right answer, and I'd expect the test to fail.

If the picture you attached is correct the right answer should be: 0F1FF01F1, right ?

Maybe we can implement an "expected failure" that turns red in case the answer gets correct

dr-jts:
> Rather than adding a broken expectance I'd expect the right answer, and I'd expect the test to fail.
> If the picture you attached is correct the right answer should be: `0F1FF01F1`, right ?

What is the "right" answer is very subtle.  The best we can say is based on whatever numerical method is used to evaluate the Orientation predicate.  (It can also depend on the fundamental imprecise representation of decimal numbers - e.g. 0.9, as in #968.). 

For current versions of GEOS the answer is as given above.  Older versions use different arithmetic and get a different answer.  I agree there should be a "correct" answer - but I don't know how to calculate it (do you?).  And in any case, even if the computed answer is wrong, it doesn't help us to have CI failing all the time.

> Maybe we can implement an "expected failure" that turns red in case the answer gets correct

Yes, I've done this in JTS in a few places, mostly in order to capture cases which are known failures.  But someone needs to figure out what the correct answer is in this case.  At the moment all we have are two different answers, with no way to choose between them.

pramsey:
It sounds like this test is actually not appropriate for main branch, as the Truth has Changed.

PR #1095, opened 2024-05-14
GEOSClipByRect: preserve Z coordinate

vadimcn:
For [multi]linestrings/linerings, Z coordinate is interpolated from that of the nearest existing vertices.

For [multi]polygons, Z coordinate of points created on the boundary is interpolated as above.  For interior points, however, it is populated using the ElevationModel.

### DIFF
--- a/src/operation/intersection/RectangleIntersection.cpp
+++ b/src/operation/intersection/RectangleIntersection.cpp
@@ -17,6 +17,7 @@
 #include <geos/operation/intersection/RectangleIntersection.h>
 #include <geos/operation/intersection/Rectangle.h>
 #include <geos/operation/intersection/RectangleIntersectionBuilder.h>
+#include <geos/operation/overlayng/ElevationModel.h>
 #include <geos/operation/predicate/RectangleIntersects.h>
 #include <geos/geom/GeometryFactory.h>
 #include <geos/geom/CoordinateSequence.h>
@@ -34,6 +35,7 @@
 
 using geos::operation::intersection::Rectangle;
 using geos::operation::intersection::RectangleIntersectionBuilder;
+using geos::operation::overlayng::ElevationModel;
 using namespace geos::geom;
 using namespace geos::algorithm;
 namespace geos {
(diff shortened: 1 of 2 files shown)


src/operation/intersection/RectangleIntersection.cpp
```diff
@@ -682,7 +691,12 @@ std::unique_ptr<geom::Geometry>
 RectangleIntersection::clip(const geom::Geometry& g, const Rectangle& rect)
 {
     RectangleIntersection ri(g, rect);
-    return ri.clip();
+    std::unique_ptr<geom::Geometry> result = ri.clip();
+
+    std::unique_ptr<ElevationModel> elevModel = ElevationModel::create(g);
```
dbaston, line 696: This can be conditioned on the input actually having Z values.

vadimcn: done

vadimcn:
@dbaston does this look okay to merge now?

dbaston:
Do you think it makes sense to modify the tests to remove the expected value and just check that the result matches `GEOSIntersection` ?

vadimcn:
I don't know enough about the GEOS project to have a strong opinion on this.
Do you postulate that the result of GEOSClipByRect should always be the same as GEOSIntersection with a box polygon?  Or should it be the same after normalization, as you have above?  Then maybe yes.

However, currently, several existing tests fail this comparison, see below.  I'd say that fixing those is out of scope of this PR.

EXP: POINT EMPTY
OBT: GEOMETRYCOLLECTION EMPTY
[1=F]

EXP: POINT (15 10)
OBT: GEOMETRYCOLLECTION EMPTY
[3=F]

EXP: LINESTRING EMPTY
OBT: GEOMETRYCOLLECTION EMPTY
[4=F]

EXP: MULTILINESTRING ((10 10, 15 10), (10 10, 10 15))
OBT: GEOMETRYCOLLECTION EMPTY
[6=F]

EXP: MULTILINESTRING ((20 10, 20 20), (10 20, 20 20), (10 10, 20 10), (10 10, 10 20))
OBT: GEOMETRYCOLLECTION EMPTY
[10=F]

EXP: MULTILINESTRING ((20 10, 20 20), (10 20, 20 20), (10 10, 20 10), (10 10, 10 20))
OBT: GEOMETRYCOLLECTION EMPTY
[11=F]

dbaston:
> However, currently, several existing tests fail this comparison, see below.

Odd, they are passing for me after normalization. Either way, it can be beyond the scope of this PR.

vadimcn:
> Odd, they are passing for me after normalization. Either way, it can be beyond the scope of this PR.

I thought the plan was to update *all* tests, including those still using `checkClipByRect`. If you only changed `checkClipByRectIdentical`, I guess it'd work.

dbaston:
Looking more closely at the examples you pointed out, I can see that the semantics are different from `GEOSIntersection`. I guess the tests should be left as-is.

PR #1097, opened 2024-05-19
Add printf format attribute

strk:
Fixes warnings on downstream projects with -Wsuggest-attribute=format


mwtoews:
Re: MSVC errors, the Internet tells me about [Microsoft source-code annotation language (SAL)](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ms235402(v=vs.100)?redirectedfrom=MSDN), described in [this Q/A](https://stackoverflow.com/q/2354784/).

Some example implementations from [GDAL](https://github.com/OSGeo/gdal/blob/9c826123e4ff65d8d8e09b3cf84d0f3955e4fcf8/port/cpl_port.h#L955-L969) and [Folly](https://github.com/facebook/folly/blob/75e0eae5e820246d97be46c7838b58241ae531ff/folly/Portability.h#L61-L75). So (untested) something like:

```c
#ifdef _MSC_VER
#include <sal.h>
typedef void (*GEOSMessageHandler)(_Printf_format_string_(const char *fmt), ...);
#else
typedef void (*GEOSMessageHandler)(const char *fmt, ...) __attribute__ (( format(printf, 1, 0) ));
#endif
```
Should `-Wsuggest-attribute=format` be added to [CMakeLists.txt](https://github.com/libgeos/geos/blob/5bf67e6ff8599a7eef8ccbe8f02945a4a4e9dfa0/CMakeLists.txt#L311)?

strk:
@mwtoews good idea about adding the -WSuggest-attribute=format in CMakeLists.txt will see about doing that.

For the record, the downstream suggestion/failures can be seen here:
https://woodie.osgeo.org/repos/30/pipeline/2187/7#L907

strk:
For the record, PostGIS uses this:
```
#ifndef __GNUC__
# define __attribute__(x)
#endif
```

pramsey:
CI builds with warnings-as-errors, so now you need to actually follow up all the warnings :)

strk:
For now I've removed the -Wsuggest-attribute=format from the Makefile as I hadn't understood which line enables it only for the compiler supporting the __attribute__ syntax.

strk:
So, MingW suggests the gnu_printf  attribute but I don't know if it supports passing __attribute__

../capi/geos_ts_c.cpp:301:31: error: function 'void GEOSContextHandle_HS::NOTICE_MESSAGE(const char*, ...)' might be a candidate for 'gnu_printf' format attribute [-Werror=suggest-attribute=format]

strk:
The Windows / mingw-w64 keeps suggesting the attribute even if we're now defining it. I surrender. I don't even know why I'm wasting my time trying to make window builds happy

mwtoews:
@strk should I take a look at the msvc/mingw aspects? I can plop some commits here if needed.

strk:
> @strk should I take a look at the msvc/mingw aspects? I can plop some commits here if needed.

Yes please. If you can push directly go ahead, otherwise send a PR against this new branch.

mwtoews:
The mingw issue is a mystery (why it want's to add attributes to built-ins), and I'm able to replicate this locally. This oddity is side-stepped to ignore the warning within `__MINGW32__` defs.

The MSVC printf format stuff is handled differently, so it is probably easiest to put some portability macros into include/geos/export.h (I'd rather not create a separate port.h for this minor thing). MSVC's `sal.h`/`_Printf_format_string_` method only triggers warnings with the `-analize` flag, which I've tested locally to work. A later enhancement could enable this, but is out-of-scope for this PR.

Additional review / commits welcome!

strk:
include/geos/export.h is not included in geos_c.h so they should be replicated there for clients

mwtoews:
> include/geos/export.h is not included in geos_c.h

Looks fine to me: https://github.com/libgeos/geos/blob/effd6cbf0526d6479e195d9694883b5351dd78b1/capi/geos_c.h.in#L86

And I've just checked that these files are installed to the prefix, e.g. `/usr/local/include/geos/export.h`

pramsey:
This change actually *added* warnings for me... 
```
In file included from rt_band.c:37:
In file included from ./librtcore_internal.h:35:
In file included from ../../liblwgeom/lwgeom_geos.h:31:
/usr/local/include/geos_c.h:105:16: warning: redefinition of typedef 'GEOSMessageHandler' is a C11 feature [-Wtypedef-redefinition]
typedef void (*GEOSMessageHandler)(const char *fmt, ...);

../../liblwgeom/lwgeom_geos.h:28:16: note: previous definition is here
typedef void (*GEOSMessageHandler)(const char *fmt, ...) __attribute__ (( format(printf, 1, 0) ));
```

pramsey:
Considering changing the macro to this:
```
#if POSTGIS_GEOS_VERSION < 31300
#if __STDC_VERSION__ >= 201112L
/* See https://github.com/libgeos/geos/pull/1097 */
/* Only redefine the typedef if compiler is C11 or higher */
typedef void (*GEOSMessageHandler)(const char *fmt, ...) __attribute__ (( format(printf, 1, 0) ));
#endif
#endif
```

rouault:
and why not simply just not defining GEOSMessageHandler in lwgeom_geos.h, since it is available in geos_c.h ... ?

pramsey:
@strk can explain I guess he wants that extra attribute when he's compiling against older GEOS versions that don't have the attribute in the library include typedef.

strk:
> and why not simply just not defining GEOSMessageHandler in lwgeom_geos.h, since it is available in geos_c.h ... ?

Because until GEOS 3.13 the typedef of GEOSMessageHandler was not passing the printf __attribute__  thus resulting in a warning about "why not add an __attribute__ printf ?"

Where did you get 201112L from ?
I've committed that macro change in https://git.osgeo.org/gitea/postgis/postgis/commit/e3529a98caef2a0018fea994feff2c15e270ea1b

pramsey:
> Where did you get 201112L from ?

ChatGPT 🫢

pramsey:
Well, that doesn't work because in CI we have C declaring itself to be 1999 and yet still also complaining about the attribute warning. Why don't we just turn off that warning for all GEOS < 3.13? Fixing GEOS warnings in PostGIS by redefining GEOS types feels off to me.
```
lwgeom_geos_relatematch.c:56:23: error: argument 2 of 'initGEOS' might be a candidate for a format attribute [-Werror=suggest-attribute=format]
   56 |  initGEOS(lwpgnotice, lwgeom_geos_error);
      |                       ^~~~~~~~~~~~~~~~~
```

strk:
May I suggest to move this discussion to the PostGIS issue tracker instead ?

PR #1099, opened 2024-05-22
Add CoveragePolygonValidator section performance optimization

dr-jts:
Improves the performance of CoveragePolygonValidator and CoverageValidator by adding logic to avoid testing sections of ring vertices if the entire section does not overlap the target adjacent polygon.

Port of https://github.com/locationtech/jts/pull/1053

### World Polygons G99_5
Statistics: 323 geometries, 329,373 polygons, 459,540 vertices
Old code: 1579 ms
New code: 400 ms

### World Polygons G65
Statistics: 323 geometries, 319,646 polygons, 27,706,942 vertices
Old code: 5308 s
New code: 59 s

### DIFF
--- a/include/geos/coverage/CoveragePolygon.h
+++ b/include/geos/coverage/CoveragePolygon.h
@@ -25,7 +25,7 @@ class Polygon;
 }
 }
 
-using geos::geom::Coordinate;
+using geos::geom::CoordinateXY;
 using geos::geom::Envelope;
 using geos::geom::Polygon;
 using geos::algorithm::locate::IndexedPointInAreaLocator;
@@ -36,19 +36,19 @@ namespace coverage {    // geos::coverage
 class GEOS_DLL CoveragePolygon {
 
     // Members
-    const Polygon* polygon;
+    const Polygon* m_polygon;
     Envelope polyEnv;
-    std::unique_ptr<IndexedPointInAreaLocator> locator;
+    mutable std::unique_ptr<IndexedPointInAreaLocator> m_locator;
 
 public:
     CoveragePolygon(const Polygon* poly);
 
-    bool intersectsEnv(const Envelope env);
-    bool intersectsEnv(const Coordinate p);
-    bool contains(const Coordinate p);
+    bool intersectsEnv(const Envelope& env) const;
+    bool intersectsEnv(const CoordinateXY& p) const;
+    bool contains(const CoordinateXY& p) const;
     
 private:
-    IndexedPointInAreaLocator* getLocator();  
+    IndexedPointInAreaLocator& getLocator() const;  
 
 };
 
--- a/include/geos/coverage/CoveragePolygon.h
+++ b/include/geos/coverage/CoveragePolygon.h
@@ -1,0 +1,57 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2024 Martin Davis <mtnclimb@gmail.com>
+ *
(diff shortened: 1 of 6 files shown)


src/coverage/CoveragePolygon.cpp
```diff
@@ -0,0 +1,76 @@
...
+CoveragePolygon::contains(const Coordinate p)
+{
+    if (! intersectsEnv(p))
+        return false;
+    IndexedPointInAreaLocator* pia = getLocator();
+    return Location::INTERIOR == pia->locate(&p);
+}
+
+/* private */
+IndexedPointInAreaLocator*
```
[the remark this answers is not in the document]
dbaston: IMO you generally want to use references instead of pointers for arguments and return types unless you are returning a new object or there is a possibility that the value is null.

include/geos/coverage/CoveragePolygon.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace geos {        // geos
+namespace coverage {    // geos::coverage
+
+class GEOS_DLL CoveragePolygon {
+
+    // Members
+    const Polygon* polygon;
+    Envelope polyEnv;
+    std::unique_ptr<IndexedPointInAreaLocator> locator;
```
dbaston, line 41:
```suggestion
    mutable std::unique_ptr<IndexedPointInAreaLocator> locator;
```

dbaston: Marking this as `mutable` lets `getLocator` modify it while being `const`.

include/geos/coverage/CoveragePolygon.h
```diff
@@ -0,0 +1,57 @@
...
+
+public:
+    CoveragePolygon(const Polygon* poly);
+
+    bool intersectsEnv(const Envelope env);
+    bool intersectsEnv(const Coordinate p);
+    bool contains(const Coordinate p);
+    
+private:
+    IndexedPointInAreaLocator* getLocator();
```
dbaston, line 51:
```suggestion
    const IndexedPointInAreaLocator& getLocator() const;  
```

include/geos/coverage/CoveragePolygon.h
```diff
@@ -0,0 +1,57 @@
...
+using geos::geom::Polygon;
+using geos::algorithm::locate::IndexedPointInAreaLocator;
+
+namespace geos {        // geos
+namespace coverage {    // geos::coverage
+
+class GEOS_DLL CoveragePolygon {
+
+    // Members
+    const Polygon* polygon;
```
dbaston, line 39: Consider using the `m_` prefix for member variables to avoid shadowing warnings

src/coverage/CoveragePolygon.cpp
```diff
@@ -0,0 +1,76 @@
...
+using geos::geom::Polygon;
+
+namespace geos {     // geos
+namespace coverage { // geos.coverage
+
+/* public */
+CoveragePolygon::CoveragePolygon(const Polygon* poly)
+    : polygon(poly)
+{
+    polyEnv = *(poly->getEnvelopeInternal());
```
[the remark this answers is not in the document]
dr-jts: In JTS `getEnvelopeInternal` makes a defensive copy to ensure no aliasing errors.  This is a performance bottleneck. But I see GEOS doesn't do that, so maybe this is not needed.

src/coverage/CoveragePolygon.cpp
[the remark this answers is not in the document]
dr-jts: Makes sense.

PR #1110, opened 2024-06-13
Fix TopologyPreservingSimplifier to remove ring endpoints safely

### LINKED ISSUE #1107: Geometry produced by ST_SimplifyPreserveTopology is not stable across multiple runs
I have an issue with ST_SimplifyPreserveTopology with geos >= 3.12.1

I need ST_SimplifyPreserveTopology to be consistent across multiple runs, i.e. produce the same results when run with the the same inputs. In some cases this does not happen with geos >= 3.12.1

I have set up a test in this dbfiddle: https://dbfiddle.uk/GvwR39tT

The function test_simplify3 runs the same test for a certain number of iterations and returns the results as a table.

poly.the_geom is the test geometry, polygon, srid 3857, valid.
ST_SimplifyPreserveTopology(poly.the_geom, 20) is called in subqueries s1,s2. 

The function outputs these results about the simplified geometries:

- run: the iteration
- diff: difference
- wkt_difference: difference, as text
- simp_equals: result of st_equals
- simp_ordering_equals: result of st_orderingequals
- eq: result of bare equality
- wkb_eq: result of equality between wkbs
- valid1: whether first geometry is valid
- valid2: same as above for the second geometry
- n1simp: how many points in first geometry
- n2simp: same as above for second geometry

In the dbfiddle the function is called with iterations=500

`select * from test_simplify3(500) where st_isempty(diff) is distinct from true;`

I have tested the same on another server:

`POSTGIS="3.4.2 c19ce56" [EXTENSION] PGSQL="150" GEOS="3.12.1-CAPI-1.18.1" (compiled against GEOS 3.10.2) SFCGAL="SFCGAL 1.4.1, CGAL 5.3.1, BOOST 1.74.0" PROJ="8.2.1 NETWORK_ENABLED=OFF URL_ENDPOINT=https://cdn.proj.org USER_WRITABLE_DIRECTORY=/tmp/proj DATABASE_PATH=/usr/share/proj/proj.db" LIBXML="2.9.13" LIBJSON="0.15" LIBPROTOBUF="1.3.3" WAGYU="0.5.0 (Internal)" (core procs from "3.4.1 ca035b9" need upgrade) (sfcgal procs from "3.4.1 ca035b9" need upgrade)`

and I get failures even with less iterations (ex: 100)

image

As you can see there are cases where:

- st_isempty(diff) = f
- wkt_difference = 'POLYGON EMPTY'
- all other results seem to indicate the two geometries are the same: simp_equals=t, simp_ordering_equals=t, eq=t, wkb_eq=t

I am not able to reproduce the issue in windows with the latest available postgis bundle: 
PostgreSQL 15.7, postgis 3.4.2, geos 3.12.1
Tried with iterations=5000, (....it takes a while but) no errors are reported.

On the ubuntu server mentioned above, I have also made a test with the very latest geos release 3.12.2. Probably a coincidence, but the results are worst (~10% failure rate with 100 iterations).

As a side note, I need to use geos >= 3.12.1 because of other ST_SimplifyPreserveTopology instabilities, see this related issue:

https://trac.osgeo.org/postgis/ticket/5737

dr-jts:
Fixes `TopologyPreservingSimplifier` to remove ring endpoints safely (avoiding use-after-free errors).  It ensures that discarded line segments adjacent to ring endpoints are removed from spatial indexes, and modified line segments are reindexed.

Fixes #1107.


pramsey:
Is it possible to add the triggering test case into one of the unit tests, so we get that sweet ASAN/valgrind automation from the CI on that code path?

dr-jts:
> Is it possible to add the triggering test case into one of the unit tests?

Yes, would be nice.  Maybe can slim the failure case down a bit.

src/simplify/TaggedLineString.cpp
```diff
@@ -281,15 +281,16 @@ TaggedLineString::addToResult(std::unique_ptr<TaggedLineSegment> seg)
 #endif
 }
 
-void
+TaggedLineSegment*
```
dbaston, line 284:
```suggestion
const TaggedLineSegment*
```

src/simplify/TaggedLineStringSimplifier.cpp
```diff
@@ -187,7 +187,15 @@ TaggedLineStringSimplifier::simplifyRingEndpoint(double distanceTolerance)
...
             isTopologyValid(line, firstSeg, lastSeg, simpSeg))
         {
-            line->removeRingEndpoint();
+            //-- don't know if segments are original or new, so remove from all indexes
+            inputIndex->remove(firstSeg);
+            inputIndex->remove(lastSeg);
+            outputIndex->remove(firstSeg);
+            outputIndex->remove(lastSeg);
+
+            TaggedLineSegment* flatSeg = line->removeRingEndpoint();
```
dbaston, line 196:
```suggestion
            const TaggedLineSegment* flatSeg = line->removeRingEndpoint();
```

include/geos/simplify/TaggedLineString.h
```diff
@@ -105,7 +105,7 @@ class GEOS_DLL TaggedLineString {
 
     void addToResult(std::unique_ptr<TaggedLineSegment> seg);
 
-    void removeRingEndpoint();
+    TaggedLineSegment* removeRingEndpoint();
```
dbaston, line 108:
```suggestion
    const TaggedLineSegment* removeRingEndpoint();
```

dbaston: returning a `const` pointer here makes it clear that the caller does not own the returned object.

dbaston (changes requested):

dbaston (approved):

eglaser77:
Is this the issue referred to in the 3.12.0 release notes as "GEOSIntersects: Fix crash with empty point inputs (#1110, Dan Baston)" listed here: https://github.com/libgeos/geos/blob/3.12.2/NEWS.md#changes-in-3120? That description doesn't seem to match.

dbaston:
The issue referenced in the 3.12.0 release notes is Trac ticket 1110: https://trac.osgeo.org/geos/ticket/1110

PR #1111, opened 2024-06-14
Curved geometries: Implement length, area, point-in-polygon testing

dbaston:
This PR extends implements `getLength()` and `getArea()` methods for curved geometry types. It also extends `SimplePointInAreaLocator` and `RayCrossingCounter` to handle rings defined by a combination of line segments and circular arcs (as in `CompoundCurve` geometries).

The PIP functionality is not yet exposed in the Geometry API. This could either be done by adding a special case to `Geometry::intersects` or by allowing `PreparedGeometry` to be constructed from curved geometries. To do this, `IndexedPointInAreaLocator` would also need to support curved geometries. I'm not sure if this is actually useful for real datasets that use curved geometries, which I'd imagine to more commonly be smaller regions like tax parcels rather than the large polygons that drive the need for `IndexedPointInAreaLocator`.

The `RayCrossingCounter` work relies on computing the intersection points between arcs and a horizontal ray so it is not as robust as the linear version, but it should be an improvement over the alternative of linearizing the arcs.

### DIFF
--- a/include/geos/geom/CircularArc.h
+++ b/include/geos/geom/CircularArc.h
@@ -92,6 +92,11 @@ public:
             return 2*MATH_PI;
         }
 
+        /// Even Rouault:
+        /// potential optimization?: using crossproduct(p0 - center, p2 - center) = radius * radius * sin(angle)
+        /// could yield the result by just doing a single asin(), instead of 2 atan2()
+        /// actually one should also likely compute dotproduct(p0 - center, p2 - center) = radius * radius * cos(angle),
+        /// and thus angle = atan2(crossproduct(p0 - center, p2 - center) , dotproduct(p0 - center, p2 - center) )
         auto t0 = theta0();
         auto t2 = theta2();
 


include/geos/geom/CircularArc.h
```diff
@@ -0,0 +1,268 @@
...
+    double getLength() const {
+        if (isLinear()) {
+            return p0.distance(p2);
+        }
+
+        return getAngle()*getRadius();
+    }
+
+    /// Return the area enclosed by the arc p0-p1-p2 and the line segment p2-p0
+    double getArea() const {
```
rouault, line 121: perhaps more descriptive?
```suggestion
    double getCapArea() const {
```

tests/unit/algorithm/AreaTest.cpp
```diff
@@ -115,6 +120,41 @@ void object::test<3>
...
+{
+    checkAreaOfRing("COMPOUNDCURVE (CIRCULARSTRING (0 0, 2 2, 4 0), (4 0, 0 0))", 2*MATH_PI);
+}
+
+template<>
+template<>
+void object::test<6>
+()
+{
+    // expected ares from PostGIS after ST_CurveToLine(geom, 1e-13, 1)
```
rouault, line 144:
```suggestion
    // expected area from PostGIS after ST_CurveToLine(geom, 1e-13, 1)
```

tests/unit/algorithm/AreaTest.cpp
```diff
@@ -115,6 +120,41 @@ void object::test<3>
...
+    checkAreaOfRing("CIRCULARSTRING (0 0, 2 2, 4 0, 2 1, 0 0)", 3.48759);
+}
+
+template<>
+template<>
+void object::test<7>
+()
+{
+    // expected ares from PostGIS after ST_CurveToLine(geom, 1e-13, 1)
```
rouault, line 153:
```suggestion
    // expected area from PostGIS after ST_CurveToLine(geom, 1e-13, 1)
```

include/geos/geom/CircularArc.h
```diff
@@ -0,0 +1,268 @@
...
+        return std::isnan(getRadius());
+    }
+
+    /// Return the inner angle of the sector associated with this arc
+    double getAngle() const {
+        if (isCircle()) {
+            return 2*MATH_PI;
+        }
+
+        auto t0 = theta0();
```
rouault, line 95: potential optimization?: using crossproduct(p0 - center, p2 - center) = radius * radius * sin(angle)  could yield the result by just doing a single asin(), instead of 2 atan2()

rouault: actually one should also likely compute dotproduct(p0 - center, p2 - center) = radius * radius * cos(angle),  and thus angle = atan2(crossproduct(p0 - center, p2 - center) , dotproduct(p0 - center, p2 - center) )

include/geos/geom/CircularArc.h
dbaston: I see "cap" used for the spherical analog to this but "segment" for a circle. I think "segment" would be confusing in a library that is mostly dealing in line segments, though.

mwtoews:
@dbaston is this feature planned for [3.13.0](https://github.com/libgeos/geos/milestone/12)?

dbaston:
@mwtoews I will rebase it to fix the conflict and then merge it if there are no objections (@pramsey , @dr-jts ?)

include/geos/geom/CircularArc.h
dbaston: Added as comments

pramsey:
Is this targeting 3.13?

dbaston:
> Is this targeting 3.13?

Unless there's reason not to

Issue #1074, opened 2024-04-19
heap-buffer overflow in LineIntersector.h


gabe-sherman:
A heap-buffer overflow occurs in the below program when provided with malformed input. This behavior occurs at line 604 in LineIntersector.h
```
#include <stdio.h>
#include <stdarg.h>
#include <stdlib.h>
#include <geos_c.h>

static void geos_message_handler(const char* fmt, ...){
    va_list ap;
    va_start(ap, fmt);
    vprintf (fmt, ap);
    va_end(ap);
    exit(1);
}

int main(int argc, char *argv[])
{
    FILE *f;
    char *fuzzData = NULL;
    long size;

    if(argc < 2)
        exit(0);
    
    f = fopen(argv[1], "rb");
    if(f == NULL)
        exit(0);

    fseek(f, 0, SEEK_END);

    size = ftell(f);
    rewind(f);

    if(size < 1) 
        exit(0);

    fuzzData = (char*)malloc((size_t)size+1);
    if(fuzzData == NULL)
        exit(0);

    if(fread(fuzzData, (size_t)size, 1, f) != 1)
        exit(0);
    fuzzData[size] = '\0';
    initGEOS(geos_message_handler, geos_message_handler);
    double GEOSLargestEmptyCirclevar0 = 2.0;
    GEOSWKTReader* GEOSWKTReader_createval1 = GEOSWKTReader_create();
    GEOSGeometry* GEOSWKTReader_readval1 = GEOSWKTReader_read(GEOSWKTReader_createval1, "GEOMETRYCOLLECTION (MULTIPOINT((0 0), (1 1)), POINT(3 4), LINESTRING(2 3, 3 4))");
    GEOSGeometry* GEOSWKTReader_readval2 = GEOSWKTReader_read(GEOSWKTReader_createval1, fuzzData);
    GEOSGeometry* GEOSUnionval1 = GEOSUnion(GEOSWKTReader_readval2, GEOSWKTReader_readval1);
    GEOSGeometry* GEOSLargestEmptyCircleval1 = GEOSLargestEmptyCircle(GEOSWKTReader_readval2, GEOSWKTReader_readval2, GEOSLargestEmptyCirclevar0);
    GEOSGeometry* GEOSIntersectionval1 = GEOSIntersection(GEOSUnionval1, GEOSLargestEmptyCircleval1);
    GEOSWKTReader_destroy(GEOSWKTReader_createval1);
[...]

PR #1115, opened 2024-06-19
LineIntersector: Fixed invalid read on mixed-dimensionality inputs

dbaston:
Fixes https://github.com/libgeos/geos/issues/1074

### DIFF
--- a/include/geos/algorithm/LineIntersector.h
+++ b/include/geos/algorithm/LineIntersector.h
@@ -597,7 +597,7 @@ private:
 #endif
                 ptInt = static_cast<const C1&>(nearest);
             } else {
-                if (&ptInt == static_cast<const geom::CoordinateXY*>(&p1) || &ptInt == static_cast<const geom::CoordinateXY*>(&p2)) {
+                if (&nearest == static_cast<const geom::CoordinateXY*>(&p1) || &nearest == static_cast<const geom::CoordinateXY*>(&p2)) {
                     ptInt = static_cast<const C1&>(nearest);
                 } else {
                     ptInt = static_cast<const C2&>(nearest);
--- a/include/geos/algorithm/LineIntersector.h
+++ b/include/geos/algorithm/LineIntersector.h
@@ -589,8 +589,20 @@ private:
     {
         geom::CoordinateXYZM ptInt(Intersection::intersection(p1, p2, q1, q2));
         if (ptInt.isNull()) {
-            // FIXME need to cast to correct type in mixed-dimensionality case
-            ptInt = static_cast<const C1&>(nearestEndpoint(p1, p2, q1, q2));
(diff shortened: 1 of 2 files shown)


include/geos/algorithm/LineIntersector.h
```diff
@@ -589,8 +589,20 @@ class GEOS_DLL LineIntersector {
...
-            ptInt = static_cast<const C1&>(nearestEndpoint(p1, p2, q1, q2));
+            const geom::CoordinateXY& nearest = nearestEndpoint(p1, p2, q1, q2);
+#if __cplusplus >= 201703L
+            if constexpr (std::is_same<C1, C2>::value) {
+#else
+            if (std::is_same<C1, C2>::value) {
+#endif
+                ptInt = static_cast<const C1&>(nearest);
+            } else {
+                if (&ptInt == static_cast<const geom::CoordinateXY*>(&p1) || &ptInt == static_cast<const geom::CoordinateXY*>(&p2)) {
```
rouault, line 600: Probably due to my lack of familiarity with that code, but I can't make sense of those address comparisons.  ptInt is a new object created at line 590. How could its address be the same as p1 or p2 ?

dbaston: It couldn't! The wrong variable is being compared here.

dbaston, line 600:
```suggestion
                if (&nearest == static_cast<const geom::CoordinateXY*>(&p1) || &nearest == static_cast<const geom::CoordinateXY*>(&p2)) {
```

tests/unit/capi/GEOSIntersectionTest.cpp
```diff
@@ -157,5 +157,15 @@ void object::test<8>()
...
 }
 
+template<>
+template<>
+void object::test<9>()
+{
+    geom1_ = fromWKT("GEOMETRYCOLLECTION(POINT(0 0 59.083333333333336), POINT(1 1 59.083333333333336), LINESTRING(2 4 111 1, 2 223 22 2), POLYGON((42 -2.222222223222222e+155 111 1, 2 3 1 1, 2 4 111 NaN, 42 -2.222222223222222e+155 111 NaN)), POLYGON((2.2222222212222224e+149 4 111 1, 42 -2.222222222222222e+149 22 2, 2 4 111 1, 22 2222222222222223 22 NaN, 2.2222222212222224e+149 4 111 NaN)))");
+    geom2_ = fromWKT("LINESTRING (1.6409301755752343e+149 -5.298190649085575e+148, 1.6666666660752404e+149 -5.55555555396982e+148)");
+
+    result_ = GEOSIntersection(geom1_, geom2_);
```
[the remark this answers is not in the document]
dbaston: The purpose of the test is to verify that no crash occurs - I'm not sure the specific result is important. But I added asserts to verify that the result is non-null and has the correct dimensionality.

PR #1116, opened 2024-06-19
CI: Disable fail-fast

dbaston:


mwtoews:
For convenience of observation, warnings and errors from GHA remain. These are shown the slightly-challenging-to-scroll-down-to "Annotations" section of the summary e.g. [here](https://github.com/libgeos/geos/actions/runs/9587803255?pr=1116). These are repetitive, but the two key messages are:

- macOS clang (11.7, 14, ASAN, macos-11): This is a scheduled macOS-11 brownout. The macOS-11 environment is deprecated and will be removed on June 28th, 2024.
- Windows (Visual Studio) (Debug, 14, windows-2022): Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/cache/save@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

These can either be addressed here, or I could potentially follow-up in a few days in a separate PR. (Most of the `actions/cache` upgrades except for one were addressed with d0d3d5a0bf9c5a3f3bcb4da64ae53fef0f7d64e2)

dbaston:
Thanks for pointing these out. I was working on the MacOS issue which is resolved in #1117. I committed an update to the "save" action in 87e98b72ebfe3510a2c235ae8b23e64503f6adab 🤠.

Issue #1121, opened 2024-06-27
Build fails in C++23 language mode due to one forward declaration


BwL1289:
GEOS version: `3.12.2`

In C++23 language mode (in both Clang 18.1.8 and GCC 14.1), the build fails due to use of forward declarations in places where the complete type should be present. 

Currently, the following lines are the root cause:
1. https://github.com/libgeos/geos/blob/3.12.2/include/geos/operation/buffer/OffsetCurveSection.h#L26
2. [geos/include/geos/operation/buffer/OffsetCurveSection.h at 3.12.2 · libgeos/geos](https://github.com/libgeos/geos/blob/3.12.2/include/geos/operation/buffer/OffsetCurveSection.h#L67C29-L67C39)

The same code builds fine in C++20 language mode in both of the above compilers. In C++23 what's changed is that if `unique_ptr` destructor is called (even implicitly, as `std::move` does under the hood) the type should be complete. Forward declarations are still allowed with `unique_ptr` if destructor isn't called in scope.

For now, by explicitly including `<geos/geom/CoordinateSequence.h>` in this file we bypass the issue and GEOS builds and works fine.

pramsey:
Can you open up a PR with the requested change please? We are not moving to C++23 any time soon, but non-breaking changes to support it are fine.

BwL1289:
@pramsey see [#1122](https://github.com/libgeos/geos/pull/1122)

BwL1289:
@pramsey thank you!

Issue #1120, opened 2024-06-27
`include/geos/constants.h` - relies on transitive inclusion of `std::size_t` which breaks build in C++23 language mode


BwL1289:
[This header](https://github.com/libgeos/geos/blob/3.12.2/include/geos/constants.h) relies on `std::size_t` (see: https://en.cppreference.com/w/cpp/types/size_t), which isn't available in the C+23 standard. 

As a result, we're getting the error below on Clang 18 and GCC 14.1 in C++23 language mode.

**Possible solution**: import one of the headers guaranteed to provide `std::size_t`. 

Error:
```
1.606 -- The C compiler identification is Clang 18.1.8
1.671 -- The CXX compiler identification is Clang 18.1.8
1.729 -- Detecting C compiler ABI info
1.952 -- Detecting C compiler ABI info - done
1.960 -- Check for working C compiler: /opt/llvm_toolchain/bin/clang - skipped
1.960 -- Detecting C compile features
1.961 -- Detecting C compile features - done
2.010 -- Detecting CXX compiler ABI info
2.235 -- Detecting CXX compiler ABI info - done
2.243 -- Check for working CXX compiler: /opt/llvm_toolchain/bin/clang++ - skipped
2.243 -- Detecting CXX compile features
2.244 -- Detecting CXX compile features - done
2.246 -- GEOS: Version 3.12.2
2.246 -- GEOS: C API Version 1.18.2
2.246 -- GEOS: JTS port 1.18.0
2.246 -- GEOS: Build type: Release
2.247 -- GEOS: Run-time output: /tmp/<redacted>/staging/geos/source/geos-3.12.2/<redacted>_build/bin
2.247 -- GEOS: Archives output: /tmp/<redacted>/staging/geos/source/geos-3.12.2/<redacted>_build/lib
2.249 -- GEOS: Require C++23
2.250 -- Looking for pow in m
2.431 -- Looking for pow in m - found
2.452 -- GEOS: Build geosop ON
2.452 -- GEOS: install runtime path for util: $ORIGIN/../lib64
2.459 -- GEOS: Build astyle OFF
2.471 -- GEOS: Configured 'dist' target
2.471 -- GEOS: Configured 'distcheck' target
2.472 -- Configuring done (1.0s)
2.516 -- Generating done (0.0s)
2.581 -- Build files have been written to: /tmp/<redacted>/staging/geos/source/geos-3.12.2/<redacted>_build
[...]

PR #1122, opened 2024-07-01
fix building in C++23 language mode

gorloffslava:
This commit resolves issues with building GEOS in C++23 language mode (both -std=c++23 and -std=gnu++23 are supported). 

In particular, this resolves the following issues:
1. https://github.com/libgeos/geos/issues/1120
2. https://github.com/libgeos/geos/issues/1121


gorloffslava:
@pramsey thank you for merging!

Issue #1129, opened 2024-07-21
MSVC 32-bit test failure with LargestEmptyCircle


mwtoews:
With master and 3.12.2 the ctest suite has one failure with MSVC x86 (32-bit). Here is the relevant excerpt:
```
...
   24/466 Test  #24: unit-algorithm-construct-LargestEmptyCircle ................***Failed    0.10 sec
  ===============================
    GEOS Unit Test Suite
  ===============================
  
  geos::algorithm::construct::LargestEmptyCircle: ...............[21=F]
  
  ---> group: geos::algorithm::construct::LargestEmptyCircle, test: test<21>
       problem: assertion failed
       failed assertion: `x coordinate does not match: expected `-7.1150000000000006e+01` actual `-1.1111111100000000e+01` with precision `2.0000000000000000e+00``
  
  tests summary: failures:1 ok:15
...
```
This test is no issue for the x64 MSCV compiler, so it could be a possible 32-bit issue.

See test [here](https://github.com/libgeos/geos/blob/6201229a66ae3eef2b0c62cc39dbab48b2e3df7e/tests/unit/algorithm/construct/LargestEmptyCircleTest.cpp#L333-L343), with xref to https://trac.osgeo.org/postgis/ticket/5626 and #1003

mwtoews:
Some additional information. This error is a bit unusual, as it seems to happen with this configuration (adjust generator as needed):
```
cmake -G "Visual Studio NN YYYY" -A Win32 -T v142 -D CMAKE_BUILD_TYPE=Release ..
cmake --build . --config Release
ctest --output-on-failure -C Release -R LargestEmptyCircle
```
where only these configuration combinations raise the error:

- `-A Win32` for 32-bit x64 build
- `-T v142` selects the [MSVC_TOOLSET_VERSION](https://cmake.org/cmake/help/latest/variable/MSVC_TOOLSET_VERSION.html), and v142 is for "VS 2019 (16.0)" (this is independent of generator)
-  `-D CMAKE_BUILD_TYPE=Release`

Changing any of these, e.g. Debug build type and `--config`/`-C` options, will _not_ raise the exception. This bug would be very challenging to find and easily side-stepped by (e.g.) changing to a newer toolset (`-T v143` for VS 2022 (17.0)), so it's very low priority to resolve.

PR #1130, opened 2024-07-21
CI: re-add MSBuild as a separate job, use matrix for ninja builds

mwtoews:
A few weeks ago I made some changes to the MSVC CI tests, replacing MSBuild with Ninja (#1118). While this adds speed, it looses coverage, so it would be worth re-adding. Also, it seemed that Ninja+Debug did not allow ccache to work.

This PR also adds some matrix items for MSVC Ninja build, adding one for 32-bit MSVC (which has one test failure #1129), and also a range of different MSVC versions and CXX standards.

One big caveat about using ccache is that compiler warnings are sometimes not shown, since it skips normal compiler processing if a cache is available. So there is value in periodically clearing the cache to see full CI logs.

This PR does not fix the repeated warning "cl : Command line warning D9025 : overriding '/W3' with '/W4'", as this may require some CMake modifications, which I'm planning to address sometime.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -350,57 +350,88 @@ jobs:
       working-directory: ./build
       run: ctest --output-on-failure
 
-  windows-msvc:
-    name: 'Windows (Visual Studio)'
-    strategy:
-      fail-fast: false
-      matrix:
-        ci:
-          - build_type: Debug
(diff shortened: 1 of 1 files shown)


pramsey:
I'm glad you're here to tune these things ;0

PR #1133, opened 2024-07-25
Fix MaximumInscribedCircle::computeMaximumIterations for small values

### LINKED ISSUE #1129: MSVC 32-bit test failure with LargestEmptyCircle

mwtoews:
This fixes an issue with `MaximumInscribedCircle::computeMaximumIterations` that sometimes returns bogus values.

For example, with [this example](https://github.com/libgeos/geos/blob/6201229a66ae3eef2b0c62cc39dbab48b2e3df7e/tests/unit/algorithm/construct/LargestEmptyCircleTest.cpp#L333-L343) the number of iterations is needed by [`LargestEmptyCircle::compute()`](https://github.com/libgeos/geos/blob/4907d593aa6aa31c25dbe73933bf66995a04e498/src/algorithm/construct/LargestEmptyCircle.cpp#L235). The geometry calculates ncells=0.0903549, which would normally have a log value -2.40401. But the early cast to `size_t` doesn't correctly handle negative values meaning the maxIter is either 0 or 4294965296 (i.e. it's up to the compiler to decide). This fix will re-establish the maxIter to 4000 for small "ncells" values (less than 1). The bug with MSVC 32-bit is that it was zero, meaning no iteration.

<s>Also clean-up one MSVC warning.</s>

Closes #1129

### DIFF
--- a/src/algorithm/construct/MaximumInscribedCircle.cpp
+++ b/src/algorithm/construct/MaximumInscribedCircle.cpp
@@ -85,9 +85,9 @@ MaximumInscribedCircle::computeMaximumIterations(const Geometry* geom, double to
     double diam = geom->getEnvelopeInternal()->getDiameter();
     double ncells = diam / toleranceDist;
     //-- Using log of ncells allows control over number of iterations
-    std::size_t factor = (std::size_t) std::log(ncells);
+    int factor = (int) std::log(ncells);
     if (factor < 1) factor = 1;
-    return 2000 + 2000 * factor;
+    return (std::size_t) (2000 + 2000 * factor);
 }
 
 /* public */


mwtoews:
Potentially a similar fix would be good with JTS [here](https://github.com/locationtech/jts/blob/0d5eac992da509d052e3d23b522c375ba902cd86/modules/core/src/main/java/org/locationtech/jts/algorithm/construct/MaximumInscribedCircle.java#L109) too.

dr-jts:
Might be nice to keep the codebases more in synch - by using the signed int idea?

tests/unit/coverage/CoverageSimplifierTest.cpp
```diff
@@ -476,6 +477,7 @@ void object::test<30> ()
         CoverageSimplifier::simplify(input, 10);
     }
     catch (geos::util::IllegalArgumentException& iae) {
+        ::geos::ignore_unused_variable_warning(iae);
```
dbaston, line 480:
```suggestion
    catch (geos::util::IllegalArgumentException&) {
```

tests/unit/coverage/CoverageSimplifierTest.cpp
```diff
@@ -475,7 +475,7 @@ void object::test<30> ()
         std::vector<std::unique_ptr<Geometry>> result =
         CoverageSimplifier::simplify(input, 10);
     }
-    catch (geos::util::IllegalArgumentException& iae) {
+    catch (geos::util::IllegalArgumentException&) {
```
dr-jts, line 478: Any reason this change is in this PR?

mwtoews: It's only mildly related to MSVC where the primary fix is for, but it is otherwise unrelated. I'm fine to remove it.

dr-jts: Probably best to put in a separate PR, to avoid any confusion down the road.

mwtoews:
Cherry-picked to the `3.12` maintenance branch 8fa289cf0146780a0be01e97501589b5a29224c4
Is this NEWS-worthy for 3.12.3? It seems so trivial and specific to me to not.

dr-jts:
>  Is this NEWS-worthy for 3.12.3? It seems so trivial and specific to me to not.

Agreed, probably no need to add to NEWS.

PR #1137, opened 2024-08-14
Performance Tweaks for RelateNG

pramsey:
Now that RelateNG is in, there are sure to be a few places the port can be made faster, and also probably a handful of places the approach itself can be sanded smoother.

### DIFF
--- a/include/geos/operation/relateng/RelateGeometry.h
+++ b/include/geos/operation/relateng/RelateGeometry.h
@@ -148,15 +148,30 @@ public:
 
     static std::string name(bool isA);
 
-    const Geometry* getGeometry() const;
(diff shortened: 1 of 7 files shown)


include/geos/operation/relateng/RelateGeometry.h
```diff
@@ -97,6 +97,13 @@ class GEOS_DLL RelateGeometry {
 
     static bool isZeroLength(const LineString* line);
 
+    bool isZeroLengthLine(const Geometry* g) {
```
rouault, line 100:
```suggestion
    bool isZeroLengthLine(const Geometry* g) const {
```

PR #1138, opened 2024-08-15
Add -Wsuggest-override for gcc, and add missing override qualifiers

rouault:


dbaston:
Maybe need to enable only for gcc >= 9.2 ?

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78010#c11

rouault:
> Maybe need to enable only for gcc >= 9.2 ?

good idea, done

pramsey:
Merge, merge, merge.

PR #1140, opened 2024-08-15
include/geos/geomgraph: remove useless virtual qualifiers, add final and improve a bit const safety

rouault:
On top of PR #1138 

The removal of the useless virtual qualifiers is safe since the addition of -Wsuggest-override in a previous commit makes sure that all overloaded methods (of methods that ware marked as 'virtual') are marked as 'override', which in turns ensures that all base methods that are overriden are marked as 'virtual'.
(This doesn't change the status quo if a method was overloaded, but the base class lacked a 'virtual' and the overloaded method lacked a 'override'...)

Said otherwise, the dangerous situation would be a
```cpp
class A { virtual void foo(); };
class B: public A { virtual void foo(); };
```
where we would erroneously remove the virtual qualifier.

But given -Wsuggest-override, the above code would have been corrected to
```cpp
class A { virtual void foo(); };
class B: public A { virtual void foo() override; };
```

And thus if we erroneously remove the virtual qualifier from A::foo() the compiler will error out.

This results in a small reduction in the size of libgeos.so (5954240 to 5946336), suggesting that de-virtualizations are done.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,6 +313,10 @@ target_compile_options(geos_developer_cxx_flags
     $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:-Wno-unknown-warning-option>
     )
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 9.2)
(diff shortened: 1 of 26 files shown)


nyalldawson:
@rouault did you identify these manually? Or was there some tool you used to find them?

pramsey:
Sounds good to me, thanks for doing a comprehensive patch, now we just have to follow the pattern henchforth!

pramsey:
mergemergemerge

PR #936, opened 2023-07-03
handle MSVC flag warnings

jmckenna:
- increases `cmake_minimum_required` to version 3.15, as the related MSVC flag policy was added in 3.15
  - see https://cmake.org/cmake/help/latest/policy/CMP0092.html
- this solves 707 repeated warnings with MSVC of `Command line warning D9025 : overriding '/W3' with '/W4'` such as:
```
[ 16%] Building CXX object CMakeFiles/geos.dir/src/geomgraph/Depth.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
Depth.cpp
[ 16%] Building CXX object CMakeFiles/geos.dir/src/geomgraph/DirectedEdge.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
DirectedEdge.cpp
[ 16%] Building CXX object CMakeFiles/geos.dir/src/geomgraph/DirectedEdgeStar.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
DirectedEdgeStar.cpp
[ 16%] Building CXX object CMakeFiles/geos.dir/src/geomgraph/Edge.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
Edge.cpp
[ 16%] Building CXX object CMakeFiles/geos.dir/src/geomgraph/EdgeEnd.cpp.obj
cl : Command line warning D9025 : overriding '/W3' with '/W4'
EdgeEnd.cpp
```

PR #1141, opened 2024-08-15
Unit tests: Improve build time

dbaston:
- Remove unnecessary includes from tut.hpp
- Add tut.hpp to precompiled headers

Improves test build time from 4:30 to 3:20 on my system.


pramsey (approved):

dbaston:
I disabled precompiled headers, since they were causing problems outside of Linux/gcc that I don't have time to track down. Remaining failures are:

MSVC:
```
 cl : command line  error D8016: '/O1' and '/RTC1' command-line options are incompatible [D:\a\geos\geos\build\tests\unit\test_geos_unit.vcxproj]
```
https://github.com/libgeos/geos/actions/runs/10410141366/job/28831289025?pr=1141#step:3:587

MacOS clang14:
```
179: geos::geom::Envelope: .....[6=F]..............
179: 
179: ---> group: geos::geom::Envelope, test: test<6>
179:      problem: assertion failed
179:      failed assertion: `FE_INVALID raised`
```
https://github.com/libgeos/geos/actions/runs/10410141366/job/28831291511?pr=1141#step:8:2684

tests/unit/CMakeLists.txt
```diff
@@ -20,6 +22,11 @@ target_include_directories(test_geos_unit
   PRIVATE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}>)
 
+# Precompile TUT
+#if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.16")
+#    target_precompile_headers(test_geos_unit PRIVATE "${CMAKE_CURRENT_LIST_DIR}/tut/tut.hpp")
+#endif()
+
 file(GLOB_RECURSE _testfiles ${CMAKE_CURRENT_LIST_DIR}/**/*Test.cpp CONFIGURE_DEPEND)
```
mwtoews, line 30: On a related note I'm considering a PR to increase the minimum CMake version to 3.16 (same as PROJ and GDAL). Xref #936

dbaston:
I suspect the "real solution" to improving the build time is to use a unity build, but that doesn't play well with TUT's use of typedefs:

```
typedef test_group<test_sgpr_data> group;
typedef group::object object;

test_group<test_sgpr_data> test_sgpr_group("geos::precision::SimpleGeometryPrecisionReducer");

template<>
template<>
void object::test<1>()
```

We need `object` to be globally unique, and doing so manually would make writing these tests even more cumbersome than it already is.

PR #1142, opened 2024-08-15
Use TemplateSTRTree for pair finding in tree, per @dbaston

pramsey:


src/operation/relateng/EdgeSetIntersector.cpp
```diff
@@ -68,26 +69,19 @@ EdgeSetIntersector::process(EdgeSegmentIntersector& intersector)
 {
     EdgeSegmentOverlapAction overlapAction(intersector);
 
-    for (const MonotoneChain& queryChain : monoChains) {
+    // Replaces JTS implementation that manaually iterates on the
```
rouault, line 72:
```suggestion
    // Replaces JTS implementation that manually iterates on the
```

dbaston:
> This appears to have improved performance on my p-i-p test quite substantially (90s => 68s)

How does this compare to the old PreparedGeometry implementation?

pramsey:
Aha, oh wait actually I accidentally flipped my tree back to the old implementation which is why it's faster. Note that the old implementation is in place still, the new implementation only kicks in for predicates that were not supported before.

src/operation/relateng/EdgeSetIntersector.cpp
```diff
@@ -68,26 +69,19 @@ EdgeSetIntersector::process(EdgeSegmentIntersector& intersector)
...
-            /**
-             * following test makes sure we only compare each pair of chains once
-             * and that we don't compare a chain to itself
-             */
-            if (testChain->getId() <= queryChain.getId())
-                continue;
+        testChain->computeOverlaps(queryChain, &overlapAction);
+
+        if (intersector.isDone())
+            return;
```
dbaston, line 82: I think you want to return `false` here, from the docstring of `queryPairs`.

dbaston: See `MCIndexNoder::intersectChains` for an example

PR #1143, opened 2024-08-16
Establish minimum CMake version 3.15

mwtoews:
This PR does the following:

- Sets CMake 3.15 as the minimum version
- Cleans up some old logic used to support older version of CMake 
- Updates CI testing with a range of versions 3.15 to 3.29

Some rationale for upgrading CMake versions:

- MSVC with Ninja builds repeat "cl : Command line warning D9025 : overriding '/W3' with '/W4'", which makes seeing "real" warnings difficult. This is the same issue and fix described with #936. See [policy CMP0092](https://cmake.org/cmake/help/latest/policy/CMP0092.html), which was attempted with 1df959cc8a39807e63a09bff6953ef43d434ccbf, but it seems to only take effect with `cmake_minimum_required(VERSION 3.15)`, thus the message finally goes away!
- CMake 3.15 is over 5 years old, released [July 17, 2019](https://www.kitware.com/cmake-3-15-0-available-for-download/). It is also currently the same minimum version used for PROJ and GDAL.
- There are several new features upgrading from 3.13 to 3.15 ([e.g. here](https://cliutils.gitlab.io/modern-cmake/chapters/intro/newcmake.html#cmake-3-15-cli-upgrade)), but these features are not used in this PR.

Downside:

- This version of CMake is not available as a system package for Linux distros older than Ubuntu 18.04 or Debian 10 (Buster). But this is not a show-stopper on older system, since modern cmake versions can be installed via `pip install cmake` (using [PyPI](https://pypi.org/project/cmake/)).

If a compelling case should be made to keep CMake at 3.13, then alternatively:
```cmake
cmake_minimum_required(VERSION 3.13..3.15)
```
([docs](https://cmake.org/cmake/help/v3.13/command/cmake_minimum_required.html)) can be used to adjust policy settings for 3.15, where feasible.

---

I think this PR is fit to be merged for GEOS 3.13. CMake 3.13 has been the minimum for several releases (since 3.10). However, if deemed too late in the development cycle (or someone _really_ wants version 3.13 for both CMake and GEOS) then hold this merge until after the `3.13` branch is made. I won't mind.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           cxxstd: 14
           arch: 64
           packages: 'g++-7-multilib gcc-7-multilib'
-          cmake: 3.13.*
+          cmake: 3.17.*
           os: ubuntu-20.04
 
         - cxx_compiler: g++-7
@@ -63,7 +63,7 @@ jobs:
           cxxstd: 14
           arch: 32
           packages: 'g++-7-multilib gcc-7-multilib g++-multilib gcc-multilib'
-          cmake: 3.13.*
+          cmake: 3.18.*
           os: ubuntu-20.04
 
         - cxx_compiler: g++-8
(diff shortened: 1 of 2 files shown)


dbaston (approved):

jmckenna:
thanks @mwtoews !  Indeed I've been applying my [PR#996](https://github.com/libgeos/geos/pull/936) for a year, so I'm very happy that other packagers are hitting those warnings now, and my voice is heard (through yours ha!).  Thanks again.

PR #1162, opened 2024-09-14
Use error bound filter in isInCircle predicate instead of long double.

tinko92:
This PR is edited to separate out the change to the orientation filter, now in #1184 .

For the incircle predicate, this PR proposes to not generally use long double (which increases precision a little but is not robust as the code notes) but to use a similar error bound filter instead (bound computation based on https://doi.org/10.1007/s10543-023-00975-x , I don't know the source for alternative expression of the incircle determinant, I have first seen it in CGAL). The rational for that is higher performance (roughly 4-10% speedup on my machine for perf_voronoi) and that a more robust computation here is probably motivated by preventing infinite loops in Delaunay Triangulation due to false cases of INTERIOR, which should be impossible with this version (except possibly for inputs near the lower end of the representable magnitude of double that produce denormalized results/underflow).

PS: I also have an implementation of exact predicates that I could port to geos to be run after failures of the fast filters if it would be considered beneficial to the library by its maintainers. I don't know enough about its use cases to make a guess about that. Exact predicates guarantee consistency wrt to predicate calls. They would not guarantee consistency between predicates and computed geometries (e.g. predicate might say some polygons overlap but the computed intersection may then be empty after rounding to double coordinates in the output geometry, even if all intermediate computations were exact).

### DIFF
--- a/src/triangulate/quadedge/TrianglePredicate.cpp
+++ b/src/triangulate/quadedge/TrianglePredicate.cpp
@@ -29,16 +29,33 @@ namespace quadedge {
 
 geom::Location
 TrianglePredicate::isInCircleNonRobust(
-    const CoordinateXY& a, const CoordinateXY& b, const CoordinateXY& c,
-    const CoordinateXY& p)
+    const CoordinateXY& p, const CoordinateXY& q, const CoordinateXY& r,
+    const CoordinateXY& t)
 {
-    auto det =
-        (a.x * a.x + a.y * a.y) * triArea(b, c, p)
-        - (b.x * b.x + b.y * b.y) * triArea(a, c, p)
-        + (c.x * c.x + c.y * c.y) * triArea(a, b, p)
-        - (p.x * p.x + p.y * p.y) * triArea(a, b, c);
-
-    return det > 0 ? geom::Location::EXTERIOR : (det < 0 ? geom::Location::INTERIOR : geom::Location::BOUNDARY);
+    // The following expression is based on a simplification of the more
+    // well-known det(ax-cx, ay-cy, (ax-cx)^2 + (ay-cy)^2, ...) found in
+    // https://github.com/CGAL/cgal/blob/e871025f364360a15671f6e825127df6207afa16/Cartesian_kernel/include/CGAL/predicates/kernel_ftC2.h#L495
+    auto qpx = q.x - p.x;
+    auto qpy = q.y - p.y;
+    auto rpx = r.x - p.x;
+    auto rpy = r.y - p.y;
+    auto tpx = t.x - p.x;
+    auto tpy = t.y - p.y;
+    auto tqx = t.x - q.x;
+    auto tqy = t.y - q.y;
+    auto rqx = r.x - q.x;
+    auto rqy = r.y - q.y;
+    auto qpxtpy = qpx * tpy;
+    auto qpytpx = qpy * tpx;
+    auto tpxtqx = tpx * tqx;
+    auto tpytqy = tpy * tqy;
+    auto qpxrpy = qpx * rpy;
+    auto qpyrpx = qpy * rpx;
+    auto rpxrqx = rpx * rqx;
+    auto rpyrqy = rpy * rqy;
+    auto det = (qpxtpy - qpytpx) * (rpxrqx + rpyrqy)
+               - (qpxrpy - qpyrpx) * (tpxtqx + tpytqy);
+    return static_cast<geom::Location>( (det > 0) - (det < 0) + 1 );
 }
 
 geom::Location
@@ -91,11 +108,45 @@ TrianglePredicate::triArea(const CoordinateXY& a,
 
 geom::Location
 TrianglePredicate::isInCircleRobust(
-    const CoordinateXY& a, const CoordinateXY& b, const CoordinateXY& c,
-    const CoordinateXY& p)
+    const CoordinateXY& q, const CoordinateXY& p, const CoordinateXY& r,
+    const CoordinateXY& t)
 {
-    // This implementation is not robust, name is ported from JTS.
-    return isInCircleNormalized(a, b, c, p);
+    // This implementation is not robust and defaults to BOUNDARY in case of
+    // uncertainty arising from floating-point round-off errors or overflow.
+    // There merits of this approach are discussed in
+    // https://github.com/libgeos/geos/pull/1162
+    // The source for the below expression is given in isInCircleNonRobust(...)
+    auto qpx = q.x - p.x;
+    auto qpy = q.y - p.y;
+    auto rpx = r.x - p.x;
+    auto rpy = r.y - p.y;
+    auto tpx = t.x - p.x;
+    auto tpy = t.y - p.y;
+    auto tqx = t.x - q.x;
+    auto tqy = t.y - q.y;
+    auto rqx = r.x - q.x;
+    auto rqy = r.y - q.y;
+    auto qpxtpy = qpx * tpy;
+    auto qpytpx = qpy * tpx;
+    auto tpxtqx = tpx * tqx;
+    auto tpytqy = tpy * tqy;
+    auto qpxrpy = qpx * rpy;
+    auto qpyrpx = qpy * rpx;
+    auto rpxrqx = rpx * rqx;
+    auto rpyrqy = rpy * rqy;
+    auto det = (qpxtpy - qpytpx) * (rpxrqx + rpyrqy)
+               - (qpxrpy - qpyrpx) * (tpxtqx + tpytqy);
+    // The following error bound computation is based on the publication
+    // https://doi.org/10.1007/s10543-023-00975-x
+    // It can be replicated by modifying the example given in the accompanying
+    // repository's https://zenodo.org/records/7539355 incircle.ipynb to render
+    // the error bound for the above incircle expression.
+    auto deterror = ((std::abs(qpxtpy) + std::abs(qpytpx))
+                    * (std::abs(rpxrqx) + std::abs(rpyrqy))
+                    + (std::abs(qpxrpy) + std::abs(qpyrpx))
+                    * (std::abs(tpxtqx) + std::abs(tpytqy)))
+                    * 9.99200719823023e-16;
+    return static_cast<geom::Location>( (det > deterror) - (det < -deterror) + 1 );
 }
 
 } // namespace geos.triangulate.quadedge


pramsey:
I don't have a voronoi test yet, but my [performance suite](https://github.com/pramsey/geos-performance) shows no differences, with one exception: the prepared point-in-poly test seems to be consistently about 10% faster.

tinko92:
@pramsey Thanks for the feedback. Interesting, point-in-poly may benefit from a faster orientation but I would have expected that to be negligible if its implemented with some ray counting then I'd expect runtime to be dominated by walking over segments that don't intersect the ray and need no orientation check (at least for polygons with many segments). I will try to look into that when I find time.

I would expect the change in the orientation test to be measurable in e.g. convex hull or triangulation without Delaunay requirement and the change in the incircle test to be measurable in benchmarks like Delaunay Triangulation or Voronoi diagram. The incircle test maybe very significantly on non-x86-architectures where no machine hardware accelerated 80-bit floats are available and long double would compile to function calls to some soft float implementation, e.g. https://godbolt.org/z/EosY5ehjq .

pramsey:
I hooked a Delaunay build into my test harness but couldn't measure any different between this PR and 3.13.

tinko92:
> I hooked a Delaunay build into my test harness but couldn't measure any different between this PR and 3.13.

I tested as follows using the Delaunay/Voronoi benchmarks in the benchmark folder from the main repo:
I checked out the main branch, went into the directory and:

```
mkdir build_gcc && build_gcc
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_BENCHMARKS=ON -DCMAKE_CXX_FLAGS="-march=native" .. 
make #will complain about missing override in another benchmark but doesn't affect the build of perf_voronoi
echo 1 | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo #only on intel
cpupower frequency-set -g performance #only on intel 
for i in {1..10}; do taskset -c 2 ./bin/perf_voronoi >> baseline; done;
cat baseline | grep "BM_DelaunayFromSeq/1000000"
BM_DelaunayFromSeq/1000000        5050644908 ns   5045715175 ns            1
BM_DelaunayFromSeq/1000000        5049845175 ns   5045245900 ns            1
BM_DelaunayFromSeq/1000000        6646025308 ns   6638513791 ns            1
BM_DelaunayFromSeq/1000000        6649783153 ns   6642201348 ns            1
BM_DelaunayFromSeq/1000000        6648571528 ns   6641021393 ns            1
BM_DelaunayFromSeq/1000000        5053053815 ns   5048554170 ns            1
BM_DelaunayFromSeq/1000000        5049133353 ns   5044510485 ns            1
BM_DelaunayFromSeq/1000000        5052486875 ns   5047828776 ns            1
BM_DelaunayFromSeq/1000000        5050604032 ns   5046051328 ns            1
BM_DelaunayFromSeq/1000000        5051929265 ns   5047293562 ns            1

#checkout the branch with this patch, same steps, then
for i in {1..10}; do taskset -c 2 ./bin/perf_voronoi >> patched; done;
cat patched | grep "BM_DelaunayFromSeq/1000000"
BM_DelaunayFromSeq/1000000        4659588868 ns   4654149493 ns            1
BM_DelaunayFromSeq/1000000        4654536370 ns   4649391288 ns            1
BM_DelaunayFromSeq/1000000        4653135324 ns   4648135157 ns            1
BM_DelaunayFromSeq/1000000        6127312010 ns   6119523080 ns            1
BM_DelaunayFromSeq/1000000        6125711504 ns   6118105844 ns            1
BM_DelaunayFromSeq/1000000        4655963070 ns   4651342713 ns            1
BM_DelaunayFromSeq/1000000        4652838792 ns   4648184104 ns            1
BM_DelaunayFromSeq/1000000        4654345257 ns   4649771922 ns            1
BM_DelaunayFromSeq/1000000        4666023790 ns   4661365412 ns            1
BM_DelaunayFromSeq/1000000        4653681302 ns   4649277328 ns            1
```

I honestly wish I knew, where the outliers come from with turbo disabled, maybe the P-core/E-core thing can affect it despite fixing the CPU core. Same steps with CXX=/usr/bin/clang++ before the cmake command yield for me:

```
cat baseline | grep "BM_DelaunayFromSeq/1000000"
BM_DelaunayFromSeq/1000000        4329365684 ns   4323526473 ns            1
BM_DelaunayFromSeq/1000000        5433784294 ns   5427474646 ns            1
BM_DelaunayFromSeq/1000000        4148863139 ns   4144892923 ns            1
BM_DelaunayFromSeq/1000000        4139543832 ns   4135714625 ns            1
BM_DelaunayFromSeq/1000000        5428346793 ns   5422042258 ns            1
BM_DelaunayFromSeq/1000000        5439176317 ns   5432822740 ns            1
BM_DelaunayFromSeq/1000000        5432680958 ns   5426345764 ns            1
BM_DelaunayFromSeq/1000000        4140026534 ns   4136116714 ns            1
BM_DelaunayFromSeq/1000000        4149153847 ns   4145284595 ns            1
BM_DelaunayFromSeq/1000000        5431052965 ns   5424639386 ns            1

cat patched | grep "BM_DelaunayFromSeq/1000000" 
BM_DelaunayFromSeq/1000000        3767896479 ns   3764021258 ns            1
BM_DelaunayFromSeq/1000000        4923848243 ns   4917679386 ns            1
BM_DelaunayFromSeq/1000000        3764235827 ns   3760602930 ns            1
BM_DelaunayFromSeq/1000000        3762541532 ns   3758951324 ns            1
BM_DelaunayFromSeq/1000000        3765034916 ns   3761276158 ns            1
BM_DelaunayFromSeq/1000000        3764714564 ns   3761178183 ns            1
BM_DelaunayFromSeq/1000000        4928664355 ns   4922710265 ns            1
BM_DelaunayFromSeq/1000000        4922105205 ns   4916260250 ns            1
BM_DelaunayFromSeq/1000000        3767456227 ns   3763747766 ns            1
BM_DelaunayFromSeq/1000000        3760288885 ns   3756814473 ns            1
```

Similar results for the other workloads.

And on an EPYC-Milan I don't have these outlier timings, so I post just the first sample for each workload:
```
cat baseline | grep 100000 | head -n5
BM_DelaunayFromSeq/1000000        4916505592 ns   4916348118 ns            1
BM_DelaunayFromGeom/1000000       4671339867 ns   4671204309 ns            1
BM_VoronoiFromSeq/1000000         5506559255 ns   5506240288 ns            1
BM_VoronoiFromGeom/1000000        5435366604 ns   5435120888 ns            1
BM_OrderedVoronoiFromGeom/1000000 7680563331 ns   7680220398 ns            1

cat patched | grep 100000 | head -n5 
BM_DelaunayFromSeq/1000000        4427675159 ns   4427496208 ns            1
BM_DelaunayFromGeom/1000000       4462223273 ns   4461665059 ns            1
BM_VoronoiFromSeq/1000000         5045797628 ns   5045404499 ns            1
BM_VoronoiFromGeom/1000000        4988454644 ns   4987171839 ns            1
BM_OrderedVoronoiFromGeom/1000000 7392607964 ns   7392126698 ns            1
```

pramsey:
Thanks for the nice writeup of your perf_* results. I did the same tests on my MacBook Pro and I'm afraid I see no consistent advantage for this particular optimization.

pramsey:
Would it make sense to gate this behind something like 
```
#if defined(__x86_64__) || defined(__x86_64) || defined(__amd64__) || defined(__amd64) || defined(_M_X64) || defined(_M_AMD64)
#   define ARCHITECTURE_X86_64
```

tinko92:
@pramsey Thanks for getting back to me on this.

The test results are interesting. I do not have a mac to test. But using clangs target option, I find that it seems to default to compile long double to 8 byte, so same as double:

```sh
echo "int s = sizeof(long double);" | clang++ -x c++ -S - -O3 -target aarch64-apple-darwin -o - | grep '.long'
	.long	8                               ; 0x8
```

So my patch would not benefit the circle predicate performance there from switching down to double. Instead it will cost a little performance due to the slightly more expensive error bound computation (visible in your benchmark, seems to be ~1-2%).

On the other hand, for other architectures that are compiling the current code to 8 byte precision also means that the current code will be a lot less robust. If I generate 100k degenerate cases, the current code will produce the an opposing result (interior when the truth is exterior or the other way around) in 8 cases with 80-bit precision but in ~14k cases with 64-bit precision. All code for the test is in this gist: https://gist.github.com/tinko92/86126244ed04b65aeb09c9223ec4c37c#file-test-sh (note that this test code will exclusively produce degenerate cases where the test point is near the circle boundary up to roughly machine precision). The patched code will never confuse interior and exterior regardless of availability of 80-bit precision doubles but always produce the correct sign or BOUNDARY. If you run that gist code on your macbook, I would expect that you see many results in which the 2nd and 3rd column show opposite signs.

So overall my opinion would be to not gate it. The architectures that lack 80-bit long doubles suffer a 2% performance hit but will benefit from it a lot robustness-wise (with an understanding of robustness where "boundary" is a safe result while confusing exterior and interior is bad) and the architectures that have 80-bit long doubles should benefit from it somewhat performance-wise and still a little robustness-wise (guarantee of not confusing exterior and interior).

Regarding the orientation predicate, I don't know any benchmark in which it matters (except the orientation micro benchmark), but this should be slightly faster on almost any architecture, except an where fabs cannot be compiled to an and-instruction masking the sign-bit (I don't know of any architecture where that is the case).

dr-jts:
It would be better if this was split into 2 PRs, one for each algorithm.  This would separate the different discussions, tests, and changes.

tinko92:
That makes sense, since most of the discussion is around the incircle predicate, I'll split out the orientation predicate into a separate PR.

tinko92:
In the mean time, I acquired a notebook with a Snapdragon X Elite CPU whose variants run the new Windows for ARM laptops. On WSL for this architecture (aarch64), long double is compiled to a 128-bit type by default

```sh
echo "int s = sizeof(long double);" | g++ -x c++ -S - -O3 -march=native -o - | grep '.word'
        .word   16
```

and clearly soft-float:

```asm
echo "long double f(long double a, long double b) { return a + b; }" | clang++ -x c++ -S - -O3 -march=native -o -
//...
        .cfi_startproc
// %bb.0:
        b       __addtf3
.Lfunc_end0:
//...
```

(__addtf3 being the name of an intrinsic add function for floating-point emulation for long double).

For this platform the performance differences are therefore very significant:

```sh
grep 1000000 baseline | head -n10
BM_DelaunayFromSeq/1000000        1.5618e+10 ns   1.5570e+10 ns            1
BM_DelaunayFromGeom/1000000       1.5766e+10 ns   1.5699e+10 ns            1
BM_VoronoiFromSeq/1000000         1.1918e+10 ns   1.1901e+10 ns            1
BM_VoronoiFromGeom/1000000        1.1941e+10 ns   1.1923e+10 ns            1
BM_OrderedVoronoiFromGeom/1000000 1.3377e+10 ns   1.3353e+10 ns            1
BM_DelaunayFromSeq/1000000        1.5252e+10 ns   1.5206e+10 ns            1
BM_DelaunayFromGeom/1000000       1.4999e+10 ns   1.4927e+10 ns            1
BM_VoronoiFromSeq/1000000         1.1669e+10 ns   1.1649e+10 ns            1
BM_VoronoiFromGeom/1000000        1.1558e+10 ns   1.1542e+10 ns            1
BM_OrderedVoronoiFromGeom/1000000 1.2732e+10 ns   1.2714e+10 ns            1
```

```sh
grep 1000000 patched | head -n10
BM_DelaunayFromSeq/1000000        3041869000 ns   3030840500 ns            1
BM_DelaunayFromGeom/1000000       2619609077 ns   2616074300 ns            1
BM_VoronoiFromSeq/1000000         3260589305 ns   3254602500 ns            1
BM_VoronoiFromGeom/1000000        3040886840 ns   3035463600 ns            1
BM_OrderedVoronoiFromGeom/1000000 4423922233 ns   4415210600 ns            1
BM_DelaunayFromSeq/1000000        3048331830 ns   3016170200 ns            1
BM_DelaunayFromGeom/1000000       2988821373 ns   2931008200 ns            1
BM_VoronoiFromSeq/1000000         3242640471 ns   3236409000 ns            1
BM_VoronoiFromGeom/1000000        3035475650 ns   3030037400 ns            1
BM_OrderedVoronoiFromGeom/1000000 4422259306 ns   4413752800 ns            1
```

It is expected that soft-floats are much slower than hardware float.

So overall, the effect of this patch on performance is:

- 1-2% slower on on ARM Mac because double and long double are the same on that platform and the added error bound computation is extra work.
- 4-10% faster on x86-64 because it uses 80-bit long double's that are hardware-accelerated through the legacy FPU features but generate slightly less performant assembly from not interacting well with the modern instruction sets.
- ~3 times faster with the default settings on the new Snapdragon X ARM CPUs, where clang and gcc generate soft-floats for long double.

The effect on portability is, that this patched version will produce the same predicate results on all of the above platform, because 64-bit double has the same behaviour for all of them, while the current code can produce different results for each of the above platforms because long double means something different for each of them.

The effect on behaviour is that this predicate will never produce the opposite of the true predicate result on any platform for degenerate cases like the current main code but more 0-results (point is found to be not inside or outside the disc but exactly on the boundary circle) for degenerate cases than the current main code.

This patch can be a step towards a filtered, exact predicate, if an exact in-circle calculation (like is proposed in a currently open PR for JTS https://github.com/locationtech/jts/pull/1094 ) is added later.

pramsey:
OK, I'm convinced. Could you throw some commentary into the patch with the references you are using and maybe a brief description. The magic constant in particular needs a reference. And a link to this PR.

tinko92:
I added comments with a link to the PR, a permalink (which is unfortunately long) to the line CGAL that uses this alternative incircle expression and to the paper and the accompanying zenodo repository that the error bound computation is based on. The Zenodo repo contains a docker image with a Jupyter Notebook where the general error bound generation for arbitrary expressions is shown.

Issue #1103, opened 2024-06-07
CompoundCurve construction should ensure sections are continuous


dbaston:
PostGIS raises an error with `COMPOUNDCURVE( (0 0, 1 1), CIRCULARSTRING (0 0, 1 1, 2 0))` but GEOS happily accepts it.

PR #1164, opened 2024-09-23
CompoundCurve: Add validateConstruction method

dbaston:
Resolves https://github.com/libgeos/geos/issues/1103

### DIFF
--- a/src/geom/CompoundCurve.cpp
+++ b/src/geom/CompoundCurve.cpp
@@ -12,6 +12,8 @@
  *
  **********************************************************************/
 
+#include <sstream>
+
 #include <geos/geom/CompoundCurve.h>
 #include <geos/geom/CoordinateFilter.h>
 #include <geos/geom/GeometryFactory.h>
@@ -25,7 +27,9 @@ CompoundCurve::CompoundCurve(std::vector<std::unique_ptr<SimpleCurve>>&& p_curve
                              const GeometryFactory& gf)
     : Curve(gf),
       curves(std::move(p_curves)),
-      envelope(computeEnvelopeInternal()) {}
+      envelope(computeEnvelopeInternal()) {
+    validateConstruction();
+}
 
 CompoundCurve::CompoundCurve(const CompoundCurve& other)
     : Curve(other),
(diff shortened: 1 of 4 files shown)


dr-jts:
Would this make sense to add to `Geometry.isValid()`?   And possibly to `IsValidOp`.

src/geom/CompoundCurve.cpp
```diff
@@ -307,5 +311,24 @@ CompoundCurve::reverseImpl() const
     return getFactory()->createCompoundCurve(std::move(reversed)).release();
 }
 
+void
+CompoundCurve::validateConstruction() const
+{
+    for (std::size_t i = 1; i < curves.size(); i++) {
+        const CoordinateXY& end = curves[i-1]->getCoordinatesRO()->back<CoordinateXY>();
+        const CoordinateXY& start = curves[i]->getCoordinatesRO()->front<CoordinateXY>();
```
rouault, line 319: I may miss something, but it is not obvious to me that you can't insert an empty curve in a CompoundCurve, in which case back() and front() are returning something invalid. At least if using directly the CompoundCurve::CompoundCurve(std::vector<std::unique_ptr<SimpleCurve>>&& p_curves,                              const GeometryFactory& gf) constructor. That might be hard/impossible to trigger through WKT.  But maybe through WKB ?

dbaston:
> Would this make sense to add to Geometry.isValid()? And possibly to IsValidOp.

I was aiming to handle it consistently with how we handle linear rings, where we simply disallow the construction of a ring with a single point, or a ring that doesn't close.

https://github.com/libgeos/geos/blob/bf2257aa3794b483fefe23497209a79d40492841/src/geom/LinearRing.cpp#L47-L68

src/geom/CompoundCurve.cpp
dbaston: You're right, this is possible and would crash. Of course it couldn't be so easy.

pramsey:
Probably the same as LinearRing, which is (wait for it) to disallow non-closed WKT but allow non-closed WKB.
```
select 'CompoundCurve((0 0, 0 1, 1 1), (1 2, 1 0, 0 0))'::geometry; 
ERROR:  incontinuous compound curve
```

rouault:
> So would this mean that a WKT or WKB serialization of an invalid compound curve could not be read? Sometimes users like to be able to read a structurally invalid geometry, so the library can be used to test validity and possible allow repair

I was wondering about that too. I've checked that GDAL allows for an extremely tiny gap of 1e-14, which probably is never met in practice.

# 1e-13 difference: rejected:
>>> g = ogr.CreateGeometryFromWkt('CompoundCurve((0 0,1 1),(1 1.0000000000001,3 3))')
ERROR 1: Non contiguous curves
ERROR 6: OGR Error: General Error

# 1e-14 difference: tolerated
>>> g = ogr.CreateGeometryFromWkt('CompoundCurve((0 0,1 1),(1 1.00000000000001,3 3))')
>>> print(g)
COMPOUNDCURVE ((0 0,1 1),(1 1,3 3))

What happens in practice though is when ingesting GML curves, and in particular when the compound curve chaining straight lines with circular portions where the circular portions are not defined as a CircularString, but as a ArcByCenterPoint, where you need to recompute the start and ending points from the center point and the start and end angles, which easily leads to precision issues. But I've ended up compensating for that in the GML import logic itself, because the gap is generally much larger than 1e-14.

dbaston:
> So would this mean that a WKT or WKB serialization of an invalid compound curve could not be read?

Correct, GEOS does not allow the construction of linear rings that are not closed, linear rings that have fewer than 3 points, or linestrings with a single point.

PR #1171, opened 2024-10-02
WIP: Add CircularArcIntersector

dbaston:
This PR adds a `CircularArcIntersector` that computes intersections between
circular arcs and/or line segments. The intersections may be points or arcs.
The interface generally mimics that of the `LineIntersector`.

Missing features:

Use of a `PrecisionModel`
Handling of Z, M
isProper()
isInteriorIntersection()

A challenge I ran into when writing test cases is the lossy conversion between how arcs are stored (three points) and how we generally must work with them (a center point, radius, two angles). This can cause small differences in the computed center point, which currently prevents the code from recognizing the arcs as co-circular. It may be necessary to add some tolerance in the center-point equality testing, possibly by using the scale of a provided `PrecisionModel`.

### DIFF
--- a/include/geos/algorithm/Angle.h
+++ b/include/geos/algorithm/Angle.h
@@ -204,6 +204,15 @@ public:
     ///
     static double normalizePositive(double angle);
 
+    /// Returns true if angle x is within the counterclockwise
+    /// arc from angle a to angle b
+    ///
+    /// @param angle angle to test
+    /// @param from starting angle of arc
+    /// @param to ending angle of arc
+    ///
+    /// @return true if `angle` is within [from, to]
+    static bool isWithinCCW(double angle, double from, double to);
 
     /// Computes the unoriented smallest difference between two angles.
     ///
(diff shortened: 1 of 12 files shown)


pramsey:
Is there a similar tolerance issue around arcs with infinite radius (aka straight lines (aka three control points in a line))?

dbaston:
> Is there a similar tolerance issue around arcs with infinite radius (aka straight lines (aka three control points in a line))?

Good point. See the final (failing) test added in https://github.com/libgeos/geos/pull/1171/commits/9910456b3acc5be3fd651f65d7fdcb6de8e65158.

dbaston:
Current status of this: tests passing except for:
- CircularArcIntersector test 40 - the computed intersection point is outside the arbitrary distance tolerance 
- CircularArcIntersector 42 and 53. These relate to arcs that intersect at an endpoint. In both cases the computed intersection points (incorrectly) do not fall on the input arcs. We probably need some special logic for endpoint intersections, as is the case for linear segment intersection.

I think it is mergeable as-is; it is not accessible though the C API, and we can return to it later to improve robustness after more of the curve integration is completed. Thoughts, @pramsey ?

pramsey:
Yes, merge it so it can be worked on in core.

PR #1183, opened 2024-10-25
CMake: use enable_testing() instead of include(CTest)

rouault:
According to Craig Scott in https://discourse.cmake.org/t/is-there-any-reason-to-prefer-include-ctest-or-enable-testing-over-the-other/1905/2 , using include(CTest) adds unnecessary clutter that is only needed for dashboard submission. enable_testing() is enough otherwise

"Port" of https://github.com/OSGeo/shapelib/pull/162

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -382,9 +382,12 @@ add_subdirectory(capi)
 #-----------------------------------------------------------------------------
 # Tests
 #-----------------------------------------------------------------------------
+
+option(BUILD_TESTING "Build the testing tree." ON)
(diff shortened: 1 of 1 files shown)


rouault:
ok, closing that as GEOS does appear to use include(CTest) extra features, at least the [Linux (clang++-11, clang-11, Debug, 14, 64, clang-11, 3.23.*, ubuntu-22.04)](https://github.com/libgeos/geos/actions/runs/11524404342/job/32084483277?pr=1183#logs) configuration

PR #1186, opened 2024-10-29
Move using-declarations into class body to avoid polluting the global namespace

kikitte:
Polygon, LineString, etc.,  are declared in the global namespace with the using-declaration keyword, which is very likely to confilict with other types of the same name.

Move them to geos::operation::buffer fixes this.

https://github.com/libgeos/geos/blob/46972ad89a143d01ed3ec9a6c2d91d904a87ab0a/include/geos/operation/buffer/OffsetCurve.h#L39-L50

### DIFF
--- a/include/geos/operation/buffer/OffsetCurve.h
+++ b/include/geos/operation/buffer/OffsetCurve.h
@@ -37,13 +37,6 @@ class SegmentMCIndex;
 }
 }
 
-using geos::geom::Coordinate;
-using geos::geom::CoordinateSequence;
-using geos::geom::Geometry;
-using geos::geom::GeometryFactory;
-using geos::geom::LineString;
-using geos::geom::Polygon;
-
 namespace geos {
 namespace operation {
 namespace buffer {
@@ -88,7 +81,12 @@ namespace buffer {
  *
  */
 class GEOS_DLL OffsetCurve {
-
+    using Coordinate = geos::geom::Coordinate;
+    using CoordinateSequence = geos::geom::CoordinateSequence;
+    using Geometry = geos::geom::Geometry;
+    using GeometryFactory = geos::geom::GeometryFactory;
+    using LineString = geos::geom::LineString;
+    using Polygon = geos::geom::Polygon;
 
 private:
 
--- a/include/geos/algorithm/CGAlgorithmsDD.h
+++ b/include/geos/algorithm/CGAlgorithmsDD.h
@@ -29,13 +29,12 @@ class CoordinateSequence;
 }
 }
 
-using namespace geos::math;
-
 namespace geos {
 namespace algorithm { // geos::algorithm
 
 /// Implements basic computational geometry algorithms using extended precision float-point arithmetic.
 class GEOS_DLL CGAlgorithmsDD {
+    using DD = geos::math::DD;
 
 public:
 
--- a/include/geos/algorithm/MinimumAreaRectangle.h
+++ b/include/geos/algorithm/MinimumAreaRectangle.h
@@ -32,15 +32,6 @@ namespace geos {
 	}
 }
 
-using geos::geom::CoordinateSequence;
-using geos::geom::CoordinateXY;
-using geos::geom::Geometry;
-using geos::geom::GeometryFactory;
-using geos::geom::LineSegment;
-using geos::geom::LineString;
-using geos::geom::Polygon;
-
-
 namespace geos {
 namespace algorithm { // geos::algorithm
 
@@ -65,6 +56,13 @@ namespace algorithm { // geos::algorithm
  *
  */
 class GEOS_DLL MinimumAreaRectangle {
+    using CoordinateSequence = geos::geom::CoordinateSequence;
+    using CoordinateXY = geos::geom::CoordinateXY;
+    using Geometry = geos::geom::Geometry;
+    using GeometryFactory = geos::geom::GeometryFactory;
+    using LineSegment = geos::geom::LineSegment;
+    using LineString = geos::geom::LineString;
+    using Polygon = geos::geom::Polygon;
 
 private:
 
--- a/include/geos/algorithm/PolygonNodeTopology.h
+++ b/include/geos/algorithm/PolygonNodeTopology.h
@@ -25,9 +25,6 @@ class CoordinateXY;
 }
 }
 
-using geos::geom::CoordinateXY;
-
-
 namespace geos {
 namespace algorithm { // geos::algorithm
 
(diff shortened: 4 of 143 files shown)


dbaston:
Should we instead remove the `using` declarations from the header file altogether?

kikitte:
> Should we instead remove the `using` declarations from the header file altogether?

Sure, looks like geos prefer explicit namespace in header file but `using namespace` in souce file.

kikitte:
Hi , I thkink I known what you mean, and I notice that a lot of code using such `using` style which would make types appear in global namespace.

Since c++ api is one way of using geos, such effort to avoid multiple declaration is necessary and meaningful. 

The key is using which code style to sove this, remove `using` then using explicit namespace in header file, or any other?

pramsey:
The practice in the code base is a mishmash, I can see trying to move to fewer uses of `using` in headers as a general practice. How about you do that for this PR, and we'll make it a long term ambition.

dr-jts:
> Should we instead remove the `using` declarations from the header file altogether?

Won't that mean a lot more scoping of names used in multiple places in the file?

Isn't `using` intended to cut down on this?  It makes sense to limit scope to namespaces, but inside namespaces it seems like a good practice for code brevity.

include/geos/operation/buffer/OffsetCurve.h
```diff
@@ -93,27 +86,27 @@ class GEOS_DLL OffsetCurve {
 private:
 
     // Members
-    const Geometry& inputGeom;
+    const geom::Geometry& inputGeom;
```
rouault, line 89: what about doing the following (generalized to similar cases), to both reduce verbosity and amount of lines changed ?
```suggestion
    using Geometry = geom::Geometry;

    const Geometry& inputGeom;
```

dr-jts: So use `using` as a [_type alias_](https://en.cppreference.com/w/cpp/language/type_alias) instead of a [_using declaration_](https://en.cppreference.com/w/cpp/language/namespace)?  Is this safer in some way?  

What is the scope of the alias?  Is it just the namespace within the `cpp` file including the header?  Or is it the entire namespace (seems unlikely, since how would other compilation units see it?)

rouault: > What is the scope of the alias?

the class in which you put it

Not saying this is a good idea. Mostly for brainstorming

pramsey:
Why not just moving the `using geos::geom::Coordinate` inside the name space that contains the class definitions in the header, as the original revision of the PR suggested? Seems like we avoid polluting the global name space and can have as much terseness as we feel like, right?

dbaston:
> The practice in the code base is a mishmash, I can see trying to move to fewer uses of using in headers as a general practice. 

If you check `grep -R "^using.*;" include` from the project root, it looks like the use of `using` in headers is a pretty recent addition.

> Isn't using intended to cut down on this? It makes sense to limit scope to namespaces, but inside namespaces it seems like a good practice for code brevity.

I think it's a good practice inside _implementation_ files, where you're not exporting the shorthand to anyone including the header.

> Why not just moving the using geos::geom::Coordinate inside the name space that contains the class definitions in the header, as the original revision of the PR suggested? Seems like we avoid polluting the global name space and can have as much terseness as we feel like, right?

Because then we've polluted the `geos::operation::buffer` namespace, so anyone including this header now has a type called `geos::operation::buffer::Coordinate`. Is a `geos::operation::buffer::Coordinate` the same as a `geos::geom::Coordinate`? You'd need to check the source to see.

If you instead put the `using` declaration inside the class, as Even showed, you can make it private and avoid exporting it to everybody.

dr-jts:
> If you instead put the `using` declaration inside the class, as Even showed, you can make it private and avoid exporting it to everybody.

Do you mean "put the `using` _**type alias**_ inside the class"?  (I.e. the `using Name = other:Name` syntax)

If so, does the following capture the recommended practice?:

* In header files,  names can be shortened by adding _type aliases_ such as `using Geometry = geom::Geometry;` **inside the class definition**
* In implementation `cpp` files, names can be shortened by adding _using declarations_ such as `using geom::Geometry;`

dbaston:
> Do you mean "put the using type alias inside the class"? (I.e. the using Name = other:Name syntax)

Yes

> If so, does the following capture the recommended practice?:
>    In header files, names can be shortened by adding type aliases such as using Geometry = geom::Geometry; inside the class definition

I don't know if I'd call it a recommend practice, just trying to clarify the difference between putting it inside and outside the class definition. If you look at a header like `OffsetCurve.h` I'm not sure it's worth it over just writing `geom::Geometry`. Others could reach a different conclusion.

>    In implementation cpp files, names can be shortened by adding using declarations such as using geom::Geometry;

Yes

kikitte:
I'am back, I've learned a lot from your  conversations. 

It looks like "put the using type alias inside the class" is an acceptable solution, I am trying to fix all the header files in this way: **type alias always on the top of the class**. So alias is private to the class without leaking. 

But according to my statistics, there are at least 59 header files need to fix, it may take several days to finish.

```C++
class GEOS_DLL OffsetCurve {
  using Coordinate = geos::geom::Coordinate;
  using CoordinateSequence = geos::geom::CoordinateSequence;
  using Geometry = geos::geom::Geometry;
  using GeometryFactory = geos::geom::GeometryFactory;
  using LineString = geos::geom::LineString;
  using Polygon =geos::geom::Polygon;

private:

    // Members
    const Geometry& inputGeom;
    double distance;
    bool isJoined = false;
```

dr-jts:
Are the forward declarations still needed?  Would be nice if that cruft could be eliminated.
```
// Forward declarations
namespace geos {
namespace geom {
class Coordinate;
class CoordinateSequence;
class Geometry;
class LineString;
class Polygon;
}
namespace operation {
namespace buffer {
class OffsetCurveSection;
class SegmentMCIndex;
}
}
}
```

pramsey:
I think there's a compile time penalty to including everything.

dbaston:
Forward declarations break the dependency chain, so a change in `LineString.h` does not force a recompilation of everything that includes `OffsetCurve.h`.

dr-jts:
> Forward declarations break the dependency chain, so a change in `LineString.h` does not force a recompilation of everything that includes `OffsetCurve.h`.

Yeesh, how fiddly.  But this usage should be encouraged, sounds like.

kikitte:
Ready for review. 

Note that sometimes I prefer `using namespace geos::geom` in implementation file, I think it is acceptable and pretty common in geos source code.

rouault (approved):

pramsey (approved):

dbaston (approved):

PR #1128, opened 2024-07-19
Fix typos

mwtoews:
This fixes numerous typos using [codespell](https://github.com/codespell-project/codespell) interactively. There are a few places that I made a guess to the correct word.

---

Some typos for object names are kept (for now?) including:

- Extracter (normally [extractor](https://en.wiktionary.org/wiki/extractor)), and related:
  -  ComponentCoordinateExtracter
  - GeometryExtracter
  - LinearComponentExtracter
  - PointExtracter
  - PolygonalExtracter
  - PolygonExtracter
- Translater (well at least Wiktionary has an entry for [translater](https://en.wiktionary.org/wiki/translater))
- Geometrys

PR #1194, opened 2024-11-18
Fix typos, add CI job to help catch them

mwtoews:
Following on #1128, this PR fixes a few more typos (some from me).

It also adds configuration to enable CI to run `codespell` to catch them before they get in. Third-party tools should be ignored, as is done with the configuration.

False-positive words to ignore can be added to `tools/codespell.ignore`.

### DIFF
--- a/src/algorithm/MinimumBoundingCircle.cpp
+++ b/src/algorithm/MinimumBoundingCircle.cpp
@@ -45,7 +45,7 @@ namespace algorithm { // geos.algorithm
 std::unique_ptr<Geometry>
 MinimumBoundingCircle::getCircle()
 {
-    //TODO: ensure the output circle contains the extermal points.
+    //TODO: ensure the output circle contains the extremal points.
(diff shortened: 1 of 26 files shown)


src/algorithm/MinimumBoundingCircle.cpp
```diff
@@ -45,7 +45,7 @@ namespace algorithm { // geos.algorithm
 std::unique_ptr<Geometry>
 MinimumBoundingCircle::getCircle()
 {
-    //TODO: ensure the output circle contains the extermal points.
```
dbaston, line 48: "extremal" is correct here

mwtoews: Good catch. This was suggested too: `extermal ==> external, extremal`

pramsey (approved):
Unless this is WIP, looks fine to merge to me.

PR #1196, opened 2024-11-18
CoordinateSequence: Add applyAt method

dbaston:
Add an `applyAt` method to make it simpler to get the correct `Coordinate` type from a sequence and pass it to a function.


pramsey (approved):
Nice to see that muck pulled back into CoordinateArray.

PR #1197, opened 2024-11-18
CI: replace deprecated macos-12 with beta macos-15 with C++23

mwtoews:
According to [the runner image docs](https://github.com/actions/runner-images/tree/main) (and in GHA warnings), `macos-12` is deprecated. Replace this with `macos-15` for macOS 15 Arm64 beta with Xcode 16.0. This is also configured to select [C++23](https://developer.apple.com/xcode/cpp/#c++23).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -450,10 +450,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


mwtoews:
New CI job has a clean build and test outputs, no warnings anywhere.

PR #1209, opened 2024-12-20
Update .gitignore to remove *.cmake

ianhbell:
I am building a package that includes geos as a dependency, with compilation targeting Python, and builds happening with ``scikit-build-core``. I had a mysterious build failure for missing files in the ``cmake`` folder inside the build folder when built with ``pipx run build``, and tracked down the problem to the fact that the build runs a git clean and doesn't copy over files that are ignored by ``.gitignore``. Fixing that (as in this PR), allowed that part of the build to work properly. I think this change is harmless since there are already ``.cmake`` files in the repo so they should probably not be in the ``.gitignore`` anyhow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ Makefile
 *.swp
 *.trs
 *.exe
-*.cmake
 CMakeFiles
 CMakeCache.txt
 CMakeLists.txt.user


dbaston:
> I think this change is harmless since there are already .cmake files in the repo so they should probably not be in the .gitignore anyhow.

I see this was initially added in https://github.com/libgeos/geos/commit/dd1be7331e5b9c3665f58679b6968a3890ea385d to support in-tree builds. I'm OK with the proposed change - I feel like the expectation is that if you do an in-tree build, you're going to end up with bunch of garbage in your tree, and the solution is to not build in your tree.

mwtoews (approved):
Thanks @dbaston for the context of the change. In-source builds are not a normal method of building, so this item shouldn't be here. For reference, recommended `.gitignore` items for CMake projects is [here](https://github.com/github/gitignore/blob/main/CMake.gitignore).

And thanks again @ianhbell for the helpful PR!

PR #1210, opened 2024-12-20
Update CMakeLists.txt to use absolute path

ianhbell:
Use an absolute path to be more sure that the path is correct


mwtoews (approved):
Thanks!

PR #1225, opened 2025-01-13
Fix inscribed circle initialization

unjambonakap:
Fix a bug in `MaximumInscribedCircle` that gives a bad result

```
import shapely, shapely.constructive
import numpy as np
poly = shapely.from_geojson('{"type":"Polygon","coordinates":[[[0.0,-10.0],[-7.07107,-7.07107],[-10.0,0.0],[-7.07107,7.07107],[0.0,10.0],[7.07107,7.07107],[10.0,0.0],[7.07107,-7.07107],[0.0,-10.0]]]}')

u = shapely.constructive.maximum_inscribed_circle(poly, tolerance=0.1)
a, b = np.array(u.coords)
print(a,b,np.linalg.norm(a-b))

```

Actual: 
`[0.       3.535535] [-2.28553181  9.05330273] 5.972387826209451`

Expected:
`[0. 0.] [ 8.53553655 -3.53553126] 9.238796754579768`

### DIFF
--- a/src/algorithm/construct/MaximumInscribedCircle.cpp
+++ b/src/algorithm/construct/MaximumInscribedCircle.cpp
@@ -164,8 +164,8 @@ MaximumInscribedCircle::distanceToBoundary(double x, double y)
 MaximumInscribedCircle::Cell
 MaximumInscribedCircle::createInteriorPointCell(const Geometry* geom)
 {
-    Coordinate c;
     std::unique_ptr<Point> p = geom->getInteriorPoint();
+    Coordinate c(p->getX(), p->getY());
     Cell cell(p->getX(), p->getY(), 0, distanceToBoundary(c));
     return cell;
 }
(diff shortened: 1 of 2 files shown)


pramsey:
As you can see, this fails regression, though only regression tests on invalid inputs. Unfortunate! 
* in test 5, the result moves from the middle of the (invalid) geometry to the end, which is "less good" though again we are talking about invalid input and a zero-radius output...
* similar story in test 7, moving from the middle to the end
```
---> group: geos::algorithm::construct::MaximumInscribedCircle, test: test<5>
     problem: assertion failed
     failed assertion: `x coordinate does not match: expected `1.5000000000000000e+02` actual `1.0000000000000000e+02` with precision `1.0000000000000000e-02``

---> group: geos::algorithm::construct::MaximumInscribedCircle, test: test<7>
     problem: assertion failed
     failed assertion: `x coordinate does not match: expected `2.0000000000000000e+00` actual `1.0000000000000000e+00` with precision `1.0000000000000000e-02``
```

unjambonakap:
The fact that it used to hit the centroid is purely by luck:
Before test5:
- First cell distance is bad (because computed with (0,0), outside)
- first iteration, gets improved with 0 (this happens to be the centroid)
- no more improvements

Now:
- first cell is the best one, attributed to the computed interior point (on the boundary, 100 100)(

PR #1226, opened 2025-01-14
Fix invalid values from CoverageUnionNG

mwtoews:
This resolves an invalid value message shown in shapely:
```pycon
>>> import shapely
>>> shapely.geos_version_string
'3.13.0'
>>> shapely.coverage_union_all([])
/home/mtoews/src/shapely/shapely/shapely/set_operations.py:505: RuntimeWarning: invalid value encountered in coverage_union
  return lib.coverage_union(collections, **kwargs)
<GEOMETRYCOLLECTION EMPTY>
```

This is caused by a division by zero in GEOS' `CoverageUnion::geomunion()` with coverage inputs that are empty or don't have an area.

[Tests 9 to 14](https://github.com/libgeos/geos/blob/3.13.0/tests/unit/operation/overlayng/CoverageUnionNGTest.cpp#L141-L205) were silently throwing this warning too, as these input coverages don't have an area.

### DIFF
--- a/src/operation/overlayng/CoverageUnion.cpp
+++ b/src/operation/overlayng/CoverageUnion.cpp
@@ -33,7 +33,6 @@ namespace overlayng { // geos.operation.overlayng
 std::unique_ptr<Geometry>
 CoverageUnion::geomunion(const Geometry* coverage)
 {
-    double area_in = coverage->getArea();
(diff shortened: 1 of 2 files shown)


mwtoews:
This is probably too minor to add to NEWS, but worth backporting where applicable.

PR #1228, opened 2025-01-14
geosop: show most floating-point exceptions in verbose mode

mwtoews:
This shows floating-point exception information with `geosop` in verbose mode.
```bash
$ ./bin/geosop -a "POLYGON EMPTY" -b "POLYGON EMPTY" coverageUnion -v
Input A: WKT literal
Read 1 geometries, 0 vertices  -- 164 usec
Input B: WKT literal
Read 1 geometries, 0 vertices  -- 9 usec
[ 1] coverageUnion: A[1] Polygon( 0 )  -> MultiPolygon( 0 )  --  82 usec
MULTIPOLYGON EMPTY
Operation raised floating-point environment flag(s): FE_INVALID
Ran 1 coverageUnion ops ( 0 vertices)  -- 82 usec    (GEOS 3.14.0dev)
```

It shows most floating-point exception flags except for FE_INEXACT, which is almost always harmless, e.g.<br /> `geosop -v -a "POINT(0 0)" -b "POINT(1 1)" distance` would raise it since the answer is irrational (sqrt(2) = 1.41421...)

### DIFF
--- a/util/geosop/GeosOp.cpp
+++ b/util/geosop/GeosOp.cpp
@@ -26,6 +26,7 @@
 #include <geos/io/WKBStreamReader.h>
 #include <geos/io/WKBWriter.h>
 
+#include <cfenv>
 #include <fstream>
 #include <iostream>
 #include <sstream>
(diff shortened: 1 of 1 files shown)


dr-jts:
Good idea.

Issue #1136, opened 2024-08-13
OverlayNG should return Point for intersection between area and zero-length LineString


pramsey:
To match with the RelateNG logic, overlay should return a Point for intersection between Polygon and zero-length LineString. Currently it returns an empty LineString.

dr-jts:
This behaviour needs to apply in multiple situations.  E.g.

* Area with zero-len LineString
* Area with MultiLineString containing zero-len element
* Linear geometry with (Multi)LineString with zero-len element
* Point geometry with (Multi)LineString with zero-len element

In general, the intersection of a zero-length LineString element with any geometry should produce an equivalent Point output (unless it is covered by another linear input element).

dr-jts:
Another question is whether zero-length lines in the result should be converted to Points, or left as zero-length LineStrings.

Converting to Points ensures that the result is valid (according to the current validity rules, which in the case of zero-length lines are perhaps overly restrictive).

Keeping as zero-length lines ensures that linear results are homogeneous, which allows them to be used in iterated overlays.  This also can be considered to follow the [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

PR #1229, opened 2025-01-15
Fix overlay heuristic for GeometryCollections with empty elements

### LINKED ISSUE #1224: Union Operations returns collections containing empty elements
Relates to: https://trac.osgeo.org/postgis/ticket/5835

Reproduce in GEOS:
```
geosop -a 'GEOMETRYCOLLECTION(POLYGON EMPTY, POINT EMPTY, LINESTRING(10 10, 10 10))' -b 'POINT(1 1)' union
-- reuslt:{GEOMETRYCOLLECTION (POLYGON EMPTY, POINT EMPTY, LINESTRING (10 10, 10 10), POINT (1 1))}

geosop -a 'GEOMETRYCOLLECTION(POLYGON EMPTY, POINT EMPTY, LINESTRING(10 10, 10 10), POINT(1 1))' unaryUnion
-- result:{GEOMETRYCOLLECTION (POINT (1 1), LINESTRING EMPTY)}
```

dr-jts:
This fixes the behaviour of the "combine disjoint" overlay heuristic for GeometryCollections containing empty elements (which are now removed).   

Fixes #1224 

(Note: part of #1224 requires #1136 to be fixed as well).

The PR also:
* centralizes overlay heuristic logic in `HeuristicOverlay` (moving code from `Geometry`)
* modularizes the `HeuristicOverlay` code
* improves `HeuristicOverlay` semantics to return empty results as an empty typed atomic geometry instead of an empty GeometryCollection (this matches `OverlayNG`)
* removes unnecessary dependencies on `HeuristicOverlay` (by calling `OverlayNGRobust` directly where it is known the heuristics do not apply)
* fixes `XMLTester` to correctly handle operation A & B argument switching

### DIFF
--- a/src/geom/HeuristicOverlay.cpp
+++ b/src/geom/HeuristicOverlay.cpp
@@ -40,31 +40,63 @@ namespace geom { // geos::geom
 using operation::overlayng::OverlayNG;
 using operation::overlayng::OverlayNGRobust;
 
+bool isDisjointNonEmpty(const Geometry* g0, const Geometry* g1)
+{
+    if (g0->isEmpty() && g1->isEmpty())
+        return false;
+    return ! g0->getEnvelopeInternal()->intersects(g1->getEnvelopeInternal());
+}
+
+void 
+extractELements(const Geometry* g, std::vector<std::unique_ptr<Geometry>>& v)
+{
+    const GeometryCollection* coll;
+    if (nullptr != (coll = dynamic_cast<const GeometryCollection*>(g))) {
+        //-- buffer num geoms, since can be expensive for nested collections
+        std::size_t numGeoms = g->getNumGeometries();
+        for(std::size_t i = 0; i < numGeoms; ++i) {
+            //-- recurse to handle nested GCs
+            extractELements(coll->getGeometryN(i), v);
+        }
+    }
+    else if (g->isEmpty()) {
+        return;
+    }
+    else {
+        v.push_back(g->clone());
+    }
+}
+
+std::unique_ptr<Geometry>
+reduceCombine(const Geometry* g0, const Geometry* g1)
+{
+    // Allocated for ownership transfer
+    std::vector<std::unique_ptr<Geometry>> v;
+    v.reserve(g0->getNumGeometries() + g1->getNumGeometries());
+    extractELements(g0, v);
+    extractELements(g1, v);
+    return g0->getFactory()->buildGeometry(std::move(v));
+}
+
 std::unique_ptr<Geometry>
 HeuristicOverlay(const Geometry* g0, const Geometry* g1, int opCode)
 {
-    std::unique_ptr<Geometry> ret;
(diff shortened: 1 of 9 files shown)


src/geom/HeuristicOverlay.cpp
```diff
@@ -40,31 +47,63 @@ namespace geom { // geos::geom
...
 
+bool isDisjointNonEmpty(const Geometry* g0, const Geometry* g1)
+{
+    if (g0->isEmpty() && g1->isEmpty())
+        return false;
+    return ! g0->getEnvelopeInternal()->intersects(g1->getEnvelopeInternal());
+}
+
+void 
+extractELements(const Geometry* g, std::vector<std::unique_ptr<Geometry>>& v)
```
dbaston, line 58:
```suggestion
extractElements(const Geometry* g, std::vector<std::unique_ptr<Geometry>>& v)
```

src/geom/HeuristicOverlay.cpp
```diff
@@ -40,31 +47,63 @@ namespace geom { // geos::geom
...
+{
+    const GeometryCollection* coll;
+    if (nullptr != (coll = dynamic_cast<const GeometryCollection*>(g))) {
```
dbaston, line 61:
```suggestion
    if (const auto* coll = dynamic_cast<const GeometryCollection*>(g))) {
```

src/geom/HeuristicOverlay.cpp
```diff
@@ -108,6 +148,24 @@ HeuristicOverlay(const Geometry* g0, const Geometry* g1, int opCode)
...
+    StructuredCollection s1(g1);
+    switch (opCode) {
+    case OverlayNG::UNION:
+        return s0.doUnion(s1);
+    case OverlayNG::DIFFERENCE:
+        return s0.doDifference(s1);
+    case OverlayNG::SYMDIFFERENCE:
+        return s0.doSymDifference(s1);
+    default: // only OverlayNG::INTERSECTION
+        return s0.doIntersection(s1);
```
dbaston, line 165:
```suggestion
    case OverlayNG::INTERSECTION:
        return s0.doIntersection(s1);    
    default:
        throw util::IllegalArgumentException("Invalid opcode");
```

tests/xmltester/XMLTester.cpp
```diff
@@ -178,6 +178,17 @@ tolower(std::string& str)
         );
 }
 
+void
+toupper(std::string& str)
```
dbaston, line 182: Already implemented in `geos/util/string.h`

dr-jts: great, thanks

src/geom/HeuristicOverlay.cpp
dr-jts: what is the change?

dbaston: second letter of "elements"

PR #1230, opened 2025-01-22
Refactor XMLTester

dr-jts:
Improves `XMLTester` by modularizing it using an additional `Test` class and a few lambda functions.  This allows factoring out common code to achieve DRY.

* improves maintainability
* easier to add new operations
* reduces risk of ops missing expected behaviour (e.g. swapping A and B args, test reporting)
* reduces code size (> 800 lines removed)
* adds test file line number reporting

This has also uncovered some issues with result testing (in particular, using topological rather than exact equals).  Fixing this requires fixing a few tests, so will be done in a subsequent PR.

### DIFF
--- a/tests/xmltester/XMLTester.cpp
+++ b/tests/xmltester/XMLTester.cpp
@@ -842,7 +842,7 @@ Test::areaDelta(const geom::Geometry* a, const geom::Geometry* b, std::string& r
 }
 
 void
-Test::checkResult( Geometry* res ) 
+Test::checkResult( const std::unique_ptr<Geometry>& res ) 
 {
     std::string expectedRes = opRes;
     std::unique_ptr<Geometry> gExpectedRes(tester.parseGeometry(expectedRes, "expected"));
@@ -852,6 +852,7 @@ Test::checkResult( Geometry* res )
     gActualRes->normalize();
 
     isSuccess = false;
+    //TODO: change to equalsExact, since compareTo doesn't check empty type
     if (gExpectedRes->compareTo(gActualRes.get()) == 0) {
         isSuccess = true;
     }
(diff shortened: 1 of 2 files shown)


tests/xmltester/XMLTester.cpp
```diff
@@ -843,31 +841,84 @@ XMLTester::areaDelta(const geom::Geometry* a, const geom::Geometry* b, std::stri
         ss << *geomSD;
         ss << std::endl;
     }
-
     return diffScore;
 }
 
+void
+Test::checkResult( const std::unique_ptr<Geometry>& result )
```
dbaston, line 848: Taking a reference to a `unique_ptr` here (and elsewhere) strikes me as odd...why not `const Geometry& result` ?

dr-jts: great, will try that out.

dr-jts: One reason is that this requires the caller to use `*gUniquePtr`, which always strikes me as ugly (and slightly magical)

dbaston: Dereferencing a pointer with `*` is pretty standard C++ usage, though.

dr-jts: After some reading ([here](https://stackoverflow.com/questions/30905487/how-can-i-pass-stdunique-ptr-into-a-function) and [here](https://github.com/CppCon/CppCon2014/blob/master/Presentations/Back%20to%20the%20Basics!%20Essentials%20of%20Modern%20C%2B%2B%20Style/Back%20to%20the%20Basics!%20Essentials%20of%20Modern%20C%2B%2B%20Style%20-%20Herb%20Sutter%20-%20CppCon%202014.pdf)) I'm convinced.  The ownership of the parameter is an external concern which should not be exposed inside the functions.

Issue #967, opened 2023-09-29
Add algorithm for overlay of geometries with a rectangular grid


dbaston:
I am working on a project to expose the [exactextract](https://github.com/isciences/exactextract) library to Python in a way similar to the R package [exactextractr](https://github.com/isciences/exactextractr).

Instead of creating separate Python bindings for this library, I am considering attempting to move much of the code of this library into GDAL where it would become accessible through GDAL's Python bindings. Related but not necessarily dependent on this, I am thinking about whether some of the code could move into GEOS -- specifically, an optimized intersection calculation between rectangular grids and polygonal/linear geometries.

The result of the overlay is most commonly expressed as the fraction of each grid cell that is covered by a polygon, although it is also possible to get the total length of linear geometries within each grid cell or (more experimentally) a Geometry representation of each polygon/grid cell intersection, e.g.

![image](https://github.com/libgeos/geos/assets/6318931/e3e96bf3-8be0-409e-b8a9-9e7a5b3739a8)

This code has a strong test suite and has had pretty good public adoption since the R package was published in 2019, so I think it can be considered reasonably mature.

I don't have a detailed proposal yet, but I imagine that a minimal-footprint addition to GEOS would add something along these lines:

```
// define a grid
GEOSGrid* GEOSGrid_create(double xmin, double ymin, double xmax, double ymax, double dx, double dy);

// calculate a Grid coincident with `grid` but covering only `geom`
GEOSGrid* GEOSGrid_clip(const GEOSGrid* grid, const GEOSGeometry* geom);

// partition a polygonal geometry by a grid
GEOSGeometry* GEOSGrid_intersection(const GEOSGrid* grid, const GEOSGeometry* geom);

// calculate a raster with the fraction of each grid cell that is covered
// to avoid introducing a matrix type (and having slow access through non-inlined function calls)
// just return a float* and access values with x[i*cols + j] or something
float* GEOSGrid_intersectionAreas(const GEOSGrid*, const GEOSGeometry* geom);
```

Just thought I'd create an issue for this in case anyone has thoughts. I'm going to aim to have a better formed proposal before the Vienna code sprint.

mwtoews:
The proposed GEOSGrid_create has xmin, xmax and dx as floats, which is potentially problematic since dx may not be a multiple of (xmax - ymin). How about something based on a grid shape with offset to the top left corner:
```
[...]

PR #1232, opened 2025-02-04
Add grid subdivision / rasterization algorithms

dbaston:
As discussed in https://github.com/libgeos/geos/issues/967, adds functions for:
(1) subdividing a geometry by a rectangular grid
(2) computing the fraction of each cell in a grid that is covered by a polygon

Performance in the included benchmark is 1-3 orders of magnitude faster than repeated calls to `Geometry::intersection`

image

Code was adapted and pared down from https://github.com/isciences/exactextract


pramsey:
This looks very exciting, I'm fine w/ the CAPI and not qualified to judge the implementation. The tests look good, maybe see if any of the codecov suggestions for complete coverage are easy to knock off.

dbaston:
Still working through a couple of minor issues with this. I think it could go in the beta, I'm not seeing anything I can't resolve before release.

dbaston:
Current status: this produces valid polygons in all cases that I've tested, but the polygons do not necessarily form a valid coverage if polygon boundaries follow grid cell boundaries. Here is an example:

image

Subdivided to produce:

image

In this case, some of the fully interior polygons are generated as squares without redundant nodes along their edges. As a result, the polygons do not form a valid coverage. Because the missing nodes may only be on a horizontal or vertical edge, they do not cause an overlap or gap.

dbaston:
> It isn't really clear to me where the missing nodes are,

I mean that polygon "A" should have a node at the location pointed to by the arrow. If it does not have a node at this location there will be no gaps or overlaps between adjacent polygons, but the polygons will still not be considered to form a valid coverage.

image

Either way, I believe all issues in this category have now been addressed.

PR #1243, opened 2025-02-20
Add a GEOS_WISDOM_OF_THE_DAY

rouault:
Similar move as https://github.com/OSGeo/gdal/pull/11865 , https://github.com/MapServer/MapServer/pull/7233 , https://github.com/OSGeo/PROJ/pull/4406

### DIFF
--- a/include/geos/version.h.in
+++ b/include/geos/version.h.in
@@ -30,6 +30,10 @@
 #define GEOS_VERSION "@VERSION@"
 #endif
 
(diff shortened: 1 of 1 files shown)


dbaston:
I am going to close this after discussing with @rouault . If I did not know the author, I would find the message conveyed here unclear (are we supporting DEI programs, or making a joke about them?). Other issues have been raised in the PRs linked above. In any case, a change like this really something that needs support of the PSC.

PR #1260, opened 2025-04-24
C API: Add GEOSGeoJSONWriter_setOutputDimension

oreilles:


capi/geos_ts_c.cpp
```diff
@@ -3745,6 +3745,14 @@ extern "C" {
         });
     }
 
+    void
+    GEOSGeoJSONWriter_setOutputDimension_r(GEOSContextHandle_t extHandle, GeoJSONWriter* writer, int dim)
+    {
+        execute(extHandle, [&]() {
+            writer->setOutputDimension(static_cast<uint8_t>(dim));
```
pramsey, line 3752: Seems like this will overflow pretty easily...

capi/geos_c.cpp
```diff
@@ -1577,6 +1577,12 @@ extern "C" {
         return GEOSGeoJSONWriter_writeGeometry_r(handle, writer, g, indent);
     }
 
+    void
+    GEOSGeoJSONWriter_setOutputDimension(GeoJSONWriter* writer, int dim)
+    {
+        GEOSGeoJSONWriter_setOutputDimension_r(handle, writer, dim);
```
pramsey, line 1583: Add a test?

capi/geos_ts_c.cpp
Oreilles: I just copied the implementation from `GEOSWKTWriter_setOutputDimension_r` and `GEOSWKBWriter_setOutputDimension_r`, but yeah that's not exactly safe.

dr-jts: The only dimensions supported are 2, 3 and 4.  And `uint8_t` is used pretty widely for dimensions in the codebase.

pramsey: Yeah, I guess if someone overflows it they just get a garbage number, which in turn is what they put in (garbage) so, it's all on them. As long as somewhere deeper in the code base we are checking for ==2 or ==3 or ==4

dr-jts: There's a check in [`GeoJSONWriter::setOutputDimension`](https://github.com/libgeos/geos/blob/main/src/io/GeoJSONWriter.cpp#L49), so that seems pretty safe.

Oreilles:
Just noticed that `GEOSWKBWriter_getOutputDimension_r` use 0 as the error code, while `GEOSWKTWriter_getOutputDimension_r` uses -1. What value is the recommended one nowadays ?

PR #1268, opened 2025-05-30
CoverageSimplify: add progress callback and add GEOSCoverageSimplifyVWWithProgress_r()

rouault:
Implement what has been discussed with @dr-jts in https://github.com/OSGeo/gdal/pull/12483#issuecomment-2920633997

GDAL use of it in https://github.com/rouault/gdal/compare/algorithm_progress...rouault:gdal:GEOSCoverageSimplifyVWWithProgress_r

CC @dbaston

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -2061,5 +2061,15 @@ extern "C" {
         return GEOSCoverageSimplifyVW_r(handle, input, tolerance, preserveBoundary);
     }
 
+    Geometry*
+    GEOSCoverageSimplifyVWWithProgress(const Geometry* input, double tolerance,
+                                       int preserveBoundary,
+                                       GEOSProgressCallback_r progressFunc,
+                                       void* progressUserData)
+    {
+        return GEOSCoverageSimplifyVWWithProgress_r(
+            handle, input, tolerance, preserveBoundary, progressFunc, progressUserData);
+    }
+
 
 } /* extern "C" */
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -1101,6 +1101,13 @@ extern GEOSGeometry GEOS_DLL *GEOSUnaryUnion_r(
     GEOSContextHandle_t handle,
     const GEOSGeometry* g);
 
+/** \see GEOSUnaryUnion */
+extern GEOSGeometry GEOS_DLL *GEOSUnaryUnionWithProgress_r(
+    GEOSContextHandle_t handle,
+    const GEOSGeometry* g,
+    GEOSProgressCallback_r progressFunc,
+    void* progressUserData);
+
 /** \see GEOSUnaryUnionPrec */
 extern GEOSGeometry GEOS_DLL *GEOSUnaryUnionPrec_r(
     GEOSContextHandle_t handle,
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -117,6 +117,23 @@ typedef void (*GEOSMessageHandler)(GEOS_PRINTF_FORMAT const char *fmt, ...)
 */
 typedef void (*GEOSMessageHandler_r)(const char *message, void *userdata);
 
+
+/**
+ * A GEOS progression and cancel function.
+ *
+ * Such function takes a progression ratio, and an optional
+ * message, and returns true if computation must continue, or false if it
+ * must be interrupted.
+ *
+ * \param progressRatio Progression ratio (between 0 and 1)
+ * \param message Information message (can be NULL)
+ * \param userdata the user data pointer that was passed to GEOS together with
+ * this callback.
+ * \return TRUE if processing must continue, or FALSE to interrupt it.
+ */
+typedef int (*GEOSProgressCallback_r)(double progressRatio, const char* message, void* userdata);
+
+
 /*
 * When we're included by geos_c.cpp, these types are #defined to the
 * C++ definitions via preprocessor. We don't touch them to allow the
@@ -885,6 +902,16 @@ GEOSCoverageSimplifyVW_r(
     const GEOSGeometry* input,
     double tolerance,
     int preserveBoundary);
+
+/** \see GEOSCoverageSimplifyVWWithProgress */
+extern GEOSGeometry GEOS_DLL *
+GEOSCoverageSimplifyVWWithProgress_r(
+    GEOSContextHandle_t extHandle,
+    const GEOSGeometry* input,
+    double tolerance,
+    int preserveBoundary,
+    GEOSProgressCallback_r progressFunc,
+    void* progressUserData);
 
 /** \see GEOSCoverageCleanParams_create */
 extern GEOSCoverageCleanParams GEOS_DLL *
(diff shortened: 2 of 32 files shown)


dbaston:
We already have a mechanism for requesting/checking interrupts, so I think a progress callback should handle progress reporting only.

I'm not sure about adding new C API signatures with a callback vs separately setting a progress callback in the way we do for interrupt handling. In part that may depend on what other operations would benefit from a progress callback, and how easy it is to pass the callback to the portion of the code that would invoke it. It seems fairly straightforward to pass it through the call chain here but I think other functions such as `GEOSUnaryUnion` would be more challenging.

@pramsey am I right that PostGIS has no way to take advantage of progress reporting, so it would largely come down to what is useful for GDAL / QGIS / others?

rouault:
> I'm not sure about adding new C API signatures with a callback vs separately setting a progress callback

I thought a bit about that, but that seems a bit messy to me. Passing all along the call chain is admittedly a bit more tedious but there is less ambiguity about which APIs are able to notify, and that could avoid issues if compositing progress of different subparts of the API was needed that would otherwise individually compete for a GEOS-wide progress callback attached to the GEOS context handle.

I'm fine removing the interrupt part of the callback if preferred.

rouault:
> does this provide a reasonable progress estimate for different datasets / parameters

I've just improved that with a new commit that adds sub-progress for the 2 createEdges() calls done in TPVWSimplifier::simplify(). 

>  is this the signature we want to use? 

It is strongly inspired from GDALProgressFunc. But I believe we should add back the boolean flag to cancel. GEOS Interrupt logic is quite inappropriate for multithreaded clients like GDAL since it is a global one (not per context). It is much more practical to hand over the cancel part to the progress function

dbaston:
> is this the signature we want to use?

Here I'm referring not to the signature of the callback itself, but whether we want functions to accept a progress callback or if we want to instead set a per-context progress handler. I can see that you are able to pass the progress handler to the spot where it needs to be called in this case, but I'm not confident that it's practical to do so for other algorithms that would benefit from progress reporting.

> GEOS Interrupt logic is quite inappropriate for multithreaded clients like GDAL since it is a global one (not per context). 

Some discussion in https://github.com/libgeos/geos/pull/761

> It is much more practical to hand over the cancel part to the progress function

If starting from scratch, yes, but we don't want to have one function that is using a different interrupt mechanism than the rest of GEOS.

rouault:
> but I'm not confident that it's practical to do so for other algorithms that would benefit from progress reporting.

what are you thinking to specifically ? I'm really skeptical about the per-context model for progress callback. In my refinement of the progression for CoverageSimplify I've used the trick of created scaled sub-progress callbacks like GDAL does, so if function A receives a progress function and function A calls B and C, then it passes a scaled progress for B from 0 to let's say 30% and a scale progress for C from 30% to 100%, and then B and C can do the same trick if they need. With a progression callback registered per-context I don't see how you would do that.

dr-jts:
>  In my refinement of the progression for CoverageSimplify I've used the trick of created scaled sub-progress callbacks like GDAL does

Right, that would have to be done directly by each algorithm, using essentially the same strategy (i.e. deciding how much of the total time each algorithm phase takes, and then pro-rating reported time based on that).

dr-jts:
I'm reluctant to introduce performance monitoring args all the way down the call stack in each algorithm.  Having a thread-local context I think would avoid that?

dbaston:
> what are you thinking to specifically ?

I don't have an alternate solution, but I'm hesitant to hurry and merge this without seeing how the same approach works in other functions, or understanding if/how other GEOS users would take advantage of it. I apologize if I was unclear in https://github.com/OSGeo/gdal/pull/12483#issuecomment-2919750442, but I was trying to convey some hesitation about the idea of progress reporting in general. I'm by no means opposed to it but I think this PR demonstrates that it's a tricky problem.

rouault:
> how the same approach works in other functions

Are there other GEOS functions in top of readers' heads where a single invokation of them is done typically on a huge object, or take a long time and that would be worth investigating ?

dr-jts:
> other GEOS functions in top of readers' heads where a single invokation of them is done typically on a huge object, or take a long time 

Here's a (complete?) list.  Some of them (e.g. coverage functions, `GEOSPolygonize`) are designed to be run on full datasets.  Others have an API which accepts single geometries as input, but are often run on large datasets by aggregating a set of geometries into a single collection (e.g. Convex Hull, Delaunay Triangulation)

```
GEOSConvsxHull
GEOSConcaveHull
GEOSDelaunayTriangulation
GEOSConstrainedDelaunayTriangulation
GEOSVoronoiDiagram
GEOSNode
GEOSPolygonize
GEOSTopologyPreserveSimplify

GEOSCoverageUnion
GEOSCoverageIsValid
GEOSCoverageClean et al
GEOSCoverageSimplifyVW

GEOSClusterDBSCAN
GEOSClusterGeometryDistance
GEOSClusterGeometryIntersects
GEOSClusterEnvelopeDistance
GEOSClusterEnvelopeIntersects
```

The clustering functions are usally run on full datasets too.

dbaston:
c3d5a9ed4248b2e3c0774224a24740b034514932 reduces the null checks and simplifies the creation of subprogress callbacks a bit.

dbaston:
I'm curious what people think about an API like:

```
GEOSSetProgressCallback(pFn, pUserData)
GEOSCoverageSimplifyVW(geom, tol, preserve)
```

The C API would store the callback in a variable to be used during the next function invocation and then cleared. If that function supports progress monitoring, it will be provided with the callback. Otherwise, the callback will be called with 1.0 (100%) after the operation is completed.

It's not super elegant, and I'm not sure it's a good idea, but I'm trying to think about whether we can avoid introducing two more signatures for each function that supports progress monitoring, and client code having to have a ton of preprocessor GEOS version checks.

pramsey:
I am fine with it, it seems no more or less aesthetically displeasing than the alternative. I assume the variable being set will be on the context, so hopefully this translates into multi-threaded situations safely.

rouault:
> I am fine with it, it seems no more or less aesthetically displeasing than the alternative

I'm also fine with it, pending:
- the documentation of the function where it is actually  implemented mentions it (so people don't have false expectations)
- and that the progress callback is really cleared in all situations after the GEOS function call, including when it exists following a GEOS C++ exception. It is important because otherwise if the progress callback remained installed and was later called in another GEOS function, the progress callback might be referencing objects (its user data) that are no longer alive.

It would be nice that GEOS after/before invoking the progress also checks for the interruption, because in a GDAL context, the GDAL progress function could typically ask GEOS for interruption.

pramsey:
I like it.

pramsey:
@dbaston you have thumbs up from me and @rouault, maybe @nyalldawson has an opinion as a prospective user of this API?

dbaston:
Updated branches tapping into GEOS progress from `gdal vector simplify-coverage` and `gdal vector dissolve`:
- GEOS: https://github.com/dbaston/libgeos/tree/context-progress-callback
- GDAL: https://github.com/dbaston/gdal/tree/simplify-coverage-progress

dbaston:
Continued in #1466

PR #1269, opened 2025-06-04
CMake: Create targets file in build tree

dbaston:
Analog of https://github.com/OSGeo/gdal/pull/12217

Lets one build GDAL (or anything else) by setting `GEOS_DIR` to a build tree.


dbaston:
To be clear, this is distinct from the targets file that would be generated at install time (and has been present in GEOS for several releases). I will add a note to NEWS.

rouault (approved):

PR #1271, opened 2025-06-04
UnaryUnionOp::Union(): fix wrong comment

rouault:
Not totally sure, but it seems to me the current comment is wrong (and if it is not, something should be better explained :-))


dr-jts:
Fixed in JTS: https://github.com/locationtech/jts/commit/ba30c6eaab31f97a72ab8f6484adb19543d5eb65

PR #1279, opened 2025-07-08
Port updates to BoundaryChainNoder

pramsey:
From https://github.com/locationtech/jts/pull/1134

### DIFF
--- a/include/geos/noding/BoundaryChainNoder.h
+++ b/include/geos/noding/BoundaryChainNoder.h
@@ -17,6 +17,7 @@
 #include <geos/noding/Noder.h>     // for composition
 #include <geos/noding/SegmentString.h>     // for composition
 #include <geos/geom/LineSegment.h> // for composition
+#include <geos/noding/BasicSegmentString.h>
 
 #include <unordered_set>
 
@@ -57,7 +58,7 @@ class GEOS_DLL BoundaryChainNoder : public Noder {
 
 private:
 
-    class BoundarySegmentMap {
+    class BoundaryChainMap {
 
         private:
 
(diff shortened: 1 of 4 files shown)


tests/unit/capi/GEOSCoverageUnionTest.cpp
```diff
@@ -66,24 +66,22 @@ template<>
...
+    // Temporary, wrap in a try/catch block until JTS upstream issue is fixed.
+    try {
+        auto result = GEOSCoverageUnion(input);
+        ensure( result != nullptr );
+        ensure( GEOSEquals(input, result) );
+        GEOSGeom_destroy(result);
+    }
+    catch(std::exception& e) {
+        (void)0;
+    }
```
jorisvandenbossche, line 82: @pramsey in shapely, we had the same test for overlapping inputs for coverage union, and so this also started to fail now. 
From your comment here, it seems that you are still planning to fix this to get back the previous behaviour of returning the input, but reading the discussion in https://github.com/locationtech/jts/pull/1134, that might not actually be the case?

The doc of GEOSCoverageUnion does mention _"It may generate an error (return NULL) for inputs that do not satisfy this constraint, however this is not guaranteed."_, so we can also add that to the shapely documentation that this can happen, and update our test to check for an error.

jorisvandenbossche: Before GEOS 3.12 (I assume before OverlayNG?), this case actually also already raised an error, but at the time the error message something about _"CoverageUnion cannot process incorrectly noded inputs"_

pramsey: I gave it a try in df6780c90.

dr-jts: > I gave it a try in [df6780c](https://github.com/libgeos/geos/commit/df6780c9064909f7a6825e3a798ce56da7394762).

That looks good to me. I'll make this change in JTS as well.

PR #1296, opened 2025-08-12
Add GEOSisSimpleDetail

dbaston:
As discussed at https://github.com/OSGeo/gdal/pull/12880#discussion_r2270081590

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -97,6 +97,7 @@
 #include <geos/operation/sharedpaths/SharedPathsOp.h>
 #include <geos/operation/union/CascadedPolygonUnion.h>
 #include <geos/operation/union/DisjointSubsetUnion.h>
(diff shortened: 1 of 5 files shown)


capi/geos_ts_c.cpp
```diff
@@ -1242,6 +1243,31 @@ extern "C" {
...
+    char
+    GEOSisSimpleDetail_r(GEOSContextHandle_t extHandle, const Geometry* g1, int returnAllPoints, Geometry** result)
+    {
+        return execute(extHandle, 2, [&]() {
+            geos::operation::valid::IsSimpleOp iso(g1);
+            iso.setFindAllLocations(returnAllPoints);
+
+            if (result != nullptr) {
+                *result = nullptr;
+            }
```
dbaston, line 1255:
```suggestion
            *result = nullptr;
```
No reason to call this with `result = nullptr`

PR #1305, opened 2025-09-15
geosop: make floating-point exceptions optional

fidomax:
commit 26292ce ("geosop: show most floating-point exceptions in verbose mode") add fenv.h dependency. Some libc implementations (uclibc-ng )may have no fenv support. So make floating-point exceptions optional and detetct fenv.h at configure stage.


mwtoews:
Thanks @fidomax this is also backported for 3.14.1

PR #1307, opened 2025-09-17
Make `GeometryFactory::addRef/dropRef` thread-safe

### LINKED ISSUE #847: Thread safety issue on GeometryFactory (C++ API)
When building geometry with GeometryFactory and then using the geometry in another thread for processing operations that may produce new geometry (e.g. Buffer, Union, etc), the _refCount variable is not thread-safe and, as such, GeometryFactory never gets destroyed in its destructor.

Rather than copying the geometry and recreating it again in a new thread with a new factory, it would be more efficient and easier to be able to work on the original geometry in a seperate thread. This all works fine, however the _refCount variable goes haywire and can end up in negative numbers or generally completely out of sync. 

Three suggestions

1) Add a std::mutex lock/guard to the addRef() and dropRef() methods on GeometryFactory.
2) Or.. change addRef(), dropRef() to protected class methods so users can subclass GeometryFactory and implement their own reference strategy.
3) ... alternatively, make it clear in the docs that GeometryFactory is not thread-safe and objects created using the factory should not be passed to different threads

ChrisThrasher:
Closes #847 

Even when using the reentrant C API, there are still potential thread safety issues when incrementing and decrementing `GeometryFactory::_refCount`. To fix this, I chose to do what `std::shared_ptr` does and make the reference count atomic.

### DIFF
--- a/include/geos/geom/GeometryFactory.h
+++ b/include/geos/geom/GeometryFactory.h
@@ -33,6 +33,7 @@
 #include <vector>
 #include <memory>
 #include <cassert>
+#include <atomic>
 
 namespace geos {
 namespace geom {
@@ -522,7 +523,7 @@ private:
     PrecisionModel precisionModel;
     int SRID;
 
-    mutable int _refCount;
+    mutable std::atomic<int> _refCount;
     bool _autoDestroy;
 
     friend class Geometry;


pramsey:
That's a wonderfully small change. Any anticipated issues? Seem to have been around since C++11, so that's great.

ChrisThrasher:
> Any anticipated issues?

Not that I can think of, but I'll admit I have no familiarity with the inner workings of this library. I've been indirectly using it for a few months but only in the last few days started poking around the source code when ThreadSanitizer started reporting issues.

My belief is that making this counter atomic will have no negative side effects except for a (likely immeasurable) performance penalty. The upside is that we can be guaranteed this value will never act erroneously in the presence of concurrent reads/writes.

PR #1312, opened 2025-10-10
CoverageRingEdges.cpp: Fix include map

Dave-Allured:
#include \<map\> was accidentally deleted in the previous commit, but it is referenced in several places within.  This commit restores the now-missing include.


mwtoews:
xref 79390691cfe0d428ab75bf10448b9cb82696e917 and #1302

PR #1329, opened 2025-11-19
SegmentString: Store coordinates as `std::shared_ptr<const CoordinateSequence>`

dbaston:
`SegmentString` coordinate handling (current)
===

All SegmentStrings carry a `CoordinateSequence` pointer. The ownership of this is pointer is determined by the `SegmentString` implementation; for example, `BasicSegmentString` does not assume ownership while `NodedSegmentString` does. This is not obvious in the interface and can result in either an unexpected class being used (`NodedSegmentString` when we are not performing noding but want ownership) or cumbersome ownership of `CoordinateSequences` in a side array. The `CoordinateSequence` is also not held as `const`, and although most `Noder` implementations do not modify the input coordinates, so do, so when constructing a `SegmentString` it is necessary to either defensively copy the coordinates, or `const_cast` and hope for the best.

Proposed changes
====

This PR attempts to address both ownership and `const` confusion by having `SegmentString`s hold coordinates as `std::shared_ptr<const CoordinateSequence>`. For a `Noder` that modifies the coordinates, a `SetCoordinates` method allows replacing the sequence with a modified version, without affecting other objects that may reference to same sequence. To avoid copies when creating `SegmentString`s, `LineString` (`SimpleCurve`) is modified to also hold its coordinates as `std::shared_ptr<const CoordinateSequence>`. This also allows constructs like `std::shared_ptr<const CoordinateSequence> removeRepeatedPoints(const std::shared_ptr<const CoordinateSequence>&)`, where the function output may be the same as the input, avoiding a copy if there were no repeated points to remove.

Related issues/PRs
====
#666

Performance
====

I've done a small amount of testing, with mixed results. The coverage union benchmark is probably a worst case because it creates lots of small polygons. It's maybe 5-10% slower on Linux/gcc, 1-2% slower on macOS/clang. A mixed intersects/intersection benchmark is 2-3% slower on Linux, 0-1% slower on macOS/clang.  A unary union benchmark is 2-3% faster on Linux, no difference on macOS/clang.

C API
====

SegmentStrings and Noders are fortunately not exposed to the C API. There is a potential impact in changing the storage of `LineString` to `std::shared_ptr<const CoordinateSequence>`.

```
GEOSCoordSequence* seq = GEOSGeom_getCoordSeq(geom);
GEOSNormalize(geom);
```

`GEOSNormalize` is one of the few (only?) functions that modifies a geometry in-place. Without this PR, the coordinates of the geometry are modified in-place, so `seq` is still valid after a call to `GEOSNormalize`. With this PR, if another geometry shares coordinates with `geom`, `GEOSNormalize` will replace the coordinates of `geom`, in which case `seq` would (unexpectedly?) point to the un-normalized coordinates. (Although not implemented here, I can imagine `LineString::clone` not actually copying the coordinates, which would produce this behavior.)


pramsey (approved):
In combination with the changers around Noder hopefully this simplifies a bunch of code .

PR #1345, opened 2025-12-16
DistanceOp::distance(): make it return +inf instead of 0 if one of the geometry is empty

rouault:
Fixes https://github.com/OSGeo/gdal/issues/12978


rouault:
Changed to returning +Inf

Issue #1206, opened 2024-12-07
Unit test unit-geom-Envelope fails on FreeBSD 14.1


diizzyy:
Hi,

One unit test fails on FreeBSD

```
183/494 Testing: unit-geom-Envelope
183/494 Test: unit-geom-Envelope
Command: "/usr/ports/graphics/geos/work/.build/bin/test_geos_unit" "geos::geom::Envelope"
Directory: /usr/ports/graphics/geos/work/.build/tests/unit
"unit-geom-Envelope" start time: Dec 07 12:21 CET
Output:
----------------------------------------------------------
===============================
  GEOS Unit Test Suite
===============================

geos::geom::Envelope: .....[6=F].......[14=F]......

---> group: geos::geom::Envelope, test: test<6>
     problem: assertion failed
     failed assertion: `FE_INVALID raised`

---> group: geos::geom::Envelope, test: test<14>
     problem: assertion failed
     failed assertion: `FE_INVALID raised`

tests summary: failures:2 ok:18
<end of output>
Test time =   0.02 sec
----------------------------------------------------------
Test Failed.
"unit-geom-Envelope" end time: Dec 07 12:21 CET
"unit-geom-Envelope" time elapsed: 00:00:00
----------------------------------------------------------
```

OS: FreeBSD 14.1 (amd64)
Version: 3.13.0 and latest main ([dd0fdc5](https://github.com/libgeos/geos/commit/dd0fdc59b9fd9aa7d0bc4eda9c16ff968e91815a)
Compiler: Clang 18.1.5
CMake: 3.31.0

diizzyy:
Friendly ping

pramsey:
Patches accepted! These things are obscure and since I do not have a FreeBSD system lying around to test it on, it is best to wait for someone with the system in question to fix it.

diizzyy:
I'm going to guess it's not a FreeBSD thing but rather clang(llvm) vs gcc, unfortunately I don't have a Linux box to verify on.

pramsey:
We test [many versions](https://github.com/libgeos/geos/actions/runs/13660897008) of gcc and clang on Linux in our CI and aren't seeing this.

robe2:
It's showing on our FreeBSD ci - -

FreeBSD bessie 14.1-RELEASE FreeBSD 14.1-RELEASE releng/14.1-n267679-10e31f0946d8 GENERIC amd64

lbartoletti:
[...]

PR #1374, opened 2026-01-14
fix(BSD): unit-geom-Envelope fails due to fp exception

lbartoletti:
It's not only a FreeBSD problem, confirmed by  @landryb on OpenBSD and maybe reproducible on other systems with clang.

Fixes https://github.com/libgeos/geos/issues/1206
Patch from https://cgit.freebsd.org/ports/commit/graphics/geos/files/patch-CMakeLists.txt?id=53d2b997aba6f272d3f78e504e3c1911a6ff5b15 thanks to @diizzyy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,11 +214,17 @@ target_compile_features(geos_cxx_flags INTERFACE cxx_std_17)
 #   https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98207
 #-----------------------------------------------------------------------------
 
+set(GEOS_USE_STRICT_FP OFF)
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT EMSCRIPTEN)
+    set(GEOS_USE_STRICT_FP ON)
+endif()
+
 target_compile_options(geos_cxx_flags INTERFACE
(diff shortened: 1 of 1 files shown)


lbartoletti:
> Does this patch have a performance impact? It doesn't seem to be needed on most systems using clang.

From [clang manual](https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-model):

`strict Enables -frounding-math and -ffp-exception-behavior=strict, and disables contractions (FMA). All of the -ffast-math enablements are disabled. Enables STDC FENV_ACCESS: by default FENV_ACCESS is disabled. This option setting behaves as though #pragma STDC FENV_ACCESS ON appeared at the top of the source file.`

So, fast-math is disabled, but is it used here?

Do you know how we can benchmark this? I can try on several machines with different systems.

> I'm also unclear on the practical consequence of `FE_INVALID` being raised if the operation still produces correct results.

I’ve never been able to reproduce this error outside GEOS. My first idea was to isolate the unit test on FreeBSD, but since it is also present on OpenBSD (where we have differences), I think it is related to the compiler and possibly libm. So it may also be present elsewhere (although it has not been observed so far).

But, I can understand that you want to disable this specific test with a note about the problem.

Let me know what you prefer, and I’ll update it accordingly.

diizzyy:
For what it's worth, this is also an issue with jpeg-turbo (see also link to upstream issue report)
https://github.com/freebsd/freebsd-ports/blob/main/graphics/jpeg-turbo/Makefile#L28

PR #1375, opened 2026-01-14
test(bsd): disable FE_INVALID check on EnvelopeTest

lbartoletti:
Supersede https://github.com/libgeos/geos/pull/1374
Fixes https://github.com/libgeos/geos/issues/1206


mwtoews (approved):
Using [anyvm](https://github.com/anyvm-org/anyvm) locally with FreeBSD, I can see the issue and can verify that this simple fix resolves the issue (via skip). I might also look into adding [freebsd-vm](https://github.com/marketplace/actions/freebsd-vm) for additional CI coverage.

PR #1424, opened 2026-04-09
Add GeometrySplitter

dbaston:
This PR adds a GeometrySplitter class for splitting linear or polygonal features using points, lines, and/or polygons.

Curved geometries are not yet supported. A couple of pieces are missing for this:
- to split a curve with a point, we need to find the closest section of the curve to that point. This requires
  adding curve support to DistanceOp. 
- ~~to split a CurvePolygon with a line/curve, we need to figure out which of the resulting polygons are inside
  the original geometry. This is done with PIP tests, but the existing InteriorPointArea algorithm does not 
  support curves. Either InteriorPointArea could be extended for this, or we could compute interior points
  by testing either (pseudo)random points or computing small offsets from segments/arcs.~~ EDIT: added RandomPointsBuilder from JTS to handle this.

I've added applicable test cases from PostGIS and QGIS, most of which are passing.

Currently not passing:

- tests relying on missing curve support described above
- splitting a self-intersecting line with another line causes splits at the self-intersection points.
  PostGIS does the same but I understand why the QGIS behavior is desirable.
- a split of a nearly-collapsed polygon produces a different result from QGIS and PostGIS. I'm not sure
  the results from that software are desirable, though.

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -721,6 +721,12 @@ extern "C" {
     }
 
     Geometry*
+    GEOSSplit(const Geometry* g, const Geometry* edge)
+    {
+        return GEOSSplit_r(handle, g, edge);
+    }
+
+    Geometry*
     GEOSUnionCascaded(const Geometry* g)
     {
         return GEOSUnionCascaded_r(handle, g);
--- a/capi/geos_c.h.in
+++ b/capi/geos_c.h.in
@@ -1168,6 +1168,12 @@ extern GEOSGeometry GEOS_DLL *GEOSNode_r(
     GEOSContextHandle_t handle,
     const GEOSGeometry* g);
 
+/** \see GEOSSplit */
+extern GEOSGeometry GEOS_DLL *GEOSSplit_r(
+    GEOSContextHandle_t handle,
+    const GEOSGeometry* g,
+    const GEOSGeometry* edge);
+
 /** \see GEOSClipByRect */
 extern GEOSGeometry GEOS_DLL *GEOSClipByRect_r(
     GEOSContextHandle_t handle,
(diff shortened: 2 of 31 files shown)


pramsey:
Clearly we just need to move PostGIS and QGIS to the GEOS implementation ASAP and resolve the inconsistency problem 😄

nyalldawson:
Thank you!! I would LOVE to drop QGIS' fragile and inefficient version of this! 🥳🥳

m-kuhn:
This is very good news for everyone !

jorisvandenbossche:
@dbaston I am testing this with Shapely, by using the new C API instead of the python homegrown splitting logic in shapely, and running the tests, and at the moment only noticing one issue / difference.

For the case of splitting a line with a line, when both lines overlap (so not just cross), Shapely at the moment detects this and raises an error disallowing such input. 

With GEOS, the split function works:

```
$ geosop split -a 'LINESTRING (0 0, 1.5 1.5, 3 4)' -b 'LINESTRING (0 0, 15 15)'
GEOMETRYCOLLECTION (LINESTRING (0 0, 1.5 1.5), LINESTRING (1.5 1.5, 3 4))
```

I don't have a strong opinion about what is the best behaviour (the above output seems the logical result if you want to have it just work), but just raising this to be sure this was a conscious design decision (or might warrant a test).

dbaston:
Thanks @jorisvandenbossche , I should have listed this difference this in the PR description but forgot to.

I don't know if either the GEOS or PostGIS behavior is a conscious design decision so much as a natural consequence of the implementation. PostGIS is taking advantage of a `GEOSDifference` behavior to perform the split, but that only produces the desired output if the intersection of the two inputs does not contain any lines. The GEOS implementation is not based on `GEOSDifference` and does not have this limitation, and I didn't see a reason to explicitly introduce it.

> or might warrant a test

There are in fact a couple:

https://github.com/libgeos/geos/blob/7ce71942d882ca8ed301181d03d4b31177c340d0/tests/unit/operation/split/GeometrySplitterTest.cpp#L371-L393

Issue #509, opened 2021-11-19
[arm64] libGeos performance degraded due to calls to soft float operations


sebpop:
Running libGeos performance benchmarks show that there is an issue on arm64.
GCC generates calls to the software operation in libgcc instead of using Arm64's FP instruction.

Here are the steps to reproduce the performance issue:
```
git clone https://git.osgeo.org/gitea/geos/geos.git 
mkdir geos/build
cd geos/build
cmake -DBUILD_BENCHMARKS:BOOL=ON ..
make -j10
perf record ./bin/perf_voronoi
perf report
```
On arm64:
```
  55.32%  perf_voronoi  libgcc_s.so.1        [.] __multf3
   9.98%  perf_voronoi  libgcc_s.so.1        [.] __addtf3
   6.88%  perf_voronoi  libgcc_s.so.1        [.] __subtf3
   4.96%  perf_voronoi  libgcc_s.so.1        [.] 0x000000000000689c
   2.32%  perf_voronoi  libgeos.so.3.11.0    [.] geos::triangulate::IncrementalDelaunayTriangulator::insertSite
[...]
```
On x64:
```
   9.29%  perf_voronoi  libgeos.so.3.11.0    [.] geos::triangulate::quadedge::TrianglePredicate::isInCircleNormalized
   8.39%  perf_voronoi  libgeos.so.3.11.0    [.] geos::triangulate::IncrementalDelaunayTriangulator::insertSite
   8.27%  perf_voronoi  libgeos.so.3.11.0    [.] geos::triangulate::quadedge::QuadEdgeSubdivision::fetchTriangleToVisit
[...]
```
A reduced testcase is:
```
long double
foo (long double x)
{
  return x * x;
}
```
On arm64:
```
aarch64 - 16:57:59:~/bugs/long-double$  gcc -O2 -S -o- a.c
foo:
	stp	x29, x30, [sp, -16]!
	mov	v1.16b, v0.16b
	mov	x29, sp
	bl	__multf3
	ldp	x29, x30, [sp], 16
	ret
```
On x64:
```
x86_64 - 16:58:05:~$ gcc -O2 -S -o- a.c 
foo:
	endbr64
	fldt	8(%rsp)
	fmul	%st(0), %st
	ret
```
I will be taking the reduced bug to gcc's bugzilla as well.

pramsey:
[...]

PR #1429, opened 2026-04-27
Fix GEOS benchmarking on Windows

MugundanMCW:
PR Description:
* This PR fixes compilation issues that prevented building benchmarks on Windows with the MSVC compiler.
 
Changes Proposed:
* Link perf_orientation target against geos library instead of geos_cxx_flags. The previous setup only linked compiler flags, causing linker errors due to missing CoordinateXY::getNull() function.
* Replace POSIX random() with standard C rand(), since MSVC does not provide random().
* Add GEOS_DLL export to IntervalRTreeLeafNode and SortedPackedIntervalRTree classes to ensure these symbols are exported in geos.dll on Windows. This fixes linker errors when building perf_spatial_index target.

### DIFF
--- a/benchmarks/algorithm/CMakeLists.txt
+++ b/benchmarks/algorithm/CMakeLists.txt
@@ -29,7 +29,7 @@ if (benchmark_FOUND)
             $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
             $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>)
     target_link_libraries(perf_orientation PRIVATE
-            benchmark::benchmark geos_cxx_flags)
+            benchmark::benchmark geos)
 
     add_executable(perf_line_intersector LineIntersectorPerfTest.cpp)
     target_include_directories(perf_line_intersector PUBLIC
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,10 +1,64 @@
-## Benchmarks
+## Building
 
-A variety of programs to execute various kinds of tests, 
-including benchmarks, stability and robustness tests.
+Building the benchmark tests must be enabled using:
 
-### Building
+```bash
+cmake -DBUILD_BENCHMARKS=ON ..
+```
 
-Building the benchmark tests must be enabled using 
+Geos uses Google Benchmarks for validating performance of Geos library. It can be built using the following ways:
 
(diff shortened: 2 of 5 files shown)


benchmarks/algorithm/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if (benchmark_FOUND)
             $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
             $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>)
     target_link_libraries(perf_orientation PRIVATE
-            benchmark::benchmark geos_cxx_flags)
+            benchmark::benchmark geos)
```
mwtoews, line 32: This is a fix for Linux too, which currently fails to compile on `main`.

mwtoews (approved):
Thanks! The benchmarks testing isn't really done in CI, and gets less attention. Checking on Linux it wasn't working before this PR, so it's a overall a good improvement.

Were you able to run benchmarks on Windows? Do you have any tips or comments? It could be worthwhile adding to [benchmarks/README.md](https://github.com/libgeos/geos/blob/main/benchmarks/README.md). (It's currently missing for Linux, but this can be added to look similar to https://github.com/libgeos/geos/issues/509)

MugundanMCW:
Hi @mwtoews,

Thanks for your comments. Yes, with the changes proposed in this PR, I was able to build and run the benchmarks successfully on Windows systems. Regarding tips or hints for running the benchmarks, I can add some instructions to benchmarks/README.md if that would be helpful.

mwtoews:
Thanks @MugundanMCW the instructions guidance is useful! Appreciate the fix and guidance enhancements.

MugundanMCW:
Hi @mwtoews,
GEOS is currently a C++ library and, as far as I know, does not include architecture-specific optimizations (such as SIMD intrinsics) to further accelerate computations. Are there any plans to introduce architecture-specific optimizations to improve the library’s performance?

Thanks!

dbaston:
There aren't any plans to, but I don't think the project necessarily opposes it. I wrote a SIMD version of the STRtree a few years ago and concluded that the benefit didn't justify the extra code. But maybe a better implementation could be written.

MugundanMCW:
Thanks @dbaston

PR #1445, opened 2026-06-12
Node a set of LineString and retain reference to original inputs

pramsey:
This feature is in support of the idea of an [ST_NodeWin()](https://trac.osgeo.org/postgis/ticket/5377) PostGIS feature, that would allow a table of LineString to be noded while retaining the associated attributes. Would look like 
```
SELECT ST_NodeWin(geom) AS noded_geom, foo, bar, baz 
FROM mygeomtable
```
Along the same lines as the ST_CoverageSimplify() implementation, providing 1:1 input/output guarantee.

This code is from my friend Claude. This PR is to make the changes reviewable by me (human) and also to get any other feedback that can improve this work.

### DIFF
--- a/capi/geos_c.cpp
+++ b/capi/geos_c.cpp
@@ -790,6 +790,12 @@ extern "C" {
     }
 
     Geometry*
+    GEOSNodeCollection(const Geometry* input, double gridSize)
+    {
+        return GEOSNodeCollection_r(handle, input, gridSize);
+    }
+
+    Geometry*
     GEOSSplit(const Geometry* g, const Geometry* edge)
     {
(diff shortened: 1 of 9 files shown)


dbaston:
Is there a particular barrier to integrating this into the existing `GeometryNoder` ? Or, putting the inputs into `OverlayNG` and extracting the noded edges?

pramsey:
@dbaston have a look at #1460 as a potential replacement, using GeometryNoder. Only kind of ugly thing, I do want to keep an option to snap inputs, but that means that curved geometries end up disabling the snapping behaviour when they are present.

PR #1453, opened 2026-06-18
KdTree: Disable optimizations in latest MSVC version

### LINKED ISSUE #1449: Unit tests segfault with latest MSVC
CI passing with MSVC 19.44.35227.0: https://github.com/libgeos/geos/actions/runs/26833807466/job/79122127162
CI failing with MSVC 19.51.36246.0: https://github.com/libgeos/geos/actions/runs/27260286741/job/81082076515

Only commits between the two are:

9d1676ca58c6878445b679d6b3228b96fab7d4e4, Bump the all-actions group across 1 directory with 4 updates 
7b2e578f4444678aa081af6c484266ed2f6392bc, Add Arc distance test case from PostGIS

dbaston:
Workaround for #1449

### DIFF
--- a/src/index/kdtree/KdTree.cpp
+++ b/src/index/kdtree/KdTree.cpp
@@ -226,6 +226,11 @@ KdTree::queryNode(KdNode* currentNode, const geom::Envelope& queryEnv, bool odd,
     }
 }
 
+#if defined(_MSC_VER) && _MSC_FULL_VER > 195035734
+// Avoid segfault with MSVC 19.51.36246.0
+// See https://github.com/libgeos/geos/issues/1449
+#pragma optimize("", off)
+#endif
 /*private*/
 KdNode*
 KdTree::queryNodePoint(KdNode* currentNode, const geom::Coordinate& queryPt, bool odd)
@@ -257,7 +262,9 @@ KdTree::queryNodePoint(KdNode* currentNode, const geom::Coordinate& queryPt, boo
     }
     return nullptr;
 }
-
+#if defined(_MSC_VER) && _MSC_FULL_VER > 195035734
+#pragma optimize("", on)
+#endif
 
 /*public*/
 void
--- a/tests/unit/index/kdtree/KdTreeTest.cpp
+++ b/tests/unit/index/kdtree/KdTreeTest.cpp
@@ -189,6 +189,24 @@ void object::test<8> ()
     ensure(node->isRepeated());
 }
 
+template<>
+template<>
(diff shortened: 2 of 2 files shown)


pramsey (approved):

rouault:
has anyone tried to report it to Microsoft as it smells now as a quite likely optimizer bug. Although the code in that function looks to me as super boring (not like math stuff were you could have autovectorizer bugs...)

src/index/kdtree/KdTree.cpp
```diff
@@ -226,6 +226,9 @@ KdTree::queryNode(KdNode* currentNode, const geom::Envelope& queryEnv, bool odd,
     }
 }
 
+#if defined(_MSC_VER) && _MSC_FULL_VER > 194435227
```
mwtoews: My local test with MSVC 19.50.35734.0 was good, so the compiler break could be narrowed.
```suggestion
#if defined(_MSC_VER) && _MSC_FULL_VER > 195035734
```

src/index/kdtree/KdTree.cpp
```diff
@@ -257,7 +260,9 @@ KdTree::queryNodePoint(KdNode* currentNode, const geom::Coordinate& queryPt, boo
     }
     return nullptr;
 }
-
+#if defined(_MSC_VER) && _MSC_FULL_VER > 194435227
```
mwtoews:
```suggestion
#if defined(_MSC_VER) && _MSC_FULL_VER > 195035734
```

mwtoews (commented):
It might also be a good idea to add a comment with a link to `https://github.com/libgeos/geos/issues/1449` as this bug will likely be resolved with a future compiler update and this logic will need to be revised.

src/index/kdtree/KdTree.cpp
dbaston: Adjusted, thanks!

dbaston:
> has anyone tried to report it to Microsoft

It's not clear to me where you would do so, do they really just track these kinds of these kinds of things in the Developer Community forum?

b0689cbf1ac4f21c9370e8485aae39c8a55d9c47 adds a minimal example to provoke the failure. It's as basic as a CS 101 example, I would think a compiler bug affecting this would be blowing up code all over the place. Bizarre.

rouault:
> do they really just track these kinds of these kinds of things in the Developer Community forum?

yes I believe that's the only place where us mortals can report and hope for the best...

PR #1455, opened 2026-06-23
OSS-Fuzz: Add new fuzzer targets geo operations

arthurscchan:
This PR adds an additional OSS-Fuzz fuzzer targeting geo operations on an geo object parsed from random input, together with some seed corpora for the new fuzzer.


pramsey (approved):
Looks good to me. Will be so "happy" to see everything it uncovers.

dbaston (approved):

PR #1459, opened 2026-06-25
OverlayNG: Merge linear outputs

dbaston:
As discussed in https://lists.osgeo.org/pipermail/geos-devel/2026-May/011167.html (May) and https://lists.osgeo.org/pipermail/geos-devel/2026-June/011179.html (June)

I have not modified the XML tests yet.


dbaston:
I've updated the XML tests to expect merged output, with one exception. `test_split.xml` intends to show how a side effect of `Geometry::difference` allows splitting a linear geometry with a second linear geometry. Since this PR removes that side effect, I removed the test file.

dbaston:
It appears PostGIS is relying on the splitting side-effect of `GEOSDifference`:

https://github.com/postgis/postgis/blob/423a5fd79c2e4d2a1fd88911ffbfb9f906d4a6a8/liblwgeom/lwgeom_geos_split.c#L115

So PostGIS would need to be updated to use `GEOSSplit` in this case instead.

tests/unit/operation/overlayng/CoverageUnionNGTest.cpp
```diff
@@ -181,27 +181,29 @@ void object::test<11> ()
...
+        "MULTILINESTRING ((1 9, 3.1 8, 5 7), (2 3, 4 3, 5 3), (5 3, 5 7), (5 3, 7 4, 9 5), (5 3, 8 1), (5 7, 7 8, 9 9))");
 }
 
-// Z values preserved in linear inpuuts
+// Z values preserved in linear inputs
 template<>
 template<>
 void object::test<12> ()
 {
+    // TODO: Should Z values be averaged?
```
pramsey, line 192: They are in overlay, so they should be in this operation too, I think is consistent. Probably better than unpredictable ordering based first-in / last-in selection.

tests/xmltester/tests/misc/split.xml
```diff
@@ -1,51 +0,0 @@
-<run>
-<!--
-This test is to verify difference() can be used to split linestrings
-and multilinestring using a "cutter" linesting.
--->
```
pramsey, line 5: Does it make sense to change (some of?) these into tests of the GEOSSplit functionality instead, so they are not lost as cases?

pramsey (commented):
Can we get a NEWS entry for this feature that notes changes in behaviour for downstream projects (what previously returned a MULTILINESTRING may now return a LINESTRING)?

tests/xmltester/tests/misc/split.xml
dbaston: Good idea, done in 77eeb1a6239fb1b2f940af3cef70fc99ac524134

tests/unit/operation/overlayng/CoverageUnionNGTest.cpp
dbaston: Agree that averaging is desirable (if the cost is not too great), but after giving the implementation some thought I think it's out-of-scope for this changeset. It's the job of the `Noder` to average Z values but the noders used in `CoverageUnion` deliberately avoid the intersection calculations that would produce averages. You can see the same winner-take-all behavior in the polygon `CoverageUnion`, which is unaffected by this PR:

```
geosop -a "GEOMETRYCOLLECTION (POLYGON Z ((0 0 0, 1 0 1, 0 1 2, 0 0 0)), POLYGON Z ((1 0 100, 1 1 200, 0 1 300, 1 0 100)))" coverageUnionNG
POLYGON Z ((1 0 1, 0 0 0, 0 1 2, 1 1 200, 1 0 100))
```

dbaston:
It looks like shapely also relies on the current behavior of `GEOSDifference` to split a line:

https://github.com/shapely/shapely/blob/e10c6da077744005290f77eaa725777d1c631f66/shapely/ops.py#L362

pramsey:
I have pushed a change to PostGIS to use GEOSSplit over GEOSDifference in the ST_Split function, from GEOS 3.15 and on.

PR #1460, opened 2026-06-26
GeometryNoder: Add collection noding with per-member identity

pramsey:
Extend GeometryNoder with a new nodeCollection() API that nodes a vector of geometries against each other while preserving 1:1 input/output identity: output element i is the noded form of input element i, as a MultiLineString (or MultiCurve for curved members).

This is distinct from the existing node() path, which dissolves all input into a single flat MultiLineString. Shared linework is retained in every member that touches it; point members yield empty MultiLineString slots; areal members contribute their boundary rings as noded linework, consistent with the existing single-geometry behaviour.

Snap-rounding (gridSize > 0) is supported for linear input. When any member is curved, gridSize is ignored and exact arc noding runs instead, so curve support is inherited from the existing SimpleNoder / ArcIntersectionAdder / CurveRebuilder machinery at no extra cost.

Expose the new functionality in the C API as GEOSNodeCollection(input, gridSize). A non-collection input is treated as a one-member collection.

### DIFF
--- a/capi/geos_ts_c.cpp
+++ b/capi/geos_ts_c.cpp
@@ -2043,6 +2043,26 @@ extern "C" {
     }
 
     Geometry*
+    GEOSNodeCollection_r(GEOSContextHandle_t extHandle, const Geometry* input, double gridSize)
+    {
+        return execute(extHandle, [&]() {
+            // Each member of the input collection is noded against the
+            // others; the result is a collection of the same size, member
+            // i being the noded form of input member i.
+            std::vector<const Geometry*> geoms(input->getNumGeometries());
+            for (std::size_t i = 0; i < geoms.size(); i++) {
+                geoms[i] = input->getGeometryN(i);
+            }
+
+            auto noded = geos::noding::GeometryNoder::nodeCollection(geoms, gridSize);
+
+            auto out = input->getFactory()->createGeometryCollection(std::move(noded));
+            out->setSRID(input->getSRID());
+            return out.release();
+        });
+    }
+
+    Geometry*
     GEOSSplit_r(GEOSContextHandle_t extHandle, const Geometry* g, const Geometry* edge)
     {
         return execute(extHandle, [&]() {
--- a/src/noding/GeometryNoder.cpp
+++ b/src/noding/GeometryNoder.cpp
@@ -37,6 +37,7 @@
 #include <geos/noding/OrientedCoordinateArray.h>
 #include <geos/noding/SegmentString.h>
 #include <geos/noding/SimpleNoder.h>
+#include <geos/noding/snapround/SnapRoundingNoder.h>
 
 #include <geos/util/Assert.h>
 
@@ -281,6 +282,14 @@ GeometryNoder::node(const geom::Geometry& geom1, const geom::Geometry& geom2)
     return noder.getNoded();
 }
 
+/* public static */
+std::vector<std::unique_ptr<geom::Geometry>>
+GeometryNoder::nodeCollection(const std::vector<const geom::Geometry*>& geoms, double gridSize)
+{
+    GeometryNoder noder(geoms, gridSize);
+    return noder.getNodedCollection();
+}
+
 /* public */
 GeometryNoder::GeometryNoder(const geom::Geometry& g)
     :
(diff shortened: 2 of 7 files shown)


capi/geos_ts_c.cpp
```diff
@@ -2042,6 +2042,26 @@ extern "C" {
...
+        return execute(extHandle, [&]() {
+            // Each member of the input collection is noded against the
+            // others; the result is a collection of the same size, member
+            // i being the noded form of input member i.
+            std::vector<const Geometry*> geoms(input->getNumGeometries());
+            for (std::size_t i = 0; i < geoms.size(); i++) {
+                geoms[i] = input->getGeometryN(i);
+            }
+
+            auto noded = geos::noding::GeometryNoder::nodeCollection(geoms, gridSize);
```
dbaston, line 2057: Does the conversion to a vector buy anything here? It seems like `nodeCollection` could take a collection as an argument instead of a vector. That would also clean up `GeometryNoder` by not having both `argGeom1` and `argColl` members, and would remove the need for the `collectionHasCurves` helper.

src/noding/GeometryNoder.cpp
```diff
@@ -426,11 +554,16 @@ GeometryNoder::getNoder()
     if(!noder) {
         const geom::PrecisionModel* pm = argGeom1->getFactory()->getPrecisionModel();
         if (argGeomHasCurves) {
+            // Snap-rounding cannot node arcs; curved input always uses
+            // exact arc noding and ignores any requested gridSize.
             noder = std::make_unique<SimpleNoder>();
```
dbaston, line 559: Maybe better to error instead of ignoring the parameter.

src/noding/GeometryNoder.cpp
```diff
@@ -426,11 +554,16 @@ GeometryNoder::getNoder()
...
 
             m_cai = std::make_unique<algorithm::CircularArcIntersector>(argGeom1->getPrecisionModel());
             m_aia = std::make_unique<ArcIntersectionAdder>(*m_cai);
             detail::down_cast<SimpleNoder*>(noder.get())->setArcIntersector(*m_aia);
+        } else if (m_gridSize > 0.0) {
+            m_pm = std::make_unique<geom::PrecisionModel>(1.0 / m_gridSize);
+            noder = std::make_unique<snapround::SnapRoundingNoder>(m_pm.get());
         } else {
             noder = std::make_unique<IteratedNoder>(pm);
```
dbaston, line 568: Should `GeometryNoder` use `SnapRoundingNoder` for the non-collection case as well?

tests/unit/noding/GeometryNoderTest.cpp
```diff
@@ -72,4 +74,366 @@ void object::test<3>()
...
+    ensure_equals("1:1 mapping", result.size(), 2u);
+
+    // The line slot splits at (0 1).
+    ensure_equals("line slot type", result[0]->getGeometryTypeId(), geos::geom::GEOS_MULTILINESTRING);
+    ensure_equals("line split count", result[0]->getNumGeometries(), 2u);
+
+    // The arc slot stays curved: a MultiCurve of two CircularStrings.
+    ensure_equals("arc slot type", result[1]->getGeometryTypeId(), geos::geom::GEOS_MULTICURVE);
+    ensure_equals("arc split count", result[1]->getNumGeometries(), 2u);
+    for (std::size_t i = 0; i < result[1]->getNumGeometries(); i++) {
```
dbaston, line 186: Loop from i = 0 to i = 1 seems unnecessary

capi/geos_ts_c.cpp
pramsey: CoverageSimplifier is the analogue I was looking at, which has 
```
std::vector<std::unique_ptr<Geometry>>
CoverageSimplifier::simplify(
    std::vector<const Geometry*>& coverage,
    double tolerance)
```
as its main entry point. On the flip in the C API the entry point is
```
    Geometry*
    GEOSCoverageSimplifyVW(
        const Geometry* input,
        double tolerance,
        int preserveBoundary)
```
so.... what would you prefer?

pramsey: `geom.hasCurvedComponents()` in place of `collectionHasCurves()` I presume

src/noding/GeometryNoder.cpp
pramsey: I don't understand, where does the logic switch on collection/singleton?

dbaston: I guess it doesn't, please disregard

capi/geos_ts_c.cpp
dbaston: No strong preference I guess, it would just be nice if there was some way to not have all of `argGeom1`, `argGeom2`, and `argColl` (whose name implies a collection but is not...)

PR #1470, opened 2026-07-07
geos_c.h: Make enums available with GEOS_USE_ONLY_R_API defined

twpayne:
The `GEOSHULL_PARAM_*` and `GEOS_VORONOI_*` enum values were only available when `GEOS_USE_ONLY_R_API` was *not* defined.

This PR moves the definitions outside the `#ifndef GEOS_USE_ONLY_R_API` block so that they are available even when `GEOS_USE_ONLY_R_API` is defined.

Credit to @nystromer for the investigation.
Refs https://github.com/twpayne/go-geos/pull/238.


dbaston:
Thanks!

Issue #1490, opened 2026-07-28
GEOSSplit with (Multi)Point input geometry


jorisvandenbossche:
Adopting the new split function in Shapely, I noticed the following inconsistency.

When splitting a (Multi)Point with linear edge, you get the expected error:

```
$ geosop -a "POINT (1 1)" -b "LINESTRING (0 0, 1 1)" split
Run-time exception: IllegalArgumentException: Input geometry must be linear or polygonal.
$ geosop -a "MULTIPOINT ((1 1), (1 2))" -b "LINESTRING (0 0, 1 1)" split
Run-time exception: IllegalArgumentException: Input geometry must be linear or polygonal.
```

but when splitting a (Multi)Point with another point-like geometry, it returns the original input geometry:

```
$ geosop -a "POINT (1 1)" -b "POINT (1 1)" split
POINT (1 1)
$ geosop -a "MULTIPOINT ((1 1), (1 2))" -b "POINT (1 1)" split
MULTIPOINT ((1 1), (1 2))
```

According to the docs this should not be supported, and given that the other case _does_ raise, I suppose this might be an oversight.
But opening this issue to verify that assumption. cc @dbaston

I was making a change to up-front raise an error on any point-dimension input in `GeometrySplitter::split`, but I was wondering if this might currently be supported to cover the case of mixed GeometryCollection input?
Personally, I would be fine with saying that splitting points is "supported" by essentially just passing through point-like geometries as-is, but then at least it should still return a GeometryCollection containing that point? (since the function is documented to always return a collection, and it is nice to be able to rely on that)

jorisvandenbossche:
>  was wondering if this might currently be supported to cover the case of mixed GeometryCollection input?

Actually, testing that right now, this seems to segfault specifically for the case of a point in the input collection and splitting with the point, so this is another reason to explicitly disallow that:

```
$ geosop -a "GEOMETRYCOLLECTION (LINESTRING (0 0, 1.5 1.5, 3 4), POINT (1 1))" -b "LINESTRING (0 0, 1 1)" split
Run-time exception: IllegalArgumentException: Input geometry must be linear or polygonal.
$ geosop -a "GEOMETRYCOLLECTION (LINESTRING (0 0, 1.5 1.5, 3 4), POINT (1 1))" -b "POINT (0 0)" split
Segmentation fault (core dumped)
[...]

PR #1491, opened 2026-07-28
GeometrySplitter: always raise error for point input geometry

jorisvandenbossche:
Potentially addresses https://github.com/libgeos/geos/issues/1490

### DIFF
--- a/src/operation/split/GeometrySplitter.cpp
+++ b/src/operation/split/GeometrySplitter.cpp
@@ -173,6 +173,10 @@ GeometrySplitter::splitAtPoints(const Geometry& geom, const Geometry& splitPoint
 std::unique_ptr<Geometry>
 GeometrySplitter::split(const Geometry &geom, const Geometry &splitGeom)
 {
+    if (geom.hasDimension(geom::Dimension::P)) {
+        throw util::IllegalArgumentException("Input geometry must be linear or polygonal.");
+    }
+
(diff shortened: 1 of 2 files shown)


dbaston:
Thanks, @jorisvandenbossche , I'm fine with this. It's also consistent with PostGIS:

```
dan=# select st_split('GEOMETRYCOLLECTION (POINT (3 7), LINESTRING (3 7, 3 8))', 'POINT (3 7)');
ERROR:  Splitting of Point geometries is unsupported
```

Thoughts, @gacarrillor?

jorisvandenbossche:
The `splitLinealWithEdge` and `splitPolygonalWithEdge` private functions also start with a similar check. Since we check up-front, I could remove them for simplifying the code? 
Although those also still verify the internal requirement that the first gets a linear input and the second a polygonal input, so that might still be useful to keep.

gacarrillor:
> Thanks, @jorisvandenbossche , I'm fine with this. It's also consistent with PostGIS:
> 
> ```
> dan=# select st_split('GEOMETRYCOLLECTION (POINT (3 7), LINESTRING (3 7, 3 8))', 'POINT (3 7)');
> ERROR:  Splitting of Point geometries is unsupported
> ```
> 
> Thoughts, @gacarrillor?

Also looks good to me, thanks!

PR #1499, opened 2026-08-05
Fix intersection computation of slightly overlapping polygons

### LINKED ISSUE #1405: Intersection of polygons gives incorrect results, depending on order of arguments
Picture shows it, code to reproduce below.  Both geometries are valid.  The adjustment is setting precision to a very small value on one geometry.  The only difference between sub-plots is the order used in calculating the intersection in the second plot, and a tiny rounding difference in the third.
Image

Code to reproduce:
```
import shapely
from matplotlib import pyplot as plt
from shapely.plotting import plot_polygon, plot_line

# Define shapes
shape1 = shapely.from_wkt(
    'POLYGON ((0.3081742230052605 0.8574917215741684, 0.3235558676951882 0.959764955575765, 0.3268384579963268 0.9761955555650583, 0.3309851357533714 0.9899402381062132, 0.3364277995673436 1.0032251410108646, 0.3431165154642001 1.015928381315176, 0.3509899175067092 1.0279334125460768, 0.3599757707991199 1.039130093982343, 0.3699916342095888 1.0494157011457903, 0.3809456167302097 1.058695868251819, 0.392737220535371 1.0668854539727242, 0.4052582630037556 1.073909322570786, 0.4183938692448725 1.079703033234622, 0.4320235260241182 1.0842134312944256, 0.4460221874170914 1.0873991358919737, 0.4602614220492758 1.0892309196312477, 0.4746105913957159 1.0896919767265496, 0.4913319068024928 1.0886253811328888, 1.6700718959842467 0.9565165637600046, 1.57979867649902 0.6026736686526396, 1.091045353275972 0.0772874773100065, 1.0880332290753003 0.0751954761072001, 1.0807249769395977 0.0690039561787506, 1.0740426168917647 0.0621416291321494, 1.0680474565569975 0.0546714537082506, 1.063366321369268 0.0475338562857705, 1.0334449647384247 0.0153698445990219, 0.7231700686154534 0.6266171384924346, 0.7149755148483247 0.6410255508351703, 0.7068170547053731 0.6526513145597219, 0.697594844851672 0.6634526894489446, 0.6873916890776646 0.6733326928059559, 0.6762991988285427 0.6822026146859029, 0.6644169706492398 0.6899828143991633, 0.6497523731524246 0.6977095626115124, 0.3081742230052605 0.8574917215741684))',
)
shape2 = shapely.from_wkt(
'POLYGON ((1.108561208420377086625308038492221385240554809570312500000000 0.096116174702266674034767390821798471733927726745605468750000, 1.579798676499019949304170040704775601625442504882812500000000 0.602673668652639604736975798005005344748497009277343750000000, 1.445650745099990475139861700881738215684890747070312500000000 0.076855572327130111665027811795880552381277084350585937500000, 1.444900805558562373320796723419334739446640014648437500000000 0.076572789294606147247890248763724230229854583740234375000000, 1.429693401346880499147573573281988501548767089843750000000000 0.072597961830626389634346651291707530617713928222656250000000, 1.414153318559506544715986819937825202941894531250000000000000 0.070237829803077911350328577100299298763275146484375000000000, 1.398451511191880936024745096801780164241790771484375000000000 0.069518356651499413345618449966423213481903076171875000000000, 1.380137763162443853559580020373687148094177246093750000000000 0.070602770087456412362314495112514123320579528808593750000000, 1.160083364028043684257340828480664640665054321289062500000000 0.095265653076203160587454021879239007830619812011718750000000, 1.150524405216126089257500098028685897588729858398437500000000 0.095875386407627882778825778586906380951404571533203125000000, 1.140950960368624311769281121087260544300079345703125000000000 0.095567779425867860343402071521268226206302642822265625000000, 1.131450861503619176318125028046779334545135498046875000000000 0.094345654285582614129168632643995806574821472167968750000000, 1.122111267723040706201231841987464576959609985351562500000000 0.092220223431104253464241082838270813226699829101562500000000, 1.113017865567405006999024408287368714809417724609375000000000 0.089210986727341734514595827931771054863929748535156250000000, 1.112913182538546097433140857901889830827713012695312500000000 0.089164814272908357595248673987953225150704383850097656250000, 1.108561208420377086625308038492221385240554809570312500000000 0.096116174702266674034767390821798471733927726745605468750000))'
)

# Define plotting function
def plot_intersection(ax, A, B):
    plot_polygon(A, ax=ax, facecolor='none', edgecolor='r', hatch="/")
    plot_polygon(B, ax=ax, facecolor='none', edgecolor='b', hatch="\\")
    plot_polygon(shapely.intersection(A, B), ax=ax, linewidth=5, linestyle="-.", edgecolor='g', alpha=0.5, hatch="||||")
    return ax

# Make sure there's nothing wrong with the geometry
assert shape1.is_valid
assert shape2.is_valid

# And do some clean up just in case...
shape1 = shape1.buffer(0)
shape2 = shape2.buffer(0)
shape1 = shape1.normalize()
shape2 = shape2.normalize()

# Plot outline of shapes and shade their intersection
fig, [ax1, ax2, ax3] = plt.subplots(1, 3, sharex=True, sharey=True)

ax1.set_title("Wrong behavior")
plot_intersection(ax1, shape1, shape2)
ax2.set_title("Expected behavior")
plot_intersection(ax2, shape2, shape1)
ax3.set_title("Behavior, after \"adjusting\" geometry")
plot_intersection(ax3, shape1, shapely.set_precision(shape2, 1e-300))
plt.show()

```

Version info:
```
Python 3.13.7 (main, Mar  3 2026, 12:19:54) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import shapely
>>> shapely.__version__
'2.1.2'
>>> shapely.geos_version_string
'3.13.1'
```
On Ubuntu 25.10

petrbel:
**Problem**: The polygon intersection computation doesn't work well when two polygons ~almost share an edge~ overlap very slightly. One of the symptoms is that intersection operation isn't commutative (as described in #1405)

**Content**: This PR consists of two commits: one adding a trivial not-working example (a unit test), two a proposed fix using `equals2D` instead of trivial independent coord-by-coord comparison. What i propose is to actually compare the dst point instead of float comparison of subtraction. That, should the points be very close, can produce 0 even for non-equivalent edges due to float arithmetics.

---

**Example description**: Below I enclose the problem description. It's similar to #1405 but, I believe, more straightforward and debuggable.
- Polygon A: 	`POLYGON ((1 1, 0 0.5, 0 0, 1 1))`  - a triangle with area 0.25
- Polygon B:   `POLYGON ((1 1, 0 0.49999999999999994, 0 1, 1 1))` - another triangle with area (almost) 0.25 which shares a common vertex `1 1` and a tiny overlap close to one of the edges with A ~and almost shares a second vertex `0 0.49999999999999994` with A (`0 0.5`)~
- ~Apart from point `(1 1)` these polygons don't share any point. Their intersection is a single point and the intersection area is mathematically `0`. Their union area is (almost) `0.5`~
- The intersection of these polygons a tiny neighborhood of edge `(1 1) -> (0 0.5)` with area close to `0`
- Prior this PR, the intersection area of these polygons was `0.25` as well as the union area. The newly proposed unit tests demonstrate this fact.
- With the second commit applied, the behavior is fixed

**Example visualization**: I used dbeaver with postgis only for the visualization purposes. I know this postgis isn't up-to-date and linked geos is only `3.13.1`, however, the same behavior is replicated directly in the tests. 
`postgis_full_version()`  

```
: POSTGIS="3.6.4 94d984b" [EXTENSION] PGSQL="180" GEOS="3.13.1-CAPI-1.19.2" SFCGAL="SFCGAL 2.0.0, CGAL 6.0, BOOST 1.83.0" PROJ="9.6.0 NETWORK_ENABLED=OFF URL_ENDPOINT=https://cdn.proj.org USER_WRITABLE_DIRECTORY=/var/lib/postgresql/.local/share/proj DATABASE_PATH=/usr/share/proj/proj.db" (compiled against PROJ 9.6.0) LIBXML="2.9.14" LIBJSON="0.18" LIBPROTOBUF="1.5.1" WAGYU="0.5.0 (Internal)" TOPOLOGY
```
image

The source sql in case anybody would like to replicate:
```sql
with cte as (
select
	'POLYGON ((1 1, 0 0.5, 0 0, 1 1))'::geometry as a,
    'POLYGON ((1 1, 0 0.49999999999999994, 0 1, 1 1))'::geometry as b
)
select
    a,
    b,
	st_area(a) as area_a,
	st_area(b) as area_b,
	st_area(st_intersection(a, b)) as area_ab,
	st_area(st_intersection(b, a)) as area_ba,
	postgis_full_version() as postgis_version
from cte
```
---

**Disclaimer**: while I found out about the problem myself and basically crafted the examples, I admit using a LLM for some hints and coding assistence. In case it's not allowed feel free to close the PR, however, please consider the revealed/described bug. I believe the code sheds some light to it.

---

Closes #1405

### DIFF
--- a/src/edgegraph/HalfEdge.cpp
+++ b/src/edgegraph/HalfEdge.cpp
@@ -179,6 +179,8 @@ HalfEdge::findLowest() const
 int
 HalfEdge::compareAngularDirection(const HalfEdge* e) const
 {
+    assert(orig().equals2D(e->orig()));
+
     // same vector
     if (directionPt().equals2D(e->directionPt()))
         return 0;
@@ -267,5 +269,3 @@ HalfEdge::toStringNode(const HalfEdge* he, std::ostream& os)
 
 } // namespace geos.edgegraph
 } // namespace geos
-
-
--- a/src/edgegraph/HalfEdge.cpp
+++ b/src/edgegraph/HalfEdge.cpp
@@ -179,17 +179,12 @@ HalfEdge::findLowest() const
 int
 HalfEdge::compareAngularDirection(const HalfEdge* e) const
 {
-    double dx = directionX();
-    double dy = directionY();
-    double dx2 = e->directionX();
-    double dy2 = e->directionY();
-
     // same vector
-    if (dx == dx2 && dy == dy2)
+    if (directionPt().equals2D(e->directionPt()))
         return 0;
 
-    int quadrant = geom::Quadrant::quadrant(dx, dy);
-    int quadrant2 = geom::Quadrant::quadrant(dx2, dy2);
+    int quadrant = geom::Quadrant::quadrant(directionX(), directionY());
+    int quadrant2 = geom::Quadrant::quadrant(e->directionX(), e->directionY());
 
     /**
     * If the direction vectors are in different quadrants,
--- a/src/geomgraph/EdgeEnd.cpp
+++ b/src/geomgraph/EdgeEnd.cpp
@@ -160,7 +160,7 @@ int
 EdgeEnd::compareDirection(const EdgeEnd* e) const
 {
     assert(e);
-    if(dx == e->dx && dy == e->dy) {
+    if(p1.equals2D(e->p1)) {
         return 0;
     }
 
--- a/src/geomgraph/EdgeEnd.cpp
+++ b/src/geomgraph/EdgeEnd.cpp
@@ -160,6 +160,8 @@ int
 EdgeEnd::compareDirection(const EdgeEnd* e) const
 {
     assert(e);
+    assert(p0.equals2D(e->p0));
+
     if(p1.equals2D(e->p1)) {
         return 0;
     }
@@ -215,4 +217,3 @@ operator<< (std::ostream& os, const EdgeEnd& ee)
 
 } // namespace geos.geomgraph
 } // namespace geos
-
--- a/tests/unit/edgegraph/EdgeGraphTest.cpp
+++ b/tests/unit/edgegraph/EdgeGraphTest.cpp
@@ -170,6 +170,25 @@ void object::test<5> ()
     checkNextPrev(*graph);
 }
 
+// testSimilarEdgesDirection
+template<>
+template<>
+void object::test<6> ()
+{
+    EdgeGraph graph;
+    HalfEdge* e1 = addEdge(graph, 1, 1, 0, 0.5);
+    addEdge(graph, 1, 1, 0, 0.49999999999999994);
+    addEdge(graph, 1, 1, 0, 1);
+
+    HalfEdge* eUpper = findEdge(graph, 1, 1, 0, 0.5);
+    HalfEdge* eLower = findEdge(graph, 1, 1, 0, 0.49999999999999994);
+    ensure("edges with distinct direction points must not compare equal", eUpper->compareTo(eLower) != 0);
+    ensure("edge comparison must be antisymmetric", eUpper->compareTo(eLower) == -eLower->compareTo(eUpper));
+
+    checkNodeValid(e1);
+    checkNextPrev(graph);
+}
+
 
 
 } // namespace tut
--- a/tests/unit/operation/overlayng/OverlayNGRobustTest.cpp
+++ b/tests/unit/operation/overlayng/OverlayNGRobustTest.cpp
@@ -61,6 +61,17 @@ struct test_overlayngrobust_data {
         ensure_NO_THROW( OverlayNGRobust::Overlay(geom_a.get(), geom_b.get(), opCode) );
     }
 
+    void
+    checkOverlaySymmetricArea(const std::string& a, const std::string& b, int opCode, double expectedArea)
+    {
+        std::unique_ptr<Geometry> geom_a = r.read(a);
+        std::unique_ptr<Geometry> geom_b = r.read(b);
+        double areaAB = OverlayNGRobust::Overlay(geom_a.get(), geom_b.get(), opCode)->getArea();
+        double areaBA = OverlayNGRobust::Overlay(geom_b.get(), geom_a.get(), opCode)->getArea();
+        ensure_equals("overlay is not commutative", areaAB, areaBA, 1e-12);
+        ensure_equals("unexpected overlay area", areaAB, expectedArea, 1e-12);
+    }
+
     std::unique_ptr<Geometry>
     double2geom(const std::vector<double>& x, const std::vector<double>& y)
     {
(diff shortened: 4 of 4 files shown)


dr-jts:
Why do you say "Apart from point (1 1) these polygons don't share any point"?

In fact they properly intersect, as can be seen in this image with magnified topology:

image

petrbel:
@dr-jts you are right. I got confused by the y-axis.

However, the behavior is still flawed, isn't it? The intersection area is most definitely not 0.25. What do you think?

edit: I updated the initial description so that it's more clear

dr-jts:
> @dr-jts you are right. I got confused by the y-axis.
> 
> However, the behavior is still flawed, isn't it? The intersection area is most definitely not 0.25. What do you think?

Looks like a bug in GEOS.  JTS gives the expected intersection result (with area = 2.7755575615628914E-1 ):
```
POLYGON ((0 0.5, 1 1, 0 0.4999999999999999, 0 0.5))
```
image

dbaston:
> Looks like a bug in GEOS

Yes and no, I think. When I pull this up in the debugger, JTS and GEOS return the same result for all invocations of `HalfEdge::compareAngularDirection`. In particular, both return `0` for when comparing the edges `(1 1, 0 0.5)` and `(1 1, 0 0.49999999999999994)`. But we know the angular direction is _not_ equal here. I suspect that treating this as equal sets us up for an insertion order dependency later on, and it currently happens to work out correctly in JTS and not in GEOS.

I think the fix here is OK, though I would be more comfortable if we also asserted that `orig().equals2D(e->orig())`

petrbel:
@dbaston I think your exaplanation is correct. My explanation is that if we use the coordinate-wise float comparison and the difference is less than what floats can handle, the points are considered the same. Then, the quadrands tie-breaker never runs.
This, I believe, leads to `HalfEdge::isEdgesSorted` (by using `compareTo` which finally uses `compareAngularDirection`) considering an edge unsorted (because the condition checks for `> 0`). In such case the order of the nodes is determined by the order of their insertion. And that, in my understanding, might swap the interior and exterior of the intersection, leading to the whole input polygon is considered the intersection.

If that makes sense.

---

> I think the fix here is OK, though I would be more comfortable if we also asserted that orig().equals2D(e->orig())

Good idea. I added the asserts

Please let me know if you have any further remarks

dr-jts:
Agreed, it's always better to use more robust logic that avoids arithmetic operations.

dbaston (approved):

dr-jts:
Mpre notes: 
*  `compareAngularDirection` should use `Quadrant.quadrant(Coordinate, Coordinate)`, since that's more robust by avoiding the arithmetic to compute direction vector components (see https://github.com/locationtech/jts/pull/1224 and https://github.com/locationtech/jts/pull/1226)
* the methods `directionX` and `directionY` are badly named.  They are not the X and Y ordinates of the direction point, but the direction **vector**.  They should probably be called `directionVectorX` and `directionVectorY`.  And with the above change they can be removed.

PR #1500, opened 2026-08-06
GeometrySplitter: Avoid dropped component

### LINKED ISSUE #1497: CurvePolygon: Missing curved geometry after splitting
Geometry A
`CURVEPOLYGON (COMPOUNDCURVE((5 0, 0 0, 0 5, 5 5), CIRCULARSTRING(5 5, 7 1, 5 0)))`

Image

Geometry B
`CURVEPOLYGON (COMPOUNDCURVE((5 0, 0 0, 5 5), CIRCULARSTRING(5 5, 7 4, 5 0)))`

Image

Split geometry result
`GEOMETRYCOLLECTION (POLYGON ((0 0, 0 5, 5 5, 0 0)))`

Image

The curved part of the geometry is missing from the result.

------------

However, if I split the same Geometry A with a non-curved polygon, the curved part of the original geometry appears in the result:

Geometry A
`CURVEPOLYGON (COMPOUNDCURVE((5 0, 0 0, 0 5, 5 5), CIRCULARSTRING(5 5, 7 1, 5 0)))`

Geometry B:
`POLYGON ((0 0, 0 5, 5 5, 0 0))`

Split geometry result
`GEOMETRYCOLLECTION ( CURVEPOLYGON (COMPOUNDCURVE ((5 0, 0 0, 5 5), CIRCULARSTRING (5 5, 7 1, 5 0))), POLYGON ((0 0, 0 5, 5 5, 0 0)))`

--------------
Testing with GEOS master (rev. https://github.com/libgeos/geos/commit/a03789b339d5adfe3930ee7e972cb156030398f4)

dbaston:
First commit adds an area check to avoid silently dropping a component of the result.

Second component handles a case where the input geometry and the splitting edge share a coincident arc defined with different control points.

Resolves #1497

### DIFF
--- a/include/geos/noding/DirectionIndependentArcString.h
+++ b/include/geos/noding/DirectionIndependentArcString.h
@@ -1,0 +1,52 @@
+/**********************************************************************
+*
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2026 ISciences, LLC
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************/
+
+#pragma once
+
+#include <memory>
+#include <geos/export.h>
+
+
+namespace geos::noding {
+class ArcString;
+}
+
+namespace geos::noding {
+
+/// The DirectionIndependentArcString is a thin wrapper over an ArcString that provides specialized
+/// equality-testing and hashing functions. DirectionIndependentArcStrings are considered equal
+/// if they cover the same path, regardless of direction. The control points of arcs are not required
+/// to be identical, but must describe the same arc.
+class GEOS_DLL DirectionIndependentArcString {
+
+public:
+
+    explicit DirectionIndependentArcString(const ArcString& as) : m_as(as) {}
+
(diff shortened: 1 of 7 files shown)


grootstebozewolf (approved):
Looks solid — matches the diagnosis on #1497 (OCA treating coordinate-identical-path arcs as distinct when control points differ, then Polygonizer dropping faces).

**What works**
- `DirectionIndependentArcString`: path equality by endpoints + orientation (+ reverse) and **center**, not control-point coordinates — right key for “same arc, different mid”.
- Splitting `GeometryNoder::toGeometry` linear (`OrientedCoordinateArray`) vs arc paths keeps the old contract for chords.
- Area-preservation throw in `splitPolygonalWithEdge` is a good hard fail instead of silent area loss.
- GH-1497 regression + unit cases for reverse / multi-section / linear degenerates / different sagitta; CI green across the matrix.

**Nits (non-blocking)**
1. Header comment typo: `DirectionIndependentArcStrings/` → drop the stray `/`.
2. New test files missing trailing newline (style only).
3. Optional follow-up: if `ArcString` can be empty when hashed, `getArc(0)` / `getSize()-1` in `std::hash` would be undefined — only matters if empty paths can reach `toGeometry`.

Happy to see this land. Thanks for the careful noding split rather than papering over it only in the splitter.

---
*Review assisted by xAI Grok; human checked design vs #1497 and the diff.*

PR #1505, opened 2026-08-10
fix: PointLocation::isOnSegment float-collinear robustness (#968)

### LINKED ISSUE #968: Potential bug in function covers

grootstebozewolf:
## Summary

Fixes [libgeos/geos#968](https://github.com/libgeos/geos/issues/968): `LINESTRING (1 0, 0 2)` should **cover** `POINT (0.9 0.2)` and the symmetric `POINT (0.1 1.8)`.

### Cause

`PointLocation::isOnSegment` used `Orientation::index` (Ozaki filter → DD). For these decimal-collinear encodings the filter is **uncertain** and DD reports non-collinear, so `covers` / `within` fail even though the printed decimals are collinear and the binary64 residual is 1 ulp of the determinant (`|det| = 2^{-54} ≈ 5.55e-17`). `@dr-jts` noted this on the issue (high-precision orientation ironically causes the miss).

Ozaki `FAILURE` only means the double filter cannot certify a side (`|det| < |detleft+detright|·K`). It is not an on-segment test. A first revision that treated every nonzero `FAILURE` as on incorrectly accepted

```
LINESTRING (0 0, 1e16 1e16)  /  POINT (5e15, 5e15+1)
```

(`|det| ≈ 2^{53}`, true cross \(10^{16}\), distance ≈ 0.637). DD separates that point. `@fallenmi` caught this on `9cfff20`.

### Fix (tip `31224e328`)

For **on-segment membership only**:

| Filter | `|det|` | `isOnSegment` |
|---|---|---|
| certain `LEFT` / `RIGHT` | — | off |
| `STRAIGHT` | — | on |
| `FAILURE` | `0 < |det| ≤ 1e-12` | **on** (treat as a binary64 residual) |
| `FAILURE` | `det == 0` or `|det| > 1e-12` | fall through to `Orientation::index` (DD) |

`1e-12` is an explicit absolute cap on `|2 * signed triangle area|`. It is **not** the published Ozaki `K` (≈ `3.33e-16` relative). It sits between the two measured scales: #968 residual \(2^{-54}\) and the review pair rounded det \(2^{53}\).

`Orientation::index` is unchanged for callers that need DD separation.

### Tests

- `PointLocation` test 7 — #968 pins `(0.9, 0.2)`, `(0.1, 1.8)`, control off-line
- `PointLocation` test 8 — large-scale off-line review pair
- `capi::GEOSCovers` tests 3 and 4 — same two geometries through `covers`

Local `geosop` on this revision: `covers=false` for the review pair (same distance as main); the #968 cases remain `true`.

### What this is not

- Not “Ozaki `FAILURE` ⇒ on the line.” That rule is withdrawn.
- Not an Ozaki soundness theorem, and not a change to the production orientation filter.
- Not a proof that every geometrically-on input has `|det| ≤ 1e-12`. The cap is a documented sandwich of two measured scales, not a length-relative height test.

### DIFF
--- a/src/algorithm/PointLocation.cpp
+++ b/src/algorithm/PointLocation.cpp
@@ -44,26 +44,31 @@ PointLocation::isOnSegment(const geom::CoordinateXY& p, const geom::CoordinateXY
     if (p.equals2D(p0) || p.equals2D(p1))
         return true;
 
-    // GEOS #968 / PostGIS #5563:
-    // Orientation::index uses an Ozaki filter then DD.  For decimal-collinear
-    // points (e.g. LINESTRING(1 0,0 2) / POINT(0.9 0.2)) the filter is
-    // uncertain and DD reports non-collinear, so covers/within fail even though
-    // the point is collinear under ordinary double residual.
+    // Membership on already-rounded coordinates.  A certain Ozaki LEFT/RIGHT
+    // certifies off; STRAIGHT certifies on.  FAILURE means the double
+    // determinant cannot certify a side.
     //
-    // For on-segment membership only: trust the filter when it is certain;
-    // when uncertain (FAILURE), treat as on-segment rather than escalating to
-    // DD.  Topology predicates that need DD separation still use
-    // Orientation::index directly.
-    //
-    // Suggested direction matches @dr-jts on #968 (prefer FP collinearity for
-    // this decision).  Updates RobustLineIntersector testA where FP is
-    // uncertain on a huge segment that DD separates.
+    // A non-zero Ozaki det that still fails the filter is a 1-ulp residual
+    // (decimal-collinear encodings such as (1,0)-(0,2) with (0.9,0.2)).
+    // Treat that as on-segment.  An exact-zero Ozaki det on a huge segment
+    // is the robustness trap: fall through to DD so a point just off the
+    // line stays off.
     const int filt = CGAlgorithmsDD::orientationIndexFilter(
         p0.x, p0.y, p1.x, p1.y, p.x, p.y);
-    if (filt == CGAlgorithmsDD::FAILURE) {
+    if (filt == CGAlgorithmsDD::LEFT || filt == CGAlgorithmsDD::RIGHT) {
+        return false;
+    }
+    if (filt == CGAlgorithmsDD::STRAIGHT) {
         return true;
     }
-    return filt == CGAlgorithmsDD::STRAIGHT;
+
+    const double detleft = (p0.x - p.x) * (p1.y - p.y);
+    const double detright = (p0.y - p.y) * (p1.x - p.x);
+    const double det = detleft - detright;
+    if (det != 0.0) {
+        return true;
+    }
+    return Orientation::index(p0, p1, p) == Orientation::COLLINEAR;
 }
 
 /* public static */
--- a/src/algorithm/PointLocation.cpp
+++ b/src/algorithm/PointLocation.cpp
@@ -48,11 +48,12 @@ PointLocation::isOnSegment(const geom::CoordinateXY& p, const geom::CoordinateXY
     // certifies off; STRAIGHT certifies on.  FAILURE means the double
     // determinant cannot certify a side.
     //
-    // A non-zero Ozaki det that still fails the filter is a 1-ulp residual
-    // (decimal-collinear encodings such as (1,0)-(0,2) with (0.9,0.2)).
-    // Treat that as on-segment.  An exact-zero Ozaki det on a huge segment
-    // is the robustness trap: fall through to DD so a point just off the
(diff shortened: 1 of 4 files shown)


dbaston:
I understand that avoiding DD makes a particular test case pass but not why that is generally desirable.

grootstebozewolf:
The reason to skip DD here is general, not specific to #968.

`isOnSegment` is a membership test on already-rounded binary64 coordinates. The Ozaki filter's `FAILURE` means `|det|` is inside the error bound of the double determinant: the point cannot be certified left or right at the precision of the input. DD then computes the exact sign of those bit patterns treated as rationals. That is the right question for `Orientation::index` (unchanged). It is the wrong question for membership. Extra precision there reports "off" for points that are not representably off the segment — which is why decimal-collinear inputs fail `covers` even though they are collinear in the coordinate model that was supplied.

So the policy is: a point is on the segment unless double arithmetic can certify it is not. That does accept some points that DD would separate (testA is the documented case). Callers that need that separation already use `Orientation::index`.

grootstebozewolf:
@dr-jts @dbaston — short decision ask.

#968 said high-precision orientation causes the miss, and sketched changing `Orientation` (FP if collinear, else DD). This PR does **not** do that. It keeps `Orientation::index` as filter→DD, and applies "uncertain ⇒ on" only in `PointLocation::isOnSegment`.

Dave's question was why skipping DD is desirable in general. The claim is: it is the right general rule for **membership**, and the wrong rule for **orientation**.

Martin: is that split what you wanted, or do you still want the global `Orientation` change?
Dave: if Martin is ok with the split, is that enough to unblock?

fallenmi (changes requested):
The new nonzero-`det` fallback introduces a false-positive topology result for exactly representable off-segment coordinates.

At exact head `9cfff20cab8acabf3c278169a9449a328ef0f0b8` versus exact PR base `14fd2d70311aae9646e5d2e5364921822074de2b`, I built the Release `geosop` target from each tree and ran:

```sh
geosop -a 'LINESTRING (0 0, 10000000000000000 10000000000000000)' \
  -b 'POINT (5000000000000000 5000000000000001)' covers
geosop -a 'LINESTRING (0 0, 10000000000000000 10000000000000000)' \
  -b 'POINT (5000000000000000 5000000000000001)' distance
```

The base reports `false` and distance `0.63690516725257718`; the head reports `true` with the same positive distance. The point's integer coordinates are exactly representable and it is one unit above the line `y = x`. Here the Ozaki filter returns `FAILURE`, while the rounded double determinant is nonzero (`9007199254740992`) but below the error bound, so the new `det != 0.0` branch accepts a point that DD correctly separates.

Please avoid treating every nonzero filter failure as membership (or put the fuzzy behavior behind an explicit, documented tolerance) and add an off-line regression at large coordinate scale. The intended `LINESTRING (1 0, 0 2)` / `POINT (0.9 0.2)` case does pass on this head.

Reviewed by OpenAI Codex for @fallenmi at the exact head above. No repository files were changed.

grootstebozewolf:
@fallenmi addressed in 31224e328.

The previous `det != 0` branch treated every nonzero Ozaki `FAILURE` as on-segment. That incorrectly accepted `POINT (5e15, 5e15+1)` on `LINESTRING (0 0, 1e16 1e16)` (`|det| = 2^{53}`, distance ≈ 0.637). DD classifies it off.

`FAILURE` is now treated as on only when `0 < |det| ≤ 1e-12` — an explicit absolute bound on `|2 * signed triangle area|`, not Ozaki `K`. Larger uncertain determinants, and exact-zero determinants on huge segments (RobustLineIntersector testA), fall through to DD.

Added the large-scale off-line regression as `PointLocation` test 8 and `GEOSCovers` test 4. Local `geosop` on this revision reports `covers=false` for the review pair with the same distance as main; `(0.9 0.2)` and `(0.1 1.8)` remain `true`.

@dbaston the general rule is no longer “skip DD whenever the filter is uncertain.” It is: certify off/on when Ozaki is certain; treat only a *tiny* nonzero residual as membership; otherwise keep DD. That is a membership heuristic for already-rounded coordinates, not a claim that uncertain orientation means the point is on the line.

PR #1507, opened 2026-08-23
CircularString: even controls valid only as ABCA circumcircle

grootstebozewolf:
## Summary

Ingest stays lenient. `isValid` is the gate:

- empty is valid
- odd control count >= 3 is valid
- even controls are valid only as closed four-control `CIRCULARSTRING(A, B, C, A)` that defines a circumcircle

Open even leftovers, closed 6-control leftover, and collinear `CIRCULARSTRING(A, B, C, A)` ingest and return `isValid() == false`.

Witness: `CIRCULARSTRING (-5 0, 0 5, 5 0, -5 0)`.

## Tests

`tests/unit/geom/CircularStringTest.cpp` pins the four-control circumcircle, rejects leftover even counts on `isValid`, and keeps `CIRCULARSTRING (0 0, 1 0, 0 0)` ingestible.

### DIFF
--- a/tests/unit/geom/CircularStringTest.cpp
+++ b/tests/unit/geom/CircularStringTest.cpp
@@ -390,7 +390,7 @@ template<>
 template<>
 void object::test<15>() {
     set_test_name("liblwgeom: 10 segments per quadrant - circular");
-    auto cs = wktreader_.read<CircularString>("CIRCULARSTRING (0 0, 1 0, 0 0)");
+    auto cs = wktreader_.read<CircularString>("CIRCULARSTRING (0 0, 0.5 0.5, 1 0, 0.5 -0.5, 0 0)");
     auto ls = cs->getLinearized(CurveToLineParams::stepSizeDegrees(90.0 / 10));
     ensure_equals(ls->getNumPoints(), 41u); // PostGIS test has 40, but this is incorrect
     const auto* seq = ls->getCoordinatesRO();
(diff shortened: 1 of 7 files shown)


dbaston:
Why would an even length be valid if the CircularString is closed?

tests/unit/geom/CircularStringTest.cpp
```diff
@@ -390,7 +390,7 @@ template<>
 template<>
 void object::test<15>() {
     set_test_name("liblwgeom: 10 segments per quadrant - circular");
```
dbaston, line 396: Given that this test comes from PostGIS, I'm inclined to continue to tolerate this representation unless there is a reason not to.

grootstebozewolf: It is not compliant with ISO

dbaston: I understand that, and perhaps `isValid` should return false on such a geometry, but I don't see a practical advantage from erroring out on something that occurs in the wild.

grootstebozewolf:
> Why would an even length be valid if the CircularString is closed?

Arcs are represented by three points, A,B,C when it is closed it is A,B,C,A at a minimum, this might be the oddedst CircularString because it is even. This harmonizes with the rule that the first and third parameter of an arc segment never be the same. I discussed this on the GEOS channel, there where no objections.

tests/unit/geom/CircularStringTest.cpp
grootstebozewolf: Agreed, make ingest lenient and only trigger on isValid if not ISO.
PostGIS also says it will store invalid geometries and let ST_IsValid catch them later (§4.4).

dbaston:
> Arcs are represented by three points, A,B,C when it is closed it is A,B,C,A

What is the control point between C and A ?

> I discussed this on the GEOS channel, there where no objections.

Sorry, I haven't seen any such discussion.

grootstebozewolf:
> What is the control point between C and A ?

There are no control points needed between C and A, a circle is defined by three points, there are three distinct points in the WKT, it is all according to ISO. The first = last is just a notation indicating it is last (and the only valid one is the circle)

dbaston:
It doesn't look like PostGIS accepts this...is there other software that does?

```
# SELECT ST_Length('CIRCULARSTRING (0 0, 1 1, 2 0, 0 0)');
ERROR:  geometry must have an odd number of points
```

It seems like a particularly annoying special case, and would require that every function that operates on the coordinates of a CircularString first check if it has 4 points and is closed.

PR #1509, opened 2026-08-29
MakeValid: process envelope-connected components of a MultiPolygon independently

### LINKED ISSUE #1504: MakeValid on a MultiPolygon costs up to 55x the identical GeometryCollection
`MakeValid::build` dispatches `GEOS_GEOMETRYCOLLECTION` inputs per member, but sends `GEOS_MULTIPOLYGON` whole into `MakeValidPoly`. One invalid part therefore drags every part through boundary noding, `extractUniquePoints`, and the BuildArea/symdiff loop. For parts with disjoint envelopes the two inputs are semantically identical, so the difference is pure dispatch:

| parts | all-valid MULTIPOLYGON | MULTIPOLYGON + 1 bowtie | GEOMETRYCOLLECTION + 1 bowtie | penalty |
| --- | --- | --- | --- | --- |
| 401 | 0.0011 s | 0.0149 s | 0.0004 s | 34.2x |
| 1,601 | 0.0032 s | 0.0603 s | 0.0014 s | 41.8x |
| 6,401 | 0.0139 s | 0.2767 s | 0.0053 s | 52.2x |
| 14,401 | 0.0288 s | 0.6777 s | 0.0123 s | 54.9x |

The penalty grows with part count: the collection path is linear, the multipolygon path is not. Measured with GEOS 3.13.1 via shapely 2.1.2 on macOS arm64; the reproducer below is synthetic (an n×n grid of unit squares plus one bowtie) and runs in under 2 seconds.

Real-world case: a marching-squares coastal partition of 88,997 polygons / 20.7M vertices containing exactly one self-intersecting ring (4.26M vertices) takes >55 minutes and >3 GB through the MultiPolygon path, while the identical parts as a collection validate member-wise in seconds. `MakeValid` short-circuits each valid member through `IsValidOp` + clone.

I also measured `main` (at a03789b3, which includes the #1353 / JTS 906 `ptNotInList` replacement) on real-data fixtures of N valid parts plus one tiny bowtie:

| parts | vertices | GEOS 3.13.1 | GEOS main |
| --- | --- | --- | --- |
| 1,001 | 219,180 | 0.941 s | 0.887 s |
| 4,001 | 740,627 | 5.011 s | 3.854 s |
| 16,001 | 2,081,271 | 18.085 s | 13.917 s |

So #1353 recovers 1.2–1.3x on this shape, not the order of magnitude. Peak RSS at 16k parts / 2.1M vertices was 2.4–2.6 GB in both builds.

Suggested direction: group a MultiPolygon's parts into envelope-connected components and run `MakeValidPoly` per component. For disjoint envelopes this is exact, valid components take the existing `IsValidOp` fast path, and the common case (many valid parts, few invalid) becomes linear.

## Reproduction

```python
"""MakeValid dispatches a GEOMETRYCOLLECTION per member but sends a
MULTIPOLYGON through MakeValidPoly whole, so ONE invalid member costs the
whole collection. Synthetic grid + one bowtie - no external data."""
import time
import shapely
from shapely.geometry import Polygon, MultiPolygon, GeometryCollection

def grid(n):
    return [Polygon([(i*2, j*2), (i*2+1, j*2), (i*2+1, j*2+1), (i*2, j*2+1)])
            for i in range(n) for j in range(n)]

print(f"shapely {shapely.__version__} GEOS {shapely.geos_version_string}")
print(f"{'parts':>7} {'all-valid MP':>14} {'MP + 1 bowtie':>15} {'GC + 1 bowtie':>15} {'penalty':>9}")
for n in (20, 40, 80, 120):
    ps = grid(n)
    bow = Polygon([(-10, -10), (-9, -9), (-9, -10), (-10, -9), (-10, -10)])
    assert not bow.is_valid
    t = time.perf_counter(); shapely.make_valid(MultiPolygon(ps)); a = time.perf_counter()-t
    t = time.perf_counter(); shapely.make_valid(MultiPolygon(ps+[bow])); b = time.perf_counter()-t
    t = time.perf_counter(); shapely.make_valid(GeometryCollection(ps+[bow])); c = time.perf_counter()-t
    print(f"{len(ps)+1:7d} {a:13.4f}s {b:14.4f}s {c:14.4f}s {b/c:8.1f}x", flush=True)
```

artemkulyk:
# MakeValid: process envelope-connected components of a MultiPolygon independently

Fixes #1504

## Problem

`MakeValid::build` sends an entire MultiPolygon containing even one invalid part
through `MakeValidPoly`, so every part is dragged through boundary noding,
`extractUniquePoints`, and the BuildArea/symdiff loop — even parts that cannot
possibly interact with the invalid one. For parts with disjoint envelopes this
costs up to ~55x the equivalent GeometryCollection, which processes members
independently (measured in GH-1504). The real-world case reported there
(88,997 parts / 20.7M vertices with one self-intersecting ring) takes >55
minutes instead of seconds.

## Algorithm

For invalid MultiPolygons, the parts are partitioned into envelope-connected
components (union-find over a sorted x-sweep of the part envelopes; envelopes
that intersect — touching included — share a component). Polygons in different
components have disjoint envelopes and therefore cannot interact topologically,
so the split is exact:

* A MultiPolygon forming a **single** component takes the untouched
  whole-geometry path — connected/overlapping/nested cases behave exactly as
  before.
* With **several** components, each is made valid independently: a component
  made entirely of valid parts is passed through (the existing `IsValidOp`
  fast path), a component with an invalid part goes through `MakeValidPoly`.
* Per-component results are merged back into the same
  area/cut-edges/collapse-points structure `MakeValidPoly` would have returned
  for the whole MultiPolygon.

Correctness: each component's result is enclosed within the envelopes of its
own parts, and component envelopes are pairwise disjoint, so merged results
are disjoint and topologically equal to the whole-geometry result. Envelopes
that are null or non-finite are excluded from the sweep before sorting (they
would otherwise trip `Envelope` accessor asserts in debug builds), and their
linework still fails in the overlay exactly as the whole-geometry path does.

## Behavior changes

* Components made entirely of valid parts are now passed through unchanged
  (the documented fast path for already-valid geometries) instead of being
  rebuilt by overlay: they keep their original coordinates, including Z and M
  values, and their original ring orientation.
* Consequently, output element order follows the input/component order for
  pass-through parts; overlay-rebuilt pieces may still come out in any order.
  (Previously *everything* was rebuilt and could be reordered.)
* Non-finite linework (Inf/NaN coordinates) still fails with the same
  `IllegalArgumentException` as the whole-geometry algorithm.

## Performance

`perf_makevalid` (added under `benchmarks/operation/`, reproduces the GH-1504
synthetic grid + one bowtie), interleaved A/B against unmodified main,
Release build, medians:

| case | parts | before | after | speedup | peak RSS before → after |
|---|---|---|---|---|---|
| disjoint grid | 401 | 0.036 s | 0.0015 s | ~24x | 10 → 4 MB |
| disjoint grid | 1,601 | 0.155 s | 0.005 s | ~30x | 30 → 7 MB |
| disjoint grid | 6,401 | 0.78 s | 0.017 s | ~46x | 105 → 14 MB |
| disjoint grid | 14,401 | 4.0 s | 0.034–0.066 s | ~60x | 218 → 27 MB |
| disjoint grid | 88,805 | 15.4 s | 0.24–0.32 s | ~50–60x | 851 → 118 MB |
| corner-touching chain + bowtie | 1,201 | 0.20 s | 0.003 s | ~60x | 29 → 6 MB |
| overlapping chain (single component, cannot split) | 4,000 | ~1.1 s | ~1.0 s | parity | 135 → 134 MB |
| many valid parts + one huge (50k-vertex) bowtie | 49,730 | ~21 s | ~7 s | ~3x | 644 → 232 MB |

The equivalent GeometryCollection reference path runs at ~0.029 s for 14,401
parts; the optimized MultiPolygon path lands within ~2x of it while keeping
MultiPolygon semantics.

## Testing

* 13 new/extended unit tests in `tests/unit/operation/valid/MakeValidTest.cpp`:
  deterministic edge cases (disjoint + invalid, touching/crossing bowties,
  shared edges, nested shells, duplicates, collapses, empty parts, Z/M and
  mixed-dimension parts, SRID handling), a randomized MultiPolygon vs
  equivalent-GeometryCollection equivalence invariant (fixed seeds), and
  non-finite envelope handling.
* Differential testing against unmodified main: 36-case harness (deterministic
  edge cases + seeded randomized grids + overlap clusters), comparing results
  per dimension by symmetric-difference emptiness — 36/36 topologically
  equivalent, 18/36 byte-identical (the rest differ only in element order /
  ring orientation of pass-through parts, as described above).
* Full test suite: 535/535 passing. Debug-build (asserts enabled) verified for
  the empty/null-envelope cases.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,12 @@
   - Add progress reporting to GEOSCoverageSimplify, GEOSUnaryUnion (GH-1466, Even Rouault / Dan Baston)
 
 - Fixes/Improvements:
+  - MakeValid groups MultiPolygon parts into envelope-connected components and
+    processes each independently, so a single invalid part no longer forces all
+    parts through the whole-geometry polygonal algorithm (GH-1504, Artem Kulyk).
+    Components made entirely of valid parts are passed through unchanged,
+    keeping their original coordinates, including Z and M values, instead of
+    being rebuilt by overlay.
   - Buffer of Linestring includes spurious hole (GH-1217, Moritz Kirmse)
   - Preserve M values in GEOSInterpolate (GH-1390, Dan Baston)
   - Preserve M values in GEOSLineMerge (GH-1364, Dan Baston)
--- a/benchmarks/operation/CMakeLists.txt
+++ b/benchmarks/operation/CMakeLists.txt
@@ -11,6 +11,13 @@
 add_subdirectory(buffer)
 add_subdirectory(predicate)
 
+add_executable(perf_makevalid MakeValidPerfTest.cpp)
+target_include_directories(perf_makevalid PUBLIC
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+        $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/benchmarks>)
+target_link_libraries(perf_makevalid PRIVATE geos geos_cxx_flags)
+
 if (benchmark_FOUND)
     add_executable(perf_distance DistancePerfTest.cpp)
     target_include_directories(perf_distance PUBLIC
--- a/benchmarks/operation/MakeValidPerfTest.cpp
+++ b/benchmarks/operation/MakeValidPerfTest.cpp
@@ -1,0 +1,289 @@
+/**********************************************************************
+ *
+ * GEOS - Geometry Engine Open Source
+ * http://geos.osgeo.org
+ *
+ * Copyright (C) 2026 GEOS contributors
+ *
+ * This is free software; you can redistribute and/or modify it under
+ * the terms of the GNU Lesser General Public Licence as published
+ * by the Free Software Foundation.
+ * See the COPYING file for more information.
+ *
+ **********************************************************************
+ *
+ * Benchmark for MakeValid on MultiPolygon inputs (issue #1504).
+ *
+ * Reproduces the scenario where a MultiPolygon containing a single
+ * invalid part costs up to ~55x the equivalent GeometryCollection,
+ * because all parts are dragged through boundary noding, unique-point
+ * extraction and BuildArea/symdiff.
+ *
+ * Cases (run one per process for clean peak-RSS readings):
+ *   grid <n>      n*n valid unit squares spaced 2 apart + 1 far bowtie
+ *                 (the issue #1504 scenario; envelope-disjoint components)
+ *   chain <n>     n corner-touching squares (one envelope-connected
+ *                 component) + 1 bowtie far away (a second component)
+ *   overlap <n>   n overlapping squares (single connected component,
+ *                 adversarial: cannot be split)
+ *   bigbowtie <n> ~n small valid polygons + one huge invalid bowtie with
+ *                 ~n vertices (mimics issue #1504 real data shape)
+ *
+ * An optional 4th argument "coll" runs the same parts as a
+ * GeometryCollection, which MakeValid processes member-wise; it provides
+ * a reference for the best achievable per-part cost.
+ *
+ * Timing uses geos::util::Profile like the other GEOS benchmarks. Peak RSS
+ * uses getrusage where available (POSIX); ru_maxrss is bytes on macOS and
+ * KiB on Linux/BSD, which is normalized below. On platforms without
+ * getrusage (e.g. Windows) peak RSS is reported as unavailable.
+ *
+ * Usage: perf_makevalid <grid|chain|overlap|bigbowtie> [n] [reps] [coll]
+ **********************************************************************/
+
+#include <geos/geom/Coordinate.h>
+#include <geos/geom/CoordinateSequence.h>
+#include <geos/geom/GeometryFactory.h>
+#include <geos/geom/Polygon.h>
+#include <geos/geom/MultiPolygon.h>
+#include <geos/geom/GeometryCollection.h>
+#include <geos/operation/valid/MakeValid.h>
+#include <geos/profiler.h>
+
+#include <cmath>
+#include <cstdlib>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <vector>
+
+#if defined(_WIN32)
+// getrusage is not available on Windows; report peak RSS as unavailable.
+#else
+#include <sys/resource.h>
+#endif
+
+using namespace geos::geom;
+using namespace geos::operation::valid;
+
+namespace {
+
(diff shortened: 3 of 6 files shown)


fallenmi (changes requested):
The active-window invariant here is not valid. `order` is sorted by `minX`, but expiration is determined by `maxX`; a single early envelope with a long X span can pin `lo` even when it is Y-disjoint from every later polygon. Every short envelope behind it then remains in `[lo, hi)` after it has expired, so each later polygon is compared with all predecessors.

I reproduced this on exact head `d6f89ac02e2d043aae0e2bde22b332beb9e495a1` with one invalid, X-wide bowtie at Y=100 plus N mutually separated unit squares at Y=0. All component envelopes are disjoint and the output remains valid, but this loop performs `N(N+1)/2` envelope tests. In a Release build, three-run averages scaled as follows:

- 16k squares: 0.181 s
- 32k squares: 0.690 s
- 64k squares: 2.678 s

That is approximately 4x per doubling. At 64k, the loop performs 2,048,032,000 envelope checks. A local prototype that expires entries independently of their `minX` position produced the same valid result in 0.264 s at 64k and did not show the quadratic scaling.

Please maintain expiry independently by `maxX` (or otherwise remove expired entries even when an earlier long-span item remains active), and add a regression/benchmark with a long-X, Y-disjoint sentinel. The current focused MakeValid tests (19/19) and full suite (535/535) pass, so this is a complexity regression in the new optimization rather than a topological-result failure.

AI disclosure: I used OpenAI Codex to help inspect the diff, construct and run the reproducer, and draft this review. I verified the code path, exact-head results, and conclusion.

artemkulyk:
Thanks — that diagnosis is correct. `order` is sorted by `minX` while expiry is by `maxX`, so a long-span envelope could pin `lo` and leave expired short envelopes in `[lo, hi)`. Topology stayed valid; the cost was `N(N+1)/2` envelope tests.

The sweep now keeps an `active` set compacted by `maxX` (`>=` so touching envelopes still join). Expired entries drop even when an earlier long-span item remains live.

Added:
- unit test 21: long-X, Y-disjoint bowtie plus spaced squares; valid squares stay `equalsExact` pass-through clones
- `perf_makevalid sentinel [n]`: the same fixture at scale (`n` = square count)

Local Release `perf_makevalid sentinel`:

| n | time |
|---|---|
| 16k | 0.094 s |
| 32k | 0.246 s |
| 64k | 0.698 s |

That is ~2.6–2.8× per doubling (MakeValid on the squares), not ~4×, and 0.70 s at 64k vs 2.68 s on the old window. Fix is in 365b803c5.

Issue #1446, opened 2026-06-16
Expose revision-level version information in headers and at runtime


strk:
It would be useful to know which exact revision does a GEOS library come from, maybe the output of `git describe`.
It's ok to report the tag, if it comes form a tagged tree.
Ideally it would contain a "dirty" flag if it's from a non-clean tree (with local changes).

strk:
Downstream issue (for PostGIS): https://trac.osgeo.org/postgis/ticket/6117

strk:
The way GDAL does it: https://github.com/OSGeo/gdal/blob/master/cmake/helpers/generate_gdal_version_h.cmake

strk:
Would you be ok with GEOSVersion including the revision or do you think a separate GEOSRevision function would be safer for backward compatibility ? I'd keep it simple and include everything in the existing function.

What's your opinion @dbaston @pramsey @robe2 ?

robe2:
> Would you be ok with GEOSVersion including the revision or do you think a separate GEOSRevision function would be safer for backward compatibility ? I'd keep it simple and include everything in the existing function.
> 
> What's your opinion [@dbaston](https://github.com/dbaston) [@pramsey](https://github.com/pramsey) [@robe2](https://github.com/robe2) ?

I'd be fine with it.  Think it would make issues easier to trace with say our PostGIS oci builds that are not always against released.

dbaston:
`GEOSversion` has been around since 2.2, I don't think it's a good idea to change the format of its return value now.

strk:
> `GEOSversion` has been around since 2.2, I don't think it's a good idea to change the format of its return value now.

How would you call that format ? Including a revision in a version string can be considered standard versioning by someone (for example Debian packages built from git repos). At the moment my `geos-config --version` outputs `3.15.0dev`,  do you really think `3.15.0rc1-6-g63c2253c6` would break someone's code ?

dbaston:
It may break some downstream code, for example this:

https://github.com/r-spatial/sf/blob/6ede89c5bf68abda997744c77106854eaeaa1c8c/R/geom-transformers.R#L568

But in any case it doesn't seem worth it for a change that benefits a tiny number of users (those who build GEOS themselves from untagged versions)

robe2:
> It may break some downstream code, for example this:
> 
> https://github.com/r-spatial/sf/blob/6ede89c5bf68abda997744c77106854eaeaa1c8c/R/geom-transformers.R#L568
> 
[...]

PR #1511, opened 2026-08-31
Expose revision-level version information

strk:
References GH-1446


dbaston:
image

Two minutes to review!

This appears to capture the revision at the point that `cmake` was run, not necessarily the point when GEOS was compiled. I don't know if that matches the intent or not.

Issue #1334, opened 2025-12-02
OpenMP SIMD compiler support detection in CMake


svenweb:
Detect if the compiler supports OpenMP and `#pragma omp simd` then set an `HAVE_OPEN_SIMD=1` flag accordingly. The `HAVE_OPEN_SIMD` flag could be used in #1331 and later on for any SIMD vectorization in the project.

The basic CMake for this would look something like:
- Check if compiler/toolchain supports basic OpenMP with [FindOpenMP](https://cmake.org/cmake/help/latest/module/FindOpenMP.html)
- If supported, set CMake OpenMP compiler flags

However detecting support for `#pragma omp simd` specifically could require more checks as different compilers support the SIMD subset of OpenMP unevenly, see [clang has limited support for vectorization](https://releases.llvm.org/16.0.0/tools/clang/docs/OpenMPSupport.html#id1).

Using a try_compile method could be a good option, I will update this issue as I try out some solutions.

PR #1522, opened 2026-09-09
Fix MSVC/pointer-width flags leaking into exported geos_cxx_flags target

### LINKED ISSUE #1517: `geos_cxx_flags` leaks MSVC-only `/fp:precise` to every consumer, breaking cross-compilation from MSVC-built GEOS
### Environment

GEOS 3.14.1 (issue still present in 3.15.0), built and installed with MSVC `cl.exe` on Windows. Consuming project built with `clang++`/`clang-cl`.

### Current Behavior

The installed `geos-targets.cmake` sets `GEOS::geos_cxx_flags`'s `INTERFACE_COMPILE_OPTIONS` to:

```cmake
"$<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:-ffp-contract=off>;$<$<CXX_COMPILER_ID:GNU>:-ffp-contract=off>;$<$<BOOL:1>:/fp:precise>;$<$<AND:$<CXX_COMPILER_ID:GNU>,$<EQUAL:4,8>>:-ffloat-store>"
```

The third entry is `$<$<BOOL:1>:/fp:precise>`. `$<BOOL:1>` is always true. Every consumer of `GEOS::geos` or `GEOS::geos_cxx_flags` receives `/fp:precise`, an MSVC-only flag, regardless of its own compiler. A project that links `GEOS::geos` and builds with `-DCMAKE_CXX_COMPILER=clang++` fails: Clang rejects `/fp:precise` as an unrecognized argument. The same failure occurs in tools that only enumerate compile options against a Clang frontend, such as `clang-scan-deps`, even with no compile step.

### Root cause

GEOS's `CMakeLists.txt` defines the flag as:

```cmake
target_compile_options(geos_cxx_flags INTERFACE
	"$<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:-ffp-contract=off>"
	"$<$<CXX_COMPILER_ID:GNU>:-ffp-contract=off>"
	"$<$<BOOL:${MSVC}>:/fp:precise>"
)
```

`${MSVC}` is a CMake variable, not a generator expression. CMake substitutes it while GEOS itself is configured, before generator expressions are evaluated per consumer. When GEOS is built with MSVC, `${MSVC}` is `1` at that point, so the string CMake writes into the exported `geos-targets.cmake` is already the literal `$<$<BOOL:1>:/fp:precise>`. This is baked in once at GEOS's own build time and applies to every consumer afterward, no matter what compiler that consumer uses.

The two `-ffp-contract=off` entries above it use `$<CXX_COMPILER_ID:Clang>` and `$<CXX_COMPILER_ID:GNU>`, generator expressions that CMake preserves unevaluated and re-evaluates against each consumer's own compiler at that consumer's generate time. That is why those two entries work correctly per consumer and the `/fp:precise` entry does not.

This is not specific to any particular packaging or distribution method. Any GEOS build performed with MSVC and then consumed by a project using a different compiler hits this, whether GEOS is built and installed directly, via a package manager, or by any other means. I first hit this via vcpkg's `x64-windows` triplet, which builds GEOS with MSVC by default; that is a common way to consume GEOS on Windows and reproduces the same exported config.

More broadly, this breaks every form of cross-compilation from a GEOS build performed with MSVC: any consumer using Clang, GCC, or any other non-MSVC toolchain on Windows, and any cross-compilation setup that targets a different platform or architecture from a different host compiler while linking against an MSVC-built GEOS. Because the flag is baked in at GEOS's own build time rather than evaluated per consumer, there is no consumer-side workaround short of patching or overriding the imported target's properties, as shown below.

This also breaks pre-compiled binary distribution of GEOS. Anyone who builds GEOS once with MSVC and distributes the resulting binaries and CMake config for others to consume, whether as a package manager binary cache, a prebuilt SDK, or any other shared build artifact, distributes an exported config that only works for MSVC consumers. Every non-MSVC consumer of that distributed binary hits this build failure, with no way to fix it on the consuming side other than overriding the imported target's properties after `find_package(GEOS)`.

### Minimal repro

Build and install GEOS directly with MSVC:

```
git clone --branch 3.14.1 https://github.com/libgeos/geos.git
cmake -B geos-build -S geos -G "Visual Studio 17 2022" -A x64 -DCMAKE_INSTALL_PREFIX=C:/geos-install
cmake --build geos-build --config Release --target install
```

Consume it from a separate project built with Clang:

```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.24)
project(geos_flag_leak_repro CXX)
find_package(GEOS CONFIG REQUIRED)
add_executable(repro main.cpp)
target_link_libraries(repro PRIVATE GEOS::geos)
```

```
cmake -B build -S . -G Ninja -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_PREFIX_PATH=C:/geos-install
cmake --build build
```

Result: build fails, Clang rejects `/fp:precise`.

### Suggested fix

Replace `${MSVC}` with the generator expression `$<CXX_COMPILER_ID:MSVC>`:

```cmake
"$<$<CXX_COMPILER_ID:MSVC>:/fp:precise>"
```

This generator expression is preserved unevaluated through `install(EXPORT ...)`/`export()` and is re-evaluated against each consumer's own compiler, instead of freezing GEOS's own build-time compiler into the exported config.

### Another latent instance of the same bug

The fourth entry in the same `INTERFACE_COMPILE_OPTIONS` list, `$<$<AND:$<CXX_COMPILER_ID:GNU>,$<EQUAL:4,8>>:-ffloat-store>`, has the same root cause: `${CMAKE_SIZEOF_VOID_P}` is a CMake variable substituted at GEOS's own build time, producing `$<EQUAL:4,8>` in the exported config rather than a generator expression that evaluates each consumer's own pointer size. On a 64-bit GEOS build this currently evaluates to false, so it causes no visible failure today. It will misbehave the same way `/fp:precise` does for any 32-bit consumer of a 64-bit-built GEOS, or a 64-bit consumer of a 32-bit-built GEOS: the flag will be applied or withheld based on GEOS's own build-time pointer size, not the consumer's. This has the same distribution consequence as the `/fp:precise` case: a GEOS binary built and distributed as 64-bit carries an exported config that only evaluates correctly for 64-bit consumers, and a 32-bit consumer of that same distributed binary has no way to fix it other than overriding the imported target's properties. The same fix applies: replace `${CMAKE_SIZEOF_VOID_P}` with a generator expression, for example `$<EQUAL:4,$<TARGET_PROPERTY:SIZEOF_VOID_P>>` or an equivalent consumer-evaluated check.

nyoma-diamond:
Resolves #1517.

## Problem

`geos_cxx_flags` is an `INTERFACE` library, installed and exported via `geos-targets.cmake` (`install(TARGETS geos geos_cxx_flags EXPORT geos-targets ...)`). Its `INTERFACE_COMPILE_OPTIONS` are usage requirements applied to every consumer that links `GEOS::geos`.

Two flags on that target were written as `$<$<BOOL:${VAR}>:...>`, where `${VAR}` is a plain CMake variable. CMake substitutes `${VAR}` at GEOS's own configure time, before the generator expression is written to `geos-targets.cmake`. The result is a literal `1` or empty string baked into the exported config, not a check re-evaluated per consumer.

- `$<$<BOOL:${MSVC}>:/fp:precise>`: A GEOS built with MSVC exports `/fp:precise` to every consumer, including ones using Clang or GCC, which reject the flag and fail to compile. This is the reported issue.
- `$<$<AND:$<CXX_COMPILER_ID:GNU>,$<EQUAL:4,${CMAKE_SIZEOF_VOID_P}>>:-ffloat-store>`: `${CMAKE_SIZEOF_VOID_P}` reflects GEOS's own pointer width, not the consumer's. Same bug, one line below the reported one.

Both differ from the sibling clauses in the same `target_compile_options` calls, which use `$<CXX_COMPILER_ID:...>`, a true generator expression CMake re-evaluates at the consumer's generate time.

## Fix

- `$<$<BOOL:${MSVC}>:/fp:precise>` → `$<$<CXX_COMPILER_ID:MSVC>:/fp:precise>`.
- `-ffloat-store`: There is no CMake target property for pointer width, so the equivalent per-consumer generator expression (`$<TARGET_PROPERTY:SIZEOF_VOID_P>`, as originally suggested in the issue) does not exist and errors at generate time (broke every GNU build and several older-CMake Clang builds in CI). The flag's own comment says it exists "to make some tests pass". I.e. it's about the numerical behavior of GEOS's own compiled object code, not a header-level requirement consumers need matched. Fixed by moving it off `geos_cxx_flags INTERFACE` entirely, onto `geos PRIVATE`, keeping the plain `${CMAKE_SIZEOF_VOID_P}` variable (now only affects GEOS's own build).

## Second instance found during testing

A cross-compiler smoke test (MSVC-built GEOS, consumed by Clang) surfaced a second case of the same root cause, not part of the original report: The OpenMP-SIMD flag block (`CMakeLists.txt`, "OpenMP SIMD" section) selects `/openmp:experimental` (MSVC), `-qopenmp-simd` (IntelLLVM), or `-fopenmp-simd` (else) based on GEOS's own `CMAKE_CXX_COMPILER_ID`, probes it with `check_cxx_compiler_flag` against GEOS's own compiler, then attached the result to `geos_cxx_flags INTERFACE` unconditionally. A GEOS built with MSVC exported `/openmp:experimental`; Clang rejected it with `no such file or directory: '/openmp:experimental'`.

`HAVE_OPEN_SIMD` and the SIMD flag are not referenced anywhere in GEOS's headers or `.cpp` sources (checked with a repo-wide search). This flag/define has no consumer-visible effect either way, so the fix moves both from `geos_cxx_flags INTERFACE` to `geos PRIVATE`, applied right after `add_library(geos ...)` where the target exists. `check_cxx_compiler_flag` still runs at its original location; only the two lines that consume its result moved, to keep the diff minimal.

### For discussion

This SIMD detection was added in `7403b6fde`, closing #1334, to back #1331 (vectorizing `Length.cpp`). #1331's merged form removed a loop-carried dependency and let the compiler auto-vectorize; it added no `#pragma omp simd`. Nothing in the current codebase reads `HAVE_OPEN_SIMD` or depends on the SIMD flag. It could be removed outright instead of made `PRIVATE`. Left as `PRIVATE` here since that's the minimal change and costs nothing if a future PR adds `#pragma omp simd` to GEOS's own source; flagging it in case maintainers prefer deletion.

## Scope check

Other `${MSVC}`/`${WIN32}`/`${CMAKE_SIZEOF_VOID_P}` occurrences in the tree were checked and are not affected:

- `_CRT_NONSTDC_NO_DEPRECATE`, `_SCL_SECURE_NO_DEPRECATE`, `_CRT_SECURE_NO_WARNINGS`, `NOMINMAX` on `geos_developer_cxx_flags`: This target is commented "Do NOT install this target for end-users!" and is not in the `install(TARGETS ... EXPORT geos-targets)` calls; build-tree only.
- `GEOS_DLL_EXPORT` on `geos`/`geos_c`: `PRIVATE`, not part of usage requirements.
- `CMAKE_SIZEOF_VOID_P` in `tests/xmltester/CMakeLists.txt`: Test-only, not part of an installed/exported target.

## CI

Added `windows-msvc-clang-consumer` to `.github/workflows/ci.yml`: builds and installs GEOS with MSVC, then builds `examples/` (both the C and C++ example executables) against that install using Clang. This reproduces the reported failure mode directly; it fails before this fix and passes after.


nyoma-diamond:
Not sure why some of the linux CI checks failed with hash mismatches. Seems like a problem with upstream CDN cache. In any case, I'm gonna make another commit to improve the naming of the new CI check and we can rerun from there and hopefully upstream will have sorted itself out by then

dbaston:
> geos_cxx_flags is an INTERFACE library, installed and exported via geos-targets.cmake

Should this be the case? My impression was that `geos_cxx_flags` was a way to store common flags for various GEOS targets, not something that should be part of GEOS' interface.

nyoma-diamond:
@dbaston 
> Should this be the case? My impression was that geos_cxx_flags was a way to store common flags for various GEOS targets, not something that should be part of GEOS' interface.

While I think that should be the intuition, to my understanding that's not how its actually implemented in practice. Namely line 218: `add_library(geos_cxx_flags INTERFACE)` and the following `target_compile_options(geos_cxx_flags INTERFACE ...)` calls mean `geos_cxx_flags` is treated like a library of its own which is then linked to `geos` publicly via `target_link_libraries(geos PUBLIC geos_cxx_flags PRIVATE $<BUILD_INTERFACE:ryu>)` (line 362) to enforce the necessary compile/link flags. Whether or not the flags being treated as a public library was the original intent, that is how it's being applied in practice. There may be an argument that this design confusion is the true source of the consumer leakage?

It does make sense that there should be a (non-library) mechanism/variable á la `geos_cxx_flags` to store common flags, but that's a non-insignificant design change from the current approach. I think a design change here would improve and clarify this behavior, but that may be better as the focus of a separate PR as I wanted to make this change relatively minimal to fix the cross-compilation error quickly and allow easy backporting.

dbaston:
Right, I'm wondering if it should be `target_link_libraries(geos PRIVATE geos_cxx_flags $<BUILD_INTERFACE:ryu>)`

nyoma-diamond:
Yeah that may be simpler. Though I worry that may break some other things unexpectedly? I fear the library-based approach is flawed and fragile to begin with and would be better handled using a `list` variable instead (so something like `$<$<COMPILE_LANGUAGE:CXX>:${geos_cxx_flags}>` instead). I'll experiment

nyoma-diamond:
Upon a quick experiment: it seems some of the options may (?) in fact be required to be passed forward in a way that just setting the flag library to link privately ends up hiding. In particular, the Linux Clang-12 build failed unexpectedly and others are producing lots of warnings that aren't normally present. It's not entirely clear why though. Investigating

nyoma-diamond:
Update: 

I think the build issues were caused by `USE_UNSTABLE_GEOS_CPP_API` being injected into `geos_cxx_flags` compile definitions under all builds (is this supposed to be the case?), which may be required to be visible to consumers, at least for CI (I think this is the cause of the warnings, and maybe the build failure).

Similarly, `geos_developer_cxx_flags` is also linked via `INTERFACE` specifically for developer builds. This is occurs when `GEOS_BUILD_DEVELOPER` is enabled, which itself occurs by proxy when `BUILD_BENCHMARKS` is enabled, which is done in by one of the CI jobs. Thus CI may require `geos_cxx_flags` to be `INTERFACE` to see the `geos_developer_cxx_flags` properly.

As a result setting `geos_cxx_flags` to private alone causes a failure because some necessary flags are not visible to the consumer. This is definitely a problematic blending of responsibilities that should be split up

**Edit:** I think I have the actual application case of some of the flags backwards between when they're on or off, but in either case there is definitely behavior that is being controlled by the existing approach and needs to be visible that just setting `geos_cxx_flags` to private ends up hiding.

nyoma-diamond:
@dbaston I've implemented a flag-list approach as suggested.

## Changes

- Replaced both `geos_cxx_flags` and `geos_developer_cxx_flags` `INTERFACE` libraries with plain CMake lists (`geos_cxx_private_*`, `geos_cxx_public_*`, `geos_developer_cxx_*`), applied directly to `geos` via `PRIVATE`/`PUBLIC`.
- Updated the three benchmark files that referenced `geos_cxx_flags` directly: targets recompiling GEOS's own sources now use the new lists; targets that just link `geos` had the redundant reference dropped.

## Caveats

- `PRIVATE`/`INTERFACE`/`PUBLIC` now mean exactly what they say per call, unlike the old design where everything on `geos_cxx_flags` reached `geos` itself regardless of keyword. Anything meant for GEOS's own build must be `PRIVATE` or `PUBLIC`, never bare `INTERFACE`.

---

While verifying this, I found an unrelated pre-existing bug: `_CRT_NONSTDC_NO_DEPRECATE` is gated on `$<BOOL:${MSVC}>`, which misses Windows-target Clang builds and breaks `GEOS_BUILD_DEVELOPER` builds with that toolchain. CI currently misses this entirely. Reported separately as #1528

nyoma-diamond:
@dbaston just bumping this and #1533 quickly as they should both be good to go :)

PR #1524, opened 2026-09-10
Detect overlay location inversions near collapsed ring edges

### LINKED ISSUE #1523: Overlay silently wrong when an operand has a zero-width collinear interior ring
# Overlay silently returns wrong areas when noding a nearly collinear interior ring

Intersection and difference of two valid polygons can silently return incorrect areas when one polygon has an extremely thin triangular hole and the other polygon's boundary crosses it. The floating-point overlay returns successfully, so the robust overlay wrapper does not retry.

## Reproduce

Save these geometries as `a.wkt` and `b.wkt`, respectively:

```text
POLYGON ((379.694383 3423, 382 3423, 382 3428, 379.694383 3428, 379.694383 3423))
```

```text
POLYGON ((380 3428.3, 382 3429, 388 3422, 377 3422, 378 3427, 380 3428.3),
         (379.153701 3427.199712952367, 380.736005 3427.1997128766443,
          379.777158 3427.199712922531, 379.153701 3427.199712952367))
```

Both geometries pass `isValid`. A lies inside B's shell. B's hole is nearly collinear, but **not exactly collinear or zero-area**: exact rational arithmetic on the parsed binary64 coordinates gives a hole area of approximately `1.3650878779e-13`.

```sh
geosop -a a.wkt isValid
geosop -a b.wkt isValid
geosop -a a.wkt -b b.wkt -f wkt intersection
geosop -a a.wkt -b b.wkt -f wkt difference
```

Without the fix, intersection returns:

```text
POLYGON ((379.694383 3427.1997129264923, 380.736005 3427.1997128766443,
          379.777158 3427.199712922531, 379.694383 3427.1997129264923))
```

| Operation | Observed area | Expected area, approximately |
| --- | ---: | ---: |
| A intersection B | 5.01746e-15 | 11.528085 |
| A difference B | 11.528085 | 0 |

The exact difference is a tiny portion of the hole, not literally empty. The error is that almost all of A is assigned to the wrong result.

Independently reproduced on GEOS 3.13.1 and the unpatched development revision `55809ca8d` (3.16.0dev). Swapping the inputs also gives incorrect results; for B difference A, the expected area is approximately `36.571915`, rather than approximately `48.1`. Keep the hole's ordinates as written: this example depends on rounding at binary64 precision.

Snap-rounding computes the expected areas for this example:

```sh
geosop -a a.wkt -b b.wkt -f wkt intersectionSR 100000000
geosop -a a.wkt -b b.wkt -f wkt differenceSR 100000000
```

Input configuration and the local winding reversal after noding

## What happens

Exact rational calculations on the parsed doubles establish the following numerical mechanism:

1. The two intersections of A's left edge with the hole differ in y by approximately `1.496359985e-13`. One ulp at this y-coordinate is `4.547473509e-13`. Both crossings round to `X = (379.694383, 3427.1997129264923)`.
2. Replacing the crossings with X pinches the hole. The remaining triangle, in vertex order X, second hole vertex, third hole vertex, has signed area approximately `-5.017455576e-15`, whereas the original ring has positive signed area. Its winding has reversed on the noded coordinates.
3. The boundary side labels come from the original ring, while the angular order at X comes from the noded coordinates. This example gets through side-location consistency checking but propagates incorrect locations to A's boundary. Intersection then assembles only the tiny triangle.

The existing result-area heuristic does not detect this case: intersection is checked against upper bounds, which a tiny result satisfies. Difference also satisfies its existing bounds here. This does not imply that every operation's area check is only an upper-bound check.

The unpatched development revision already includes the angular-ordering change from GH-1499. This reproduction therefore needs additional handling beyond that change.

## Proposed fix

After labelling, check for inconsistent edge locations at nodes incident on a collapsed edge. For each eligible non-boundary, non-collapsed edge, compare its label with a point-in-area query against the original input, using a midpoint sample from its longest segment with an eligible midpoint. Skip samples that round to a segment endpoint, are non-finite, or locate on the input boundary. A mismatch raises `TopologyException`, allowing `OverlayNGRobust` to retry with its existing snapping strategies.

**Apply this check only to the default floating-point noding path for linear inputs.** Do not apply it with a fixed precision model, a caller-supplied noder, or curved inputs. Snapping and precision reduction can legitimately change locations relative to the original geometry; rejecting those changes would break the fallback strategies. A supplied noder can snap even when the overlay precision model is floating.

This is a targeted consistency heuristic. It detects the inversion in this reproduction without adding a distance tolerance to the check. It is not a proof that all possible inversions have collapsed edges, nor a complete validation of the overlay graph. The existing robust retries retain their own snapping tolerances.

d-hage:
Fixes #1523 

Floating-point noding can merge two crossings of a nearly collinear triangular hole into one node, reversing the winding of the remaining triangle. Side-location propagation can then assign self-consistent but incorrect labels without raising an exception. In the reported example, intersection returns a tiny sliver instead of almost all of A, while difference retains almost all of A.

This change adds a targeted consistency check after labelling. At nodes incident on collapsed edges, it samples eligible non-boundary edges and compares their propagated locations against the original input areas. A mismatch raises TopologyException so OverlayNGRobust can retry using its existing snapping strategies.

The check applies only to default floating-point noding for linear inputs. Fixed precision and caller-supplied noders are excluded because snapping can legitimately change locations relative to the original geometry. Curved inputs are also excluded. This is a heuristic for detecting collapse-adjacent inversions, not a complete validation of overlay topology.

## Validation

- Regression coverage checks input validity, detection by floating overlay, and valid robust results with the expected intersection and difference areas in both argument orders.
- Explicit snapping of the reported example produces the expected areas.
- Fixed-precision and caller-supplied snap-rounding tests cover cases that an unconditional location check would incorrectly reject.
- The inversion test fails without the fix; both precision-reduction tests fail with the earlier unconditional check.
- All 535 configured CTest tests pass in a Release build on arm64 macOS with AppleClang.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 2027-xx-xx
 
 - Fixes/Improvements:
-  -
+  - Detect overlay location inversions near collapsed edges of nearly collinear rings
 
 ## Changes in 3.15.0
 2026-09-01
--- a/include/geos/operation/overlayng/OverlayLabeller.h
+++ b/include/geos/operation/overlayng/OverlayLabeller.h
@@ -81,6 +81,13 @@ private:
     * but may be more efficient and accurate to do it here.)
     */
     void labelCollapsedEdges();
+
+    /**
+    * Verifies a propagated (non-boundary) edge location for a geometry
+    * against the original input geometry, at a point in the interior of
+    * the edge.  Throws TopologyException on a mismatch.
+    */
+    void checkEdgeLocation(OverlayEdge* edge, uint8_t geomIndex) const;
     static void labelCollapsedEdge(OverlayEdge* edge, uint8_t geomIndex);
 
     /**
@@ -170,6 +177,17 @@ public:
     * Computes the topological labelling for the edges in the graph->
     */
     void computeLabelling();
+
+    /**
+    * Checks for inconsistent propagated locations at nodes incident on
+    * collapsed edges, by sampling non-boundary edges in the input areas.
+    * This is a heuristic for detecting topology inversions introduced by
+    * floating-point noding, not a complete validation of the labelling.
+    * Must only be used for unsnapped linear overlay: snapping and precision
+    * reduction can legitimately change locations relative to the inputs.
+    * Throws TopologyException on a mismatch so robust overlay can retry.
+    */
+    void checkCollapseNodeLocations() const;
 
     /**
     * Scans around a node CCW, propagating the side labels
--- a/src/operation/overlayng/OverlayLabeller.cpp
+++ b/src/operation/overlayng/OverlayLabeller.cpp
@@ -22,6 +22,7 @@
 #include <geos/util/Assert.h>
 #include <geos/util/TopologyException.h>
 
+#include <cmath>
 #include <sstream>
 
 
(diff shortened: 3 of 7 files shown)


d-hage:
@dbaston @dr-jts, could you review this fix? I found the failure while checking the areas returned by overlays of valid polygons: almost all of one input was silently assigned to the wrong result. The reduced example in [#1523](https://github.com/libgeos/geos/issues/1523) shows how a nearly collinear hole can collapse during noding and invert the propagated labels. Because the operation succeeds, `OverlayNGRobust` does not retry. This PR detects that inconsistency so the existing robust retry can run.

The same validation work exposed two separate defects: `CoverageCleaner` with zero snapping can merge adjacent polygons ([#1525](https://github.com/libgeos/geos/issues/1525), fixed in [#1526](https://github.com/libgeos/geos/pull/1526)), and an explicitly requested fixed-grid intersection can fail on valid triangles ([#1535](https://github.com/libgeos/geos/issues/1535), with a separate fix). All three have small, self-contained reproducers. The first two patches are already in a custom GEOS build; carrying them through upgrades is difficult, and these failures can silently change results or stop processing. I would appreciate a review soon and, if the approach holds up, getting the fixes upstream.

PR #1526, opened 2026-09-10
Node collinear overlaps when snapping tolerance is zero

### LINKED ISSUE #1525: CoverageCleaner merges adjacent polygons when snapping distance is zero
With snapping disabled, `CoverageCleaner` can remove a shared boundary and assign both polygons' area to one input. This occurs with exact integer coordinates; it does not require floating-point rounding error.

## Reproducer

Both polygons are individually valid. Their interiors are disjoint, but their shared edge has an unmatched vertex at `(1 1)`, so they are not yet a valid edge-matched coverage.

```text
A = POLYGON ((1 0, 1 1, 1 3, 4 0, 1 0))
B = POLYGON ((0 4, 1 3, 1 0, 0 4))
```

```cpp
std::vector<const Geometry*> coverage{a.get(), b.get()};
CoverageCleaner cleaner(coverage);
cleaner.setSnappingDistance(0);
cleaner.setGapMaximumWidth(0);
cleaner.clean();
auto result = cleaner.getResult();
```

Actual result, allowing for ring rotation:

```text
0: POLYGON ((1 0, 0 4, 1 3, 4 0, 1 0))  -- area 6
1: POLYGON EMPTY                         -- area 0
```

Expected: retain A's area 4.5 and B's area 1.5, inserting `(1 1)` into B's shared edge. Automatic snapping produces that result.

A PostGIS reproducer, using a full partition ordered by id:

```sql
WITH cov(id, g) AS (VALUES
  (0, 'POLYGON ((1 0, 1 1, 1 3, 4 0, 1 0))'::geometry),
  (1, 'POLYGON ((0 4, 1 3, 1 0, 0 4))'::geometry))
SELECT id, ST_AsText(
  ST_CoverageClean(g, 0.0, 0.0, 'MERGE_MIN_INDEX') OVER (
    ORDER BY id ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
  )
)
FROM cov
ORDER BY id;
```

The C++ failure was independently reproduced on GEOS main commit `55809ca8d`. The same sample was also reported as reproduced with PostGIS 3.6.3 / GEOS 3.14.1; that installation was not part of the independent C++ verification.

## Cause

`SnappingIntersectionAdder::processIntersections` handles single-point intersections directly, but delegates two-point collinear overlaps to `processNearVertex`. That method inserts a node only when `distance < snapTolerance`. At zero tolerance, even an endpoint lying exactly on the other segment cannot satisfy this condition.

Consequently, `SnappingNoder(0)` leaves both shared-boundary chains:

```text
(1 0, 1 1, 1 3)
(1 3, 1 0)
```

`LineDissolver` cannot deduplicate these differently segmented chains. Polygonization then produces one face covering both polygons. Its interior point `(1.5625, 1.5)` is inside A, so the entire face is assigned to A. No overlap is recorded and no exception is thrown.

The incorrect output still passes `CoverageValidator::isValid()`. The issue is loss of the shared boundary and incorrect per-input area assignment, rather than an invalid output coverage.

[The documented zero setting](https://libgeos.org/doxygen/classgeos_1_1coverage_1_1CoverageCleaner.html) disables snapping. This example only needs insertion of an existing vertex on an exactly collinear segment; no coordinate movement is required.

## Proposed fix and workaround

At zero snapping tolerance, process both collinear overlap endpoints returned by the line intersector. Keep the existing near-vertex handling for positive tolerances. This fixes the omission in the noder used by CoverageCleaner.

Automatic snapping (`snappingDistance = -1`) works around this sample. It permits coordinate movement, so it is not equivalent to a repaired zero-tolerance mode.

d-hage:
Fixes #1525

`SnappingIntersectionAdder` delegates collinear overlaps to near-vertex processing, which requires `distance < snapTolerance`. At zero tolerance, this skips even exact intersections. When adjacent polygons have differently segmented shared edges, CoverageCleaner can consequently polygonize across that boundary and assign both polygons to one input.

Process both overlap endpoints returned by the line intersector when the tolerance is zero. Positive-tolerance noding retains its existing behavior. No tolerance is introduced and the fix does not move the collinear endpoints.

## Validation

- The coverage regression checks preservation of each polygon, matched edges, valid output coverage, and no reported overlaps or merged gaps. It covers both input orders, all four overlap strategies, and zero versus automatic snapping.
- Noder regressions check an unmatched shared-edge vertex and a contained diagonal segment, with output noding validation. Both fail before the fix.
- Nearby non-intersecting lines remain unchanged at zero tolerance.
- The coverage regression fails before the fix, returning areas 6 and 0 instead of 4.5 and 1.5.
- All 535 configured CTest tests pass in a Release build on arm64 macOS with AppleClang.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 2027-xx-xx
 
 - Fixes/Improvements:
+  - Fix CoverageCleaner merging adjacent polygons when snapping is disabled
   -
 
 ## Changes in 3.15.0
--- a/src/noding/snap/SnappingIntersectionAdder.cpp
+++ b/src/noding/snap/SnappingIntersectionAdder.cpp
@@ -60,16 +60,19 @@ SnappingIntersectionAdder::processIntersections(SegmentString* seg0, std::size_t
     if (!isAdjacent(seg0, segIndex0, seg1, segIndex1)) {
         li.computeIntersection(p00, p01, p10, p11);
         /**
-         * Process single point intersections only.
-         * Two-point (collinear) ones are handled by the near-vertex code
+         * With a positive tolerance, two-point (collinear) intersections
+         * are handled by the near-vertex code.  At zero tolerance that
+         * code does not add nodes, so process both overlap endpoints here.
          */
-        if (li.hasIntersection() && li.getIntersectionNum() == 1) {
+        if (li.hasIntersection() && (li.getIntersectionNum() == 1 || snapTolerance == 0)) {
+            for (std::size_t i = 0; i < li.getIntersectionNum(); ++i) {
 
-            const auto& intPt = li.getIntersection(0);
-            const auto& snapPt = snapPointIndex.snap(intPt);
+                const auto& intPt = li.getIntersection(i);
+                const auto& snapPt = snapPointIndex.snap(intPt);
 
-            static_cast<NodedSegmentString*>(seg0)->addIntersection(snapPt, segIndex0);
-            static_cast<NodedSegmentString*>(seg1)->addIntersection(snapPt, segIndex1);
+                static_cast<NodedSegmentString*>(seg0)->addIntersection(snapPt, segIndex0);
+                static_cast<NodedSegmentString*>(seg1)->addIntersection(snapPt, segIndex1);
+            }
         }
     }
 
--- a/tests/unit/coverage/CoverageCleanerTest.cpp
+++ b/tests/unit/coverage/CoverageCleanerTest.cpp
@@ -463,4 +463,47 @@ void object::test<20> ()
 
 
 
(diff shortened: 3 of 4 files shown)


d-hage:
@dbaston @dr-jts, could you also review this zero-tolerance fix? I found it by checking each input's area after `CoverageCleaner` ran with snapping disabled. In the reduced integer-coordinate example in [#1525](https://github.com/libgeos/geos/issues/1525), two adjacent polygons with areas 4.5 and 1.5 become one polygon of area 6 and an empty result. The output still passes coverage validation, so a caller can silently lose the per-input assignment. This PR nodes exact collinear overlap endpoints at zero tolerance without moving coordinates.

